PR #44, opened 2018-07-14
[WIP] Stores And Store system.

ThisMushroom:
Added ItemHelper to create items.
Added [Lorencia] Pasi The Mage Store (Global like).
Added [Lorencia] Hanzo The Blacksmith Store (Global like).
Added [Lorencia] Wandering Merchant Store (Global like).

[WIP] [Lorencia] Potion Girl Store (Has TODOs) (Global like).
[WIP] NpcInitialization - Maybe create separate store initialization system to read stores from xml.
[WIP] ItemHelper - Maybe add some enums for more readability in item creation process. (Like named item groups).
[WIP] ItemHelper - Further and **HARD** refactoring of ItemHelper (make it more consistant, remove code duplicates). It was made just to fit needs right now. Should be hardly refactored.

Still I don't really like ItemHelper as non static class.
Should refactor DataInitialize as well (who cares that item creation methods are used there for testing data creation purposes).
ItemHelper could become nice thing for GM item related stuff.

### DIFF
--- a/src/Persistence/Initialization/Items/ItemHelper.cs
+++ b/src/Persistence/Initialization/Items/ItemHelper.cs
@@ -1,0 +1,175 @@
+﻿namespace MUnique.OpenMU.Persistence.Initialization.Items
+{
+    using System.Linq;
+    using MUnique.OpenMU.AttributeSystem;
+    using MUnique.OpenMU.DataModel.Configuration;
+    using MUnique.OpenMU.DataModel.Configuration.Items;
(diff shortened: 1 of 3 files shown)


sven-n:
Some comments are missing (because the methods are now public), but everything else seems okay.

About creating Items and ItemDefinitions: I think we could create something like a fluent AP for this stuff. Passing a lot of (numerical) parameters is not readable.

About NPC stores: I'm not a friend of putting of adding all sets with luck into the stores, but that's just personal preference ;)

Issue #39, opened 2018-07-10
[Error] Monster walking behaviour


ThisMushroom:
### Overview:

1) Add new SpawnArea with any kind of monster (e.g. Giants (Lorencia))
2) Make SpawnArea as a single cell spot (x1 == x1 && y1 == y2)
3) Make SpawnArea spawning 20 monsters
4) Reinit database for spawn area initialization 
5) Walk to the spawn area 
6) Observe empty spot

### Observation:

1) Monsters are walking towards 0,0 coordinates of the map.
2) Some monsters remain standing/walking around the spot area.
3) For DirectHit spell system monsters that remain standing/walking are acting correctly when being hit.
3.1) If you want to hit them with DirectHit spell - hit them - they will die, you will get exp/zen. No drop.
4) For AreaHit spell system those monsters that remain standing/walking are invisible.
4.1) If you want to hit them with AreaHit spell - try to - the will act like being hit but no damage would be done. Therefore as we can see from debugging - system doesn't detect monsters around you in the specified skill range. 

### Ideas:

1) Yup, monsters are walking thats obvious. - Need to implement correct work of walk range (like the definition of how far they can move from initial spawn area (for the single cell spawns) and new Point distance (for area spot that is bigger then walk range). Moreover implement logic to prevent monsters from leaving the SpawnArea.

2(+ 3-4)) Didn't debug it well yet, just shortlooked. But first ideas and observations were: 
- If you take Evil Spirits and start casting somewhere in the corner of the map near 0,0 (as close as you can) - you will hit those monsters (but you cannot see them). Why do I think they stand at 0,0 - because of EXP trail from monster to caster after being killed (not alway working) + you get exp and zen.
- It may be problem with server-client sync and maybe somehow bucket system bug. 
When on server side monsters walked to 0,0 but on client side they are displayed on other places. (take a look at the map in the admin panel - all monsters are on correct places but ingame - they are not).
Why we can hit them with DirectHit spells - because client send the data about monster and it is ok to calculate hit.
Why we cannot hit them with AreaHit spell - because non-target spell use being sent to the server and the server should determine monsters being hit int the spell range around the caster.

**PS. I would investigate it deeper this eve and add comments about to make it more detailed.**

sven-n:
Thanks again 👍 You're probably the first one who tried to learn and perform a skill on the OpenMU server 🥇 
I wondered that it even worked somehow, btw :)

Just a wild guess:
[...]

PR #47, opened 2018-07-18
Fix for the invisible respawns after walking

sven-n:
This should fix the issue of invisible respawns, see also #39.
Created a branch and pull request to test the new behavior.


ThisMushroom:
Tested.

Problem remains.
Got several ideas why. Goint to test them today.

sven-n:
Are you sure that the issue still occurs? I tested it in the dungeon and it works perfectly for me. I even let evil spirit running for several hours. At the end, all monsters which were visible on the Admin Panel Map were also visible in game.

sven-n:
Oh just forgot to mention that there is an issue in the map of the admin panel... monsters jumping, because y is passed instead of x in the WorldObserverHub when it sends the walk steps

Issue #55, opened 2018-08-08
Admin panel not working on Windows


tehKaiN:
So I've just installed VS2017 with .net worload. Server starts (I guess, everything's green in console), but I've got blank page and errors in web console:

```
Loading failed for the <script> with source “http://localhost:1234/content/js/app.js”. admin:13
Loading failed for the <script> with source “http://localhost:1234/components/App”. admin:1
Error: "Fetching http://localhost:1234/components/App
  Instantiating http://localhost:1234/components/App
  Loading components/App"
ahttp://localhost:1234/content/js/system-production.js:4:11962
```
There was a problem earlier with unhanled exception which I fixed by typing command left in comment in elevated command prompt: `netsh http add urlacl http://+:1234/ user=[Username]` where username I've wrote my OS user name.

I've tried to install "JavaScript & TypeScript Language Support" package but no luck.

Using latest VS2017 Community, Win10 Enterprise

sven-n:
Hi, it seems like the TypeScript build doesn't run. As far as I can tell by looking at my own VS installations, you need the following Visual Studio extensions:
  * TypeScript Build for Microsoft Visual Studio
  * TypeScript for Microsoft Visual Studio

If it works for you, I'll add it as requirement in the quick start guide.

It also doesn't hurt to have the newest VS update, I guess. I'm currently running versions 15.3.4 and 15.7.1 on my machines.

sven-n:
Sure, if it's not too much work and helpful for others ;-)

sven-n:
I think I know what's the only requirement to build the Typescript stuff: The Typescript SDK ;)
That's what I had to install after an update to 3.0.

Alex-G:
if I create a new user through the Admin panel, the password field is empty in the database

sven-n:
Oh, yes that's a bug... the AdminPanel (typescript code) transmits the property as "Password", but the AccountModule has it defined in AccountDto as "NewPassword".
Thanks for reporting, I'll fix it soon ;)

sven-n:
I guess we can close this issue now ;)

PR #56, opened 2018-08-08
Fixes for newest packages

tehKaiN:
Result of #55 

I have 51 untracked files, mostly .js, so I guess .gitignore is missing something.


sven-n:
Hey, thanks for the pull request first. I hope you didn't waste too much time with it.
I didn't know you updated so many packages at once. It's really difficult to understand some of the changes (e.g. deleted files).
I have to admit some packages are not up to date, some also by purpose. In the past I often encountered incompatiblities between some of the dependencies, so I better left them as they are. For example, updates of Netwonsoft.Json were always a lottery. Some of the issues also just appear during execution of the project, so it's a bit tricky...

I started migrating to .net core/standard today where I also had to update some of these packages. It's not finished yet, but I don't think the changes of this pull request would live long. Therefore it's not really worth the trouble at the moment.

tehKaiN:
nah, it was 5 mins of work. I'm really happy to hear about dotnet core. Let's close it then.

PR #57, opened 2018-08-09
Fix incorrect parsing gameServer's ipAddress

afrokick:
* fixed ipAddress parsing in ServerListItem.cs
* refactoring ServerListItem with ArrayExtensions for working with data array


sven-n:
Thanks, good work 👍🏻

PR #60, opened 2018-08-11
[WIP] Migration to .net core/standard

sven-n:
Basically, I converted all project files to the new format and tried to resolve some issues of stuff which is not available anymore in .net core:
  * The painting of the maps in the LiveMapModule is now done via SkiaSharp instead of System.Drawing
  * Replaced AspNet.SignalR by AspNetCore.SignalR. Because of the changed API I had to change the code (C# and TypeScript) a lot, too. The log4net logging appender is not done yet!
  * Replaced RhinoMocks by Moq, since RhinoMocks isn't available for .net standard/core

The unit tests are all green.

# Work in Progress
Because the new log4net-Appender for SignalR is not done yet, and I didn't do some tests with the real game client, this pull request is not yet ready to merge.
The SystemHub is also not ready yet, since PerformanceCounters are not included in .net standard. I'm unsure if there is a cross-platform replacement.


sven-n:
Begin/EndAccept was added with .net standard 2.0.
Probably the Travis CI build environment isn't up-to-date or the configuration is wrong.

sven-n:
I have a very strange problem now when running the server under .net core. After logging in, the client gets stuck if the login was actually successful. For some reason the client doesn't request the character list - that would be the next expected step. The next problem: After disconnecting, the account is still logged in at the LoginServer.
If I would have to guess, Sockets behave somehow different under .net core :-/

sven-n:
I don't know, but I'll check if it can work like that on windows, too.

sven-n:
I found the problem with the changed Socket behavior. In fact it's a bug in my code which doesn't come up running with the normal.net framework. Socket.SendAsync returns a boolean value which isn't considered in the code. In the normal framework it always returns true which means the i/o operation is pending and will be done asynchronous. In .net core it can happen that it returns false and the data was already sent synchronous.

sven-n:
The SignalR stuff should work now, more or less...

@afrokick   can you check if it works for you now?

sven-n:
I guess I’ll remove the SystemHub together with the System Page in the AdminPanel. At the moment there is no proper replacement for these Performance Counters in .net standard/core.
If someone needs to monitor the state of a server system there are better tools anyway.

Apart of that, I think the migration is finished :)

afrokick:
@sven-n I checked it on commit #98abb4e3 but it can't build

`~/git/OpenMU/src/AdminPanel/AdminPanel.cs(19,19): Error CS0234: The type or namespace name 'Log4Net' does not exist in the namespace 'MUnique' (are you missing an assembly reference?) (CS0234) (MUnique.OpenMU.AdminPanel)`

sven-n:
Did you restore nuget packages? This missing thing got added ;)

afrokick:
I always restore packages for solution. But it seems like you should do it for all projects separately...
My issue disappeared. Thank you) Now admin panel works properly. I'm going to try to connect from unity to gs...

sven-n:
I would assume that your socket creation code should work. The connect server listens on all local network interfaces (IPv4), including 127.0.0.1.
The Network.Analyzer project, for example, uses similar code:
```csharp
        private void ClientConnected(object sender, ClientAcceptEventArgs e)
        {
            var clientConnection = e.AcceptedConnection;
            var serverSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
            serverSocket.Connect(this.txtOtherServer.Text, (int)this.numRealGSPort.Value);
            var encryptor = new ComposableEncryptor().AddEncryptor(new Xor32Encryptor()).AddEncryptor(new SimpleModulusEncryptor(SimpleModulusEncryptor.DefaultClientKey));
            var decryptor = new ComposableDecryptor().AddDecryptor(new SimpleModulusDecryptor(SimpleModulusDecryptor.DefaultClientKey) { AcceptWrongBlockChecksum = true });
            var serverConnection = new Connection(serverSocket, encryptor, decryptor);
            // ...
        }
```

Alex-G:
how to run the Gameserver on linux?
I tried with dotnet run in the Gameserver folder ...
I get "Unable to run your project"

Issue #35, opened 2018-05-21
Use Span<T> (and Pipe API) for network code and packet handling


sven-n:
C# offers a new way to handle pieces of arrays: [C# - All About Span: Exploring a New .NET Mainstay](https://msdn.microsoft.com/en-us/magazine/mt814808.aspx). I think it could help to improve the performance in the network code and would result in a lower amount of heap allocations in this project. This means, the garbage collector would have less work to do.

Currently, we make good use of byte arrays, e.g. for every received network packet at least one byte array is created. If a packet was encrypted, e.g. by SimpleModulus, even more byte arrays are created.

sven-n:
I tested it a bit and created a branch for it: https://github.com/MUnique/OpenMU/tree/span

sven-n:
The new Pipe API could be amazing for this projection, too :)

More info about the new API can be found here: https://blog.marcgravell.com/2018/07/pipe-dreams-part-1.html

sven-n:
The new Pipe API could be amazing for this project, too :)

More info about the new API can be found here: https://blog.marcgravell.com/2018/07/pipe-dreams-part-1.html

sebastian-heinz:
I would have the following suggestions to improve the network performance:
1. To avoid GC to collect objects, all objects should be initialised on startup and reused. So that no `new`-keywords are necessary during the network read / write operations and references are kept.
2. A big buffer array (to pin it on the large object heap [https://stackoverflow.com/questions/8951836/why-large-object-heap-and-why-do-we-care]) should be created and potions of it should be assigned to the `SocketAsyncEventArgs` via `SetBuffer()`-Method. This prevents compacting of the buffer thus improves performance.
3. To avoid copying memory / object instantiation we should use Span<T>
4. This will be only possible by limiting the total number of connections from the beginning. As the necessary resources would be calculated on startup. (Ex. BufferSize = MaxConnections * BufferSize, Number of instantiated `SocketAsyncEventArgs`at the beginning, etc. will all be based on that number)
5. Additionally investigate if the c# Pipeline is better than a big buffer. (https://github.com/dotnet/corefx/issues/31953) As there might be some concerns with implementation.

The following changes should be applied to https://github.com/MUnique/OpenMU/blob/master/src/Network/Listener.cs and https://github.com/MUnique/OpenMU/blob/master/src/Network/Connection.cs as they look like the Server / Client core parts of the application if I'm not mistaken.

I would propose the following process of change:
[...]

PR #62, opened 2018-10-02
[WIP] Pipelined network connections

sven-n:
See also #35.
I added a first version of a PipelinedConnection.

The encryption & decryption is not yet optimized and still works on manually created byte arrays. I think there are several ways to refactor it:
1. They could work directly with pipes, too.
2. They could work on Span<byte> using the shared MemoryPool if necessary.

I prefer the first option ;-)


sven-n:
I added pipelined versions for the simple Xor32 de- & encryptors. I think the Pipelines API is pretty awesome :)

TODO:
  * Adding pipelined versions for the SimpleModulus and Xor3 algorithm
  * Unit & Integration tests
  * Benchmark tests
  * Actual usage of the pipelined stuff in OpenMU

sven-n:
Added a pipelined version for the SimpleModulus algorithm and some unit tests.
I have to say, the API is much cleaner than my previously used approach where I have actually passed references to byte arrays into public en/decrypt methods or sometimes even directly modified the source array and returning it as result. Such surprises will be things of the past :) Passed in data is not modified anymore.

sven-n:
From the code-side it's finished. I would like to add bit documentation before I merge this.

PR #66, opened 2018-10-14
Added test characters to test elfs and wizards

sven-n:


sven-n:
Character names of the test accounts got now the suffixes "Dk", "Elf" and "Wiz".

PR #74, opened 2018-10-24
Made Life Swell Skill working

sven-n:


sven-n:
One known limitation: It just has effect on the own player. Effects on the party is not yet implemented.
We need to extend the Skill or MagicEffectDefinition configuration by a flag, e.g. "AddEffectToPartyMembers", or similar. The configuration is probably missing some more stuff which will be required for other skills... I might have to analyze this first.

PR #79, opened 2018-10-29
Reusing PathFinder in Monster class instead of always creating new one

### LINKED ISSUE #76: Reuse PathFinder objects
Currently, every started walk operation of a NPC [creates](https://github.com/MUnique/OpenMU/blob/fdf0c16f0eab75a4b0948f8cf6d4640197004878/src/GameLogic/NPC/Monster.cs#L118) a new PathFinder object. As this class itself creates many other instances as well, the garbage collector gets a lot of work to do.
So, it would be nice if the created PathFinder object could be reset to an initial state and reused. To reset a PathFinder, it's probably only required to clear the openList, the resultList and setting stop to false.

hatalaef:
Resolves #76

### DIFF
--- a/src/GameLogic/NPC/Monster.cs
+++ b/src/GameLogic/NPC/Monster.cs
@@ -29,6 +29,7 @@ namespace MUnique.OpenMU.GameLogic.NPC
         private Timer respawnTimer;
         private int health;
         private bool isCalculatingPath;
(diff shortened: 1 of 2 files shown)


sven-n:
Thanks, I‘ll take a look at it soon :)

PR #88, opened 2019-01-31
Added ItemRepairPrice calc

And3rsL:
Add basic repair price calculation, not 100% same as client shows but is ok for now

-- Need to find NPC repair discount

### DIFF
--- a/src/GameLogic/ItemPriceCalculator.cs
+++ b/src/GameLogic/ItemPriceCalculator.cs
@@ -263,6 +263,30 @@ namespace MUnique.OpenMU.GameLogic
             return RoundPrice(sellingPrice);
         }
 
+
+        /// <summary>
+        /// Calculates the repair price of the item, which the player has to pay if he wants to repair the item.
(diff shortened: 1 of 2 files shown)


sven-n:
Nice, thanks. I‘ll look at it soon :)

sven-n:
In game the npc(e.g. at Hanzo the Blacksmith) is 2.5x cheaper. Example: normal repair: 2500, at npc: 1000

I didn’t check yet if there is another discount when repairing the whole inventory at once.

I think the slight differences could also come from a missing call to RoundPrice.

In my limited time only did some quick checks with high excellent items and I think there might be a general error in the item price calculation :-/ I’ll take a look next week again.

sven-n:
why did you close it? ;)

And3rsL:
i think i messed up with github and deleted it! 
i will push itemprice when its 100% done and tested

sry

sven-n:
No problem, I’ll work on it using your code as a base :)

PR #90, opened 2019-02-06
Now killing player increment player kill status

And3rsL:
Need to think about self defense system and how to implement it


sven-n:
Thank you, I will look at it but I‘m quite busy this week.

Issue #92, opened 2019-02-10
Solution Organization


Apollyon221:
Hey i am new to this project, so i wanted to help out, started reading the code and while i was doing that i saw what a mess the solution is(at least for me), "*.Tests" projects together with normal projects and etc.

I will be starting a fork today and do some heavy modifications to the Solution and projects organization, i dont know if you will like it, but if in any case you do i will do a pull request if not i will have to develop in my fork you are free to copy and paste my code.

And a question would you mind if i started a new repo and reworkd the project in my way and continued the development there?

Sorry if i was rude, if i was i didnt mean to, my english is not that good so sometimes i might seem rude and that is not what i intended, i really liked the project and wanted to get into it, i just dislike its organization.

sven-n:
Hello,
the solution structure is as it was intended by me.

Why would you remove the *.Tests from the solution? If you refactor your code (Renaming things, for example), the tests are updated almost automatically. If you remove them from the solution, they are not. Also, you can let them run all the time during development...

I can’t tell you if I would merge your changes as I don’t clearly see what’s your plan ;)

Apollyon221:
I dont mean to remove them, i mean to organize them in folders so it would look like something like this

[https://i.imgur.com/v7geasT.png](Image)

This way it is more friendly for other devs to read and know the project much faster, as it will be more organized.

Another critic from me, it seems like you are splitting projects just for the sake of splitting, an example MUnique.OpenMU.ChatServer and MUnique.OpenMU.ChatServer.ExDbConnector could be on the same project that way it would be way more easier to find what you need and more organized, i am just trying to help my english is not that good sometimes my words doesnt mean what i want they to mean.

Do you have discord or something similar so we could talk more easily?

sven-n:
If this structure of your Screenshot doesn’t require changes in the actual folder structure, I think it’s fine.

And about the ChatServer and ExDbConnector: The ExDbConnector is a side product, which is a replacement for the original ChatServer.exe. That means, it’s not used by the rest of the project. It’s also built for .net core and framework 4.6.1, while ChatServer is built for .net standard.

The only project which is currently so small that it probably doesn’t need a separate project is the LoginServer.

I don’t have Discord, just Gitter and Skype. However, I’m not much more “responsive” there as in the issues ;)

sven-n:
I added a folder structure to the solution (master branch). What do you think about it?

sven-n:
Closed since no further feedback came back.

PR #94, opened 2019-02-10
Organizing the solution

Apollyon221:
Like we discussed before on #92 just organizing the solution and certain projects a little bit better, having 29 projects on the same folder is not a good idea, hopefully you like the folder structure if not tell me what you would like changed or change it.


sven-n:
Thank you, it looks mostly reasonable to me. I just have two suggestions about the changes:
 * Can you please remove the "Interfaces"-namespaces in PathFinding, GameLogic and Network again? It makes no sense to create namespaces just for interfaces.
 * I would like to have the documentation inside the solution. I saw you deleted the MUnique.OpenMU.Documentation.csproj (I'm fine with that), but it would be great to have somehow access to its files.

And a more general suggestion about pull requests: Next time, please create a branch for your changes. A pull request from master into master is a bit problematic, since you can't change your master branch without affecting the pull request.

sven-n:
Closed due missing feedback and merge conflicts.

PR #100, opened 2019-03-13
Update PipelinedSimpleModulusEncryptor.cs

wplatform:
fix contentOfFirstBlockLength bug


sven-n:
Thanks, I’ll look at it soon.

sven-n:
Thanks for bringing this bug to my attention, I can't believe I didn't notice it.

I added a unit test to identify the issue and verify the solution. I found that 'size' has already the correct value, so nothing has to be added or subtracted at it.

Issue #111, opened 2019-05-09
Packet handler and view plugins for 0.75


sven-n:
The version 0.75 is very old and contains just a subset of the features of this server.
Some packet structures might not have been changed from 0.75 to season 6, so there is probably not much work to do. From a historical point of view it’s also interesting how messages passed between client and server got changed or extended over time. The simplicity of the early server also might reveal missing but important messages in season 6.

Another thing to note: If a message structure of season 6 from server to client contains additional fields (but rest is valid), it’s not required to remove these fields for 0.75. The client would just not map these fields to a struct field.

I analyzed what needs to be changed. The following list can be seen as a to do list, too. Italic entries were supported initially.

General:
  * [x] Appearance Serializer
  * [x] Item Serializer

Packet Handlers:
  * [x]  0xF1 - Login Group
    * [x] 0x01 - Login Request
    * [x] *0x02 - Log out Request*
  * [x] *0xF3 - Character Group*
    * [x] *0x00 - Request Character List*
    * [x] *0x01 - Character Create Request*
    * [x] *0x02 - Character Delete Request*
    * [x] *0x03 - Character Selection Request*
    * [x] *0x06 - Add level up point*
    * [x] *0x12 - Data loaded*
  * [x] *0x00 - Chat message*
  * [x] *0x01 - Whisper message*
  * [x] 0x10 - Move/walk
  * [x] 0x11 - Set position (e.g. by skills)
  * [x] 0x15 - Hit
  * [x] *0x18 - Animation*
  * [x] 0x19 - Skill
  * [x] 0x1C - Teleport
  * [x] 0x1D - AreaSkillHit
  * [x] 0x1E - Area Skill
  * [x] *0x22 - Item pickup request*
  * [x] *0x23 - Item drop request*
  * [x] *0x24 - Item move request*
  * [x] *0x26 - Item use/consume request*
  * [x] *0x30 - Talk npc request*
  * [x] *0x32 - Item buy request*
  * [x] *0x33 - Item sell request*
  * [x] *0x36 - Trade request*
  * [x] *0x37 - Trade request response*
  * [x] *0x3A - Trade money*
  * [x] *0x3C - Trade button state change*
  * [x] *0x3D - Trade finished*
  * [x] *0x40 - Party request*
  * [x] *0x41 - Party request response*
  * [x] *0x42 - Party list request*
  * [x] *0x43 - Party member kick*
  * [x] *0x50 - Guild request*
[...]

PR #114, opened 2019-05-14
Version 0.75 Support (Packet Handlers and Views)

sven-n:
This includes all changes which are required to adapt messages between the 0.75 client and the server.
It's work in progress (draft) - see #111 for a detailed progress summary.


sven-n:
Will merge this PR soon because it contains a lot of new stuff which will be required for 0.97d, too.
I‘m also thinking about stopping 0.75 development, because the game client seems to have too many bugs :(

PR #122, opened 2019-06-04
Add map requirements

nmarazov:
add AttributeRequirement repository to retrive already generated requirements
updated game logic
add custom SQL in migration Grant select acces for the new mapping table(ATGMD)

things to improve:
- migration 0002 need to remove unused column from AttributeRequirement table


sven-n:
I would like to merge it soon. Do you think there are still open points?

nmarazov:
I don`t think there is. It looks very clean with the extension method.

sven-n:
Thanks! 👍

PR #123, opened 2019-06-07
Pets and fixes for pet skills

sven-n:
During testing of #122, I added some pets to the initialization data.
I also noticed that some skills of pets (Dark Horse, Fenrir) are not working correctly. I fixed them.
Both is included in this PR.


sven-n:
See also #6, but a system for pets which can be leveled up is **not** included in this PR.

sven-n:
Build fails, because merging #122 is required first.

PR #128, opened 2019-07-09
Added missing ClientTimeout column to migrations

CodeKJ:
This fixes initial StartUp error regarding missing ClientTimeout column in ChatServerDefinition table


sven-n:
Thank you!

PR #130, opened 2019-09-15
Generate message structs by data

sven-n:
See also:
  * [Handling packet structures in .NET - a new way?](https://munique.net/handling-packet-structures-dotnet/)
  * [Generating message structs by data](https://munique.net/generating-message-structs-by-data/)

For the data, I basically took the XML files of the network analyzer and extended them for the new requirements.
Then I wrote a XSL transformation which transforms this xml data to C# code.
I set up the transformation to run as a pre-build event, so that it first runs the transformation, then compiles the code. This way, it gave me a fast feedback-loop to adapt/modify the transformation until I was satisfied with the result.
In this process I also began to go through the ClientToServer packets and extended some missing data - this work is still in progress, a lot of messages are still missing.

The plan is, to use this project in the GameServer project and the Network.Analyzer. Later maybe also in the ChatServer and ExDbConnector.


sven-n:
Oh no... build fails because of this issue: https://github.com/Microsoft/msbuild/issues/3368

sven-n:
The CI-Build still fails with .net core 3.0 :-( On my dev machine it builds without problems, though.

```
D:\a\1\s\src\Network\Packets\MUnique.OpenMU.Network.Packets.csproj(33,5): error MSB4062: The "Microsoft.Build.Tasks.XslTransformation" task could not be loaded from the assembly Microsoft.Build.Tasks.Core, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a.  Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.
```

sven-n:
Almost done. There is one last message missing - AddCharacterToScope (0xC2, 0x12). This one is really hard to depict as struct, because each character block might have a variable size. At the end of each block, the protocol specifies the number of effects and then adds the specified number of bytes. Because this is different for each player, we can't easily calculate this beforehand, and also can't easily calculate the position of the n-th player in the struct. We would have to go through all the data in a linear fashion to count these number of effects somehow.

PR #132, opened 2019-09-25
Docker compose

Riicksouzaa:
Let's try up the server and database at same compose.

### DIFF
--- a/QuickStart.md
+++ b/QuickStart.md
@@ -12,27 +12,32 @@ This guide describes two ways of starting the server. Use Docker, if you just wa
 
 ## Docker
 
+This guide assumes you know how to use docker in general and have docker installed (e.g. by using Docker Desktop on Windows).
+
+### Demo Mode
 If you just want to play around with the server, you can find the newest docker image on the Docker Hub:
 https://hub.docker.com/r/munique/openmu
-
-This guide assumes you know how to use docker in general and have docker installed (e.g. by using Docker Desktop on Windows).
 
 To pull and run the latest docker image, run this command:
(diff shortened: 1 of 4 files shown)


sven-n:
Wow, thank you very much. It works great, but I just have one question. I'm not familiar with docker compose and data storage with postgres. Can you tell me where the data is saved? Is it inside the container, or in some external volume?

When I understand this configuration correctly, it's saved outside of the container (that's what we need), isn't it?
```
volumes:
  dbdata:
    driver: local
```

Apart of that, I did some minor adaptions. I introduced some environment variables to make the connection strings more configurable.

Riicksouzaa:
> Wow, thank you very much. It works great, but I just have one question. I'm not familiar with docker compose and data storage with postgres. Can you tell me where the data is saved? Is it inside the container, or in some external volume?
> 
> When I understand this configuration correctly, it's saved outside of the container (that's what we need), isn't it?
> 
> ```
> volumes:
>   dbdata:
>     driver: local
> ```
> 
> Apart of that, I did some minor adaptions. I introduced some environment variables to make the connection strings more configurable.

Ohhhh, you are correctly, actualy the database is stored by docker and docker will do that.

we need to: Create a data directory on the host system (outside the container) and mount this to a directory visible from inside the container. This places the database files in a known location on the host system, and makes it easy for tools and applications on the host system to access the files. The downside is that the user needs to make sure that the directory exists, and that e.g. directory permissions and other security mechanisms on the host system are set up correctly.


something like: ```/my/own/datadir:/var/lib/postgresql/data```


```yml
database:
    image: postgres
    restart: always
    container_name: database
    environment:
      POSTGRES_PASSWORD: admin
      POSTGRES_DB: openmu
      POSTGRES_USER: postgres
    networks:
      - munique-network
    volumes:
      - dbdata:/var/lib/postgresql/data
```
I think something like that gona work fine.

PR #134, opened 2019-09-26
docker-compose fix database not being stored on docker-compose down/up

Riicksouzaa:
Fixes problem when database is not stored when the container dies. (docker-compose down)


sven-n:
Thanks again :)

PR #136, opened 2019-10-02
Implemented automatic database migration for Docker deployment

### LINKED ISSUE #133: Automatically applying EF Core Migrations when using Docker is not possible
Assume that we have the following scenario:
  * An existing openmu database
  * An updated openmu docker image with a new entity framework core migration

In our Startup project, in the Program.cs, we request input from the console in this case:
```
            else if (!manager.IsDatabaseUpToDate())
            {
                Console.WriteLine("The database needs to be updated before the server can be started. Apply update? (y/n)");
                var key = Console.ReadLine()?.ToLowerInvariant();
                if (key == "y")
                {
                    manager.ApplyAllPendingUpdates();
                    Console.WriteLine("The database has been successfully updated.");
                }
                else
                {
                    Console.WriteLine("Cancelled the update process, can't start the server.");
                    return null;
                }
            }
```

The requirement of manual intervention is probably bad in this case, since it's not obvious when running in a container.

Possible quick solution: Add another parameter, e.g. "-autoupdate" to the code, which is also added to the Dockerfile when starting the server process.

Renerick:
Closes #133

### DIFF
--- a/src/Startup/Program.cs
+++ b/src/Startup/Program.cs
@@ -56,7 +56,7 @@ namespace MUnique.OpenMU.Startup
             }
             else
             {
-                this.persistenceContextProvider = this.PrepareRepositoryManager(args.Contains("-reinit"));
+                this.persistenceContextProvider = this.PrepareRepositoryManager(args.Contains("-reinit"), args.Contains("-autoupdate"));
             }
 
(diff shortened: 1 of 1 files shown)


sven-n:
Thank you very much :)

Happy Hacktoberfest 🎉

PR #137, opened 2019-10-10
Upgrade to .NET Core 3.0 and Standard 2.1

sven-n:
Tasks:
- [x] Adapted documentation
- [x] Upgraded to newest Entity Framework Core
- [x] Upgraded all other nuget packages (except Mapster, need to test first)
- [x] Changed all Windows.Forms-Tools to .NET Core, too
- [x] Adapted Dockerfile
- [x] Did a quick test if it starts and the AdminPanel shows up

- [x] Log Messages on the AdminPanel are missing
  The SignalRAppender does not seem to forward the log messages to the connected clients correctly.
- [x] Test connecting with a game client
- [x] Fix T4 Template transformation


sven-n:
While debugging, I had a lot of these messages:
```
Exception thrown: 'System.IO.InvalidDataException' in Microsoft.AspNetCore.SignalR.Protocols.Json.dll
Exception thrown: 'System.NotSupportedException' in System.Text.Json.dll
```

So, I dig deeper and found the code throws the following System.NotSupportedException:
"Deserialization of reference types without parameterless constructor is not supported. Type 'MUnique.Log4Net.CoreSignalR.LogEntry'"
So it's more of an issue of the MUnique.Log4Net.CoreSignalR project.

PR #141, opened 2019-11-14
Add param ip to config ip resolver

thinhnotes:


sven-n:
Hi,
Thanks for your pull request. I just had a quick look at it and it seems to solve the issue.

However, I would have the following suggestions to improve it:
  * Instead of two parameters (-local and -ip) one parameter would be better imho. For example:
    * -ip:local
    * -ip:public
    * -ip:192.168.0.20
  * Instead of putting the new logic into the existing LocalIpResolver, there should be a new implementation for IIpAddressResolver. Depending on the part behind „ip:“ the right implementation could be created/selected.
  * The port of the AdminPanel has been changed. Is there a reason for it: should we add another parameter?
  * The readme of the Startup project needs to be updated as well

thinhnotes:
Whould you like to provide a param of port admin panel for it, example -adminport:1234

Issue #147, opened 2020-01-21
Use Blazor instead of React in the AdminPanel project


MarioGK:
Hi, i have been away from the project for quite a while, but i am still coming back to know the progress from time to time.

Could we not use Blazor instead of React for the MainWebsite and the AdminPanel?
afaik you want to keep everything C# based.

sven-n:
Hi, welcome back :)
Yes, I planned with Blazor, too. I would use Blazor for the AdminPanel and just normal Razor for the web site. Unfortunately, my time for this Project is pretty limited at the moment :-/
I would like to refactor the composition of the whole project first, also for a web api (#125).

I'm also thinking about how much scope this project should have, as there are websites which could be adapted to be compatible (e.g. WebEngine) and it's a big topic of its own.

MarioGK:
Well in March the Client Side Blazor will be released, that would be perfect for the MainWebsite(if you want to do that).
The server side version would be perfect for the admin panel, you have access to the server without any kind of API.

I have not tried to help more because your coding style is really different from mine, if you are disposed to help me out on fixing my style and some bugs, i can try start doing the WebAPI(with this finished i can port my Launcher to .Net Core 3 and resume working on it as well) and the Websites.

PR #154, opened 2020-02-07
Implementation of the admin panel with server side blazor

sven-n:
See also #147


sven-n:
Build fails because of this error:
```
2020-02-11T06:22:40.2240692Z        (WebCompile target) -> 
2020-02-11T06:22:40.2240847Z          /home/runner/.nuget/packages/buildwebcompiler/1.12.405/build/BuildWebCompiler.targets(12,9): error : No such file or directory [/home/runner/work/OpenMU/OpenMU/src/AdminPanelBlazor/MUnique.OpenMU.AdminPanelBlazor.csproj]
```

sven-n:
I worked on the live map, and the main stuff has been done. It shows a map and the player list. However, it currently has some issues:
* Objects (Players, NPCs) are not shown on the map. There are no errors on the server and at the browser side and the right functions are called in the right way (I think). I added log entries on the browser side to confirm that.
* The layout is not so great:
  * The FPS display should integrate somewhere without taking additional space, depending on the device width.
  * When resizing the window, or generally starting, the player list may not be visible.

sven-n:
A lot of improvements have been made. It's now almost ready to be merged.

PR #155, opened 2020-02-20
Create С2F4 - Server list response (by server).md

Tocher:


sven-n:
Thanks :) I'll need to do some minor adjustments, then I'll merge it. It may take a few days.

Tocher:
i'm trying create unreal engine 4 mu client based on mu protocol, so will  be commiting  some packet descriptions, but for this i think i missed with server load info ;)

sven-n:
Nice :)
However, I have to tell you, these packet descriptions will get obsolete sooner or later. I plan to generate them based on the xml files of the Network.Packets project:  https://github.com/MUnique/OpenMU/tree/master/src/Network/Packets

If you are motivated, you can add a new xml file which describes the connect server packets. The goal is to have data about all kind of packets between the client and all kind of servers. Currently, only packets between the client and the game server are defined there.

sven-n:
Hi, before you spend more time in this issue, I have some news. I‘ve put a packet description of the connect and chat server together and implemented a xsl transformation to automatically create markdown documentation files. It’s not 100 % done yet, but I plan to push it within the next week.

Issue #161, opened 2020-03-06
[GameLogic] /item chat command


jhonynet:
For all kind of testing propuse i've been working on `/item` command.

`/item Group Number Level Durability Skill Exc Opt Ancient`

sven-n:
Nice 👍🏻 Had this in my big todo list, too 😄

jhonynet:
In the first version of the command the arguments will be the followings
/item {Group} {Number} {Level} {Exc} {Skill} {Luck} {Opt}

Group (byte): Item Group
Number (sort): Item Number 
Level (byte): **0-15**
*Exc (byte): Limit the ItemOptions(excellent) collection (sorting comes from db engine) **0-N**
*Skill (bool): Add skill to item (excellent override this value) **0-1**
*Luck (bool): Add luck to item **0-1**
*Opt (byte): add option to item **0-4**

(*) Optional Arguments

next release may contain socket and ancient options?

PR #163, opened 2020-03-09
[Enhancement] /item command

jhonynet:
Details of this command are already explained in the issue.

https://github.com/MUnique/OpenMU/issues/161

### DIFF
--- a/src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
+++ b/src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
@@ -106,7 +106,6 @@ namespace MUnique.OpenMU.GameLogic.PlugIns
 
             if (arguments.Exc > 0)
             {
-
                 var excellentOptions = item.Definition.PossibleItemOptions.SelectMany(o => o.PossibleOptions)
                     .Where(o => o.OptionType == ItemOptionTypes.Excellent)
                     .Take(arguments.Exc);
--- a/src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
+++ b/src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
@@ -2,18 +2,17 @@
 // Licensed under the MIT License. See LICENSE file in the project root for full license information.
 // </copyright>
 
-using System.Collections.Generic;
-using MUnique.OpenMU.DataModel.Configuration.Items;
-
 namespace MUnique.OpenMU.GameLogic.PlugIns
 {
-    using System.Runtime.InteropServices;
-    using MUnique.OpenMU.PlugIns;
     using System;
+    using System.Collections.Generic;
     using System.Linq;
+    using System.Runtime.InteropServices;
+    using MUnique.OpenMU.DataModel.Configuration.Items;
     using MUnique.OpenMU.DataModel.Entities;
     using MUnique.OpenMU.GameLogic.Views;
     using MUnique.OpenMU.Interfaces;
+    using MUnique.OpenMU.PlugIns;
 
     /// <summary>
     /// A chat command plugin which handles item creation command.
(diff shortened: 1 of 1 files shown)


src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
```diff
@@ -0,0 +1,146 @@
+﻿// <copyright file="CreateItemChatCommandPlugIn.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+using System.Collections.Generic;
+using MUnique.OpenMU.DataModel.Configuration.Items;
```
sven-n: usings should be inside the namespace

src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        private static Item CreateItem(Player player, Arguments arguments)
+        {
+            var item = new TemporaryItem();
+            var itemDefinition = player.GameContext.Configuration.Items.FirstOrDefault(def => def.Group == arguments.Group && def.Number == arguments.Number);
+            if (itemDefinition == null)
+            {
+                throw new ArgumentException($"[GM][/item] {arguments.Group} {arguments.Number} does not exists");
+            }
+            item.Definition = itemDefinition;
```
sven-n: please add one empty line here after the closing bracket

src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
```diff
@@ -0,0 +1,146 @@
...
+                throw new ArgumentException(Usage);
+            }
+
+            return new Arguments()
+            {
+                Group = byte.Parse(argumentsList.ElementAt(0)),
+                Number = short.Parse(argumentsList.ElementAt(1)),
+                Level = byte.Parse(argumentsList.ElementAtOrDefault(2) ?? "0"),
+                Exc = byte.Parse(argumentsList.ElementAtOrDefault(3) ?? "0"),
+                // Every excellent item has Skill
```
sven-n: Every excellent item has Skill, but only if a Skill is available in the ItemDefinition

src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
```diff
@@ -0,0 +1,146 @@
...
+                        .First(o => o.OptionType == ItemOptionTypes.Luck)
+                };
+                item.ItemOptions.Add(optionLink);
+            }
+
+            if (arguments.Exc > 0)
+            {
+                var excellentOptions = item.Definition.PossibleItemOptions.SelectMany(o => o.PossibleOptions)
+                    .Where(o => o.OptionType == ItemOptionTypes.Excellent)
+                    .Take(arguments.Exc);
```
sven-n: I think ```Take``` here isn't correct. The Parameter is probably meant as a bit-mask, like a flags enum?
If this is the case you'd need to do smth like this:
```
.Where(o => (o.Number & arguments.Exc) > 0)
```

src/GameLogic/PlugIns/CreateItemChatCommandPlugIn.cs
```diff
@@ -0,0 +1,146 @@
...
+                        ItemOption = option
+                    };
+                    item.ItemOptions.Add(optionLink);
+                });
+            }
+
+            return item;
+        }
+
+        private static Arguments MarshallArguments(List<string> argumentsList)
```
sven-n: I'd name the method ```ParseArguments```

jhonynet:
All fixed plus Codacy interface suggestion and code style

Regards!!

sven-n:
Great, Thank you :)

PR #164, opened 2020-03-10
[Enhancement] Move chat commands to folder

jhonynet:
due the long list of future commands i think is a good idea to move to separated folder

### DIFF
--- a/src/GameLogic/PlayerActions/ChatMessageAction.cs
+++ b/src/GameLogic/PlayerActions/ChatMessageAction.cs
@@ -11,6 +11,7 @@ namespace MUnique.OpenMU.GameLogic.PlayerActions
     using log4net;
     using MUnique.OpenMU.GameLogic.PlugIns;
     using MUnique.OpenMU.GameLogic.Views;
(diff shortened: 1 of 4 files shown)


jhonynet:
I can't see the failed log :S

sven-n:
Me too, I’m not on my desk atm :(
Have you tried building the whole solution?

Sorry for the inconvenience

jhonynet:
Yes, it builds in my local.
Can we re-dispatch checks from here without any push?

sven-n:
Error in the Azure Pipeline Log:
```
PlugIns\CreateItemChatCommandPlugIn.cs(26,48): error CS0246: The type or namespace name 'IChatCommandPlugIn' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\src\GameLogic\MUnique.OpenMU.GameLogic.csproj]
```

sven-n:
I think the original CreateItemChatCommandPlugIn.cs wasn't committed as moved or deleted, that's why the error occurs

jhonynet:
Ok, now with the log i can fix it
Thanks you!

PR #167, opened 2020-03-24
[Enhancement] /move command

jhonynet:
The title explains everything...

should be any feedback when desired map does not exists?


sven-n:
> should be any feedback when desired map does not exists?

I think it's okay as it is.

Good extension, thank you :)

PR #176, opened 2020-04-07
Direction to target FIX

And3rsL:


sven-n:
Thanks, I‘ll try it out soon :)

sven-n:
Thanks, it looks much better now.
Monsters are still attacking to the wrong direction, but I think that's caused by another bug. This happens when the Monster has the same internal coordinates as the player, which the Game Client doesn't show.

Issue #177, opened 2020-04-07
Add Monster Spell


And3rsL:
Some Monsters can cast spells
Example: Dungeon WIzard (and lorencia) can cast lightining

Currently its not supported

sven-n:
Yes, it's not implemented yet. Some Monsters can even cast multiple spells, which we would need to consider when implementing it.

sven-n:
I wrote a Regex to find monsters with skills in a Monsters.txt config file:
```regex
^(\d+)\s+\d+\s+"[^"]+"\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+([1-9]\d*)
```
First group is the monster number, second group is the skill number.

PR #178, opened 2020-04-09
Trap Support

And3rsL:
It creates basic trap support, at least they dont walk around :)


And3rsL:
i just noticed you changed somethings in EntityDataContextModelSnapshot, you should fix it before merge, but add TrapType

sven-n:
Thanks, I’ll check it soon.
I somehow remember a similar issue a long time ago - will try to find it ;)

sven-n:
I'm just checking your PR. It works, but I'd like to change how the correct behavior is defined in the data and created in the logic.
I found a flaw in the initial code design, which I'd like to fix with this extension, as well. Currently, we have just an interface "IAttackable" for objects which can receive attacks, but also perform attacks. In case of some objects, this is not sufficient. Traps can't be attacked, but only attack. (Castle) Gates or Statues can only be attacked, but not attack.
So, I'd like to introduce the interface IAttacker. Since this involves a lot of code changes, I'll do it myself. Once I'm done, I'll push it into this PR :)

sven-n:
@And3rsL have you got some attack animations and working attacks with your changes?
I found that some traps (Lost Tower, Meteorite Traps) seem to cause damage, but I see no animations :-(
And some other traps don't attack at all (Fire Traps, Dungeon).

I put a lot of changes into this PR. However, I think some things in the attack logic still needs some fine tuning ;-)

And3rsL:
I think meteorite traps and all area attack in general just cast spells.. but maybe I'm wrong

sven-n:
Two traps cast a flame skill according to the Monsters.txt, others don't.
So, I think it makes sense to solve #177 within this PR here, too.

sven-n:
I think I got it working :) I get hits and skill animations from the traps and monsters.
Could you do a quick test in game?

PR #179, opened 2020-04-10
[Enhancement] reset system

jhonynet:


jhonynet:
Do not merge yet, is only for build check

sven-n:
The build pipeline has an issue. Nuget restore is broken somehow :( Not your fault ;)

sven-n:
Thanks for your contribution!
I made some minor improvements at the code and will merge it.

It would be great if this system is somehow configurable. I have to admit, there is currently no other way than extending the DataModel. We probably need an easy system to store plugin configurations independently from that. I have some ideas and need more time to think this through, though. But for now, the current state should be good enough to merge.

PR #180, opened 2020-04-13
[Fix] reset system messages from the npc

jhonynet:
Show message from NPC (like leo the helper) instead in the chat.


sven-n:
Nice idea 👍🏻

PR #183, opened 2020-04-19
[Enhancement] Drop zen

jhonynet:
This is ready but tests are not modified, so it might fail.
drops zen by default


sven-n:
Thanks again for your contributions :)

PR #192, opened 2020-08-29
Fixed Network Analyzer publishing

guser9822:
I was playing with the Network Analyzer tool and I wanted to run it as  an application outside VS, so I tried to publish it but I got this error :

image

I found the solution in this [stackoverflow post](https://stackoverflow.com/questions/43320772/the-publish-target-is-not-supported-without-specifying-a-target-framework), but I'm not really sure it is the correct one. However now everything works fine

image


sven-n:
Thank you! 👍🏻😃

PR #201, opened 2020-11-09
Allow Proxied GS

jhonynet:
i've done this to allow to define custom GS port to be registered in CS so the network analyzer could be reached by the client.
I don't know if this is right or there is another way to do this but maybe is usefull


sven-n:
Nice idea! 👍
I renamed it to be more generic, because I identified more use cases than just using a network proxy.
I extended the description accordingly.

PR #213, opened 2021-05-05
Feature/walk fixes

muonlinedev:
**Description**
This PR fixes steps/directions computation and consumption when the walking path to the target location is not a straight line.

**How to reproduce the issue**
1. Log in to any character and go to Lorencia, position (137, 122).
2. Switch character and log back in to the same character and check the current position.
3. If the position is different from (137, 122), then the issue was reproduced successfully.
4. If not, go to (136, 124) and switch character and log back in. The current position will be (136, 123), instead of (136, 124).


muonlinedev:
Great!
Thank you too!

PR #217, opened 2021-07-13
Fix the problem that reflection damage is over-calculated

najoast:
When `CurrentShield` is not enough, `oversd` will be added to `HealthDamage`. When the reflection damage is calculated finally, this part of the overflow damage is also included, causing the reflection to become more than the actual value.


sven-n:
Thanks!

PR #241, opened 2022-01-30
Doc/add client

caiobessa:
When I was starting the setup of the project I had some difficulty finding a compatible client.

I think would be useful to have the link of a client in the documentation.


sven-n:
Hey there, thanks for your input :)

Unfortunately, I can't merge your pull request. The reason is, I don't want to risk the repository with a link to copyright protected binaries.
However, I adapted the QuickStart guide with a little hint where you can find one.

caiobessa:
Hey @sven-n yeah it make sense. The quick start now is better thanks ;)

PR #243, opened 2022-05-07
Added track chat command

addRange:


src/GameLogic/PlugIns/ChatCommands/TrackChatCommandPlugIn.cs
```diff
@@ -0,0 +1,46 @@
+﻿// <copyright file="TraceChatCommandPlugIn.cs" company="MUnique">
```
sven-n, line 1: Please fix the file name here ;-)

addRange: Done

PR #265, opened 2022-06-14
Increase AddStat commmand to handle unlimited points for GameMaster

ja0n:
this change turned out helpful for testing things out

### DIFF
--- a/src/GameLogic/PlayerActions/Character/IncreaseStatsAction.cs
+++ b/src/GameLogic/PlayerActions/Character/IncreaseStatsAction.cs
@@ -24,18 +24,31 @@ public class IncreaseStatsAction
             throw new InvalidOperationException("No character selected");
         }
 
-        if (player.SelectedCharacter.LevelUpPoints > 0)
+        var selectedCharacter = player.SelectedCharacter;
+        if (!selectedCharacter.CanIncreaseStats())
         {
-            var attributeDef = player.SelectedCharacter.CharacterClass?.StatAttributes.FirstOrDefault(a => a.Attribute == statAttributeDefinition);
-            if (attributeDef is { IncreasableByPlayer: true })
-            {
-                player.Attributes![attributeDef.Attribute]++;
-                player.SelectedCharacter.LevelUpPoints--;
-                player.ViewPlugIns.GetPlugIn<IStatIncreaseResultPlugIn>()?.StatIncreaseResult(statAttributeDefinition, true);
-                return;
-            }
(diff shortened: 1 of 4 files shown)


src/GameLogic/PlayerActions/Character/IncreaseStatsAction.cs
```diff
@@ -24,16 +24,19 @@ public void IncreaseStats(Player player, AttributeDefinition statAttributeDefini
...
         {
-            var attributeDef = player.SelectedCharacter.CharacterClass?.StatAttributes.FirstOrDefault(a => a.Attribute == statAttributeDefinition);
-            if (attributeDef is { IncreasableByPlayer: true })
-            {
-                player.Attributes![attributeDef.Attribute]++;
-                player.SelectedCharacter.LevelUpPoints--;
-                player.ViewPlugIns.GetPlugIn<IStatIncreaseResultPlugIn>()?.StatIncreaseResult(statAttributeDefinition, true);
-                return;
-            }
+            return;
```
sven-n, line 30: Because of this return, no result is sent anymore.

src/GameLogic/PlayerActions/Character/IncreaseStatsAction.cs
```diff
@@ -24,16 +24,19 @@ public void IncreaseStats(Player player, AttributeDefinition statAttributeDefini
             throw new InvalidOperationException("No character selected");
         }
 
-        if (player.SelectedCharacter.LevelUpPoints > 0)
+        var selectedCharacter = player.SelectedCharacter;
+        if (selectedCharacter.CharacterStatus != CharacterStatus.GameMaster && player.SelectedCharacter?.LevelUpPoints <= 0)
```
sven-n, line 28: Maybe introduce an extension method for a character, which determines, if a player can add points. So that the same check can be used in the AddStatChatCommandPlugIn

ja0n: Fixed

src/GameLogic/PlayerActions/Character/IncreaseStatsAction.cs
ja0n: Fixed

src/GameLogic/PlayerActions/Character/IncreaseStatsAction.cs
ja0n: Yeah, good call. Done

PR #266, opened 2022-06-19
Fix GuildName length on GuildCreateRequest packet

ja0n:
It's breaking the guild creation for names with 8 characters since it's adding an extra noise character


juliolitwin:
It's necessary to fix in the XML. That's file is auto generated.

PR #282, opened 2022-11-23
Fix build project on MacOS

afrokick:
- [x] add `EnableWindowsTargeting=true` https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/netsdk1100
- [x] use `npx` to generate packets' docs


afrokick:
we can use `npx` (util from npm) to generate docs like this:

```bash
#!/bin/bash

rm -rf ../../../docs/Packets/C*.md

npx xslt3 --no-install -xsl:GenerateMarkdownDocs.xslt -o:../../../docs/Packets/ClientToServer.md -s:ClientToServer/ClientToServerPackets.xml
npx xslt3 --no-install -xsl:GenerateMarkdownDocs.xslt -o:../../../docs/Packets/ServerToClient.md -s:ServerToClient/ServerToClientPackets.xml
npx xslt3 --no-install -xsl:GenerateMarkdownDocs.xslt -o:../../../docs/Packets/ConnectServer.md -s:ConnectServer/ConnectServerPackets.xml
npx xslt3 --no-install -xsl:GenerateMarkdownDocs.xslt -o:../../../docs/Packets/ChatServer.md -s:ChatServer/ChatServerPackets.xml
```

You don't need to manually install software like `SaxonHE9.9N` but we need NodeJS (it is more chance that users have NodeJS)

@sven-n What do you think?

afrokick:
I need a help to solve an issue this EOL for `xslt3` library. Does anyone know how to configure it?

sven-n:
Hey, if that works, it would be great!

Can you explain what's the issue about EOL?

sven-n:
Hmmm, actually I can't see the changed EOL on my machine - BUT I have probably changed how git handles line endings on my machine. I have set core.autocrlf to true...
See also: https://stackoverflow.com/questions/4181870/git-on-windows-what-do-the-crlf-settings-mean

afrokick:
@sven-n you'r right! Thanks!

PR #292, opened 2022-11-30
fix drop chance generation

afrokick:
If we have two or more groups with 1.0 chance, all except first will never be selected.

```csharp
double lot = 1; //or 0

for(group of groups) {
  // ALWAYS BE FALSE when lot = 0..1 !!!
  if(lot > group.Chance) {
    lot -= group.Chance;
  } else {
    return group;
  }
}
```

### DIFF
--- a/src/GameLogic/DefaultDropGenerator.cs
+++ b/src/GameLogic/DefaultDropGenerator.cs
@@ -402,7 +402,7 @@ public class DefaultDropGenerator : IDropGenerator
 
     private DropItemGroup? SelectRandomGroup(IEnumerable<DropItemGroup> dropGroups)
     {
-        var sum = dropGroups.Sum(group => group.Chance);
+        var sum = Math.Max(1.0, dropGroups.Sum(group => group.Chance));
         double lot = this._randomizer.NextDouble() * sum;
 
         foreach (var group in dropGroups)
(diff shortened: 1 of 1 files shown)


src/GameLogic/DefaultDropGenerator.cs
```diff
@@ -402,7 +402,9 @@ private void AddRandomExcOptions(Item item)
 
     private DropItemGroup? SelectRandomGroup(IEnumerable<DropItemGroup> dropGroups)
     {
-        double lot = this._randomizer.NextDouble();
+        var sum = dropGroups.Sum(group => group.Chance);
+        double lot = this._randomizer.NextDouble() * sum;
```
afrokick, line 406: Now it will be from 0 to SUM

src/GameLogic/DefaultDropGenerator.cs
```diff
@@ -402,7 +402,9 @@ private void AddRandomExcOptions(Item item)
 
     private DropItemGroup? SelectRandomGroup(IEnumerable<DropItemGroup> dropGroups)
     {
-        double lot = this._randomizer.NextDouble();
+        var sum = Math.Max(1.0, dropGroups.Sum(group => group.Chance));
```
afrokick, line 405: Is it OK to have sum greater than 1?

It is a case when you have 'rare items' group with chance 0.000001 and it shouldn't drop all the time. Math.Max fix it.

sven-n:
I need some time to think about that 😅

Maybe it makes sense to always choose a group (at SelectRandomGroup), when it has 1.0 as chance defined?
Do you think there is more than 1 group for special monsters like the golden ones?

afrokick:
@sven-n Agree, I think one group should be enough for 99% game play features.

Is #293 a good solution?

afrokick:
Closed because #293 solve this issue

PR #293, opened 2022-12-01
Change the order of DropItemGroup

afrokick:
Another approach (the first #292) to solve a problem with 1.0 chance for some group.

### DIFF
--- a/src/GameLogic/DefaultDropGenerator.cs
+++ b/src/GameLogic/DefaultDropGenerator.cs
@@ -63,7 +63,7 @@ public class DefaultDropGenerator : IDropGenerator
                     map.DropItemGroups,
                     questGroups)
                 .Where(group => IsGroupRelevant(monster, group))
-                .OrderBy(group => group.Chance);
+                .OrderByDescending(group => group.Chance);
 
         IList<Item>? droppedItems = null;
         for (int i = 0; i < monster.NumberOfMaximumItemDrops; i++)


src/GameLogic/DefaultDropGenerator.cs
```diff
@@ -63,7 +63,7 @@ public IEnumerable<Item> GenerateItemDrops(MonsterDefinition monster, int gained
                     map.DropItemGroups,
                     questGroups)
                 .Where(group => IsGroupRelevant(monster, group))
-                .OrderBy(group => group.Chance);
+                .OrderByDescending(group => group.Chance);
```
sven-n, line 66: I understand your change, but I have the feeling that it will cause, that groups with a low chance will never drop, once the total sum gets over 1.
However, it's difficult to handle the new logic in an optimal way with the previous OrderBy, too. 👀
I need some time to think about that 😅

sven-n:
What do you think about the following approach?
```csharp
        IEnumerable<DropItemGroup> dropGroups;
        if (monster.DropItemGroups.MaxBy(g => g.Chance) is { Chance: >= 1.0 } alwaysDrops)
        {
            dropGroups = alwaysDrops.GetAsEnumerable();
        }
        else
        {
            dropGroups = monster.ObjectKind == NpcObjectKind.Destructible
                ? monster.DropItemGroups
                : CombineDropGroups(
                        monster.DropItemGroups,
                        character.DropItemGroups,
                        map.DropItemGroups,
                        questGroups)
                    .Where(group => IsGroupRelevant(monster, group))
                    .OrderBy(group => group.Chance);
        }
```

afrokick:
Update PR with your solution @sven-n

PR #294, opened 2022-12-01
fix MonsterAttributeHolder concurrent access

afrokick:


src/GameLogic/Attributes/MonsterAttributeHolder.cs
```diff
@@ -163,7 +164,11 @@ public IElement GetOrCreateAttribute(AttributeDefinition attributeDefinition)
             result = monsterDef.Attributes.ToDictionary(
                 m => m.AttributeDefinition ?? throw Error.NotInitializedProperty(m, nameof(m.AttributeDefinition)),
                 m => m.Value);
-            MonsterStatAttributesCache.Add(monsterDef, result);
+
+            if (!MonsterStatAttributesCache.TryAdd(monsterDef, result))
+            {
+                return GetStatAttributeOfMonster(monsterDef);
```
afrokick, line 170: I don't know how to properly return only one instance per key. Maybe there is more elegant way for it.

sven-n: I think ConcurrentDictionary.GetOrAdd instead of TryGetValue and TryAdd might be better in this case

afrokick: @sven-n done

sven-n:
Thanks again :)

Issue #298, opened 2022-12-06
[Feature] Red Dragon invasion


afrokick:
Can be implemented based on #291

PR #300, opened 2022-12-07
[Feature] Add red dragon invasion plugin

afrokick:
An implementation for #298.

I want to configure it with parameters from #299.


sven-n:
Thanks again for the PR :)

The GoldenInvasionPlugIn and the RedDragonInvasionPlugIn are pretty similar and just differ in some details. You could create a base class for the same stuff, and inherit from it.

afrokick:
@sven-n ready for review.

Issue #297, opened 2022-12-06
[Feature] Happy Hour event


afrokick:
**Description**
 Increase an experience rate for some period.

**An example**
Every 6 hours exp rate for all maps increased by +15% for all players. Duration: 30min.

afrokick:
Implemented

PR #302, opened 2022-12-07
[Feature] Happy Hour event

afrokick:
An implementation for #297

Increase exp rate for all maps by X.

~~Needs some refactoring, but will do it after #300.~~

UPD: It works!
Screenshot 2023-07-27 at 20 15 14


src/GameLogic/PlugIns/InvasionEvents/HappyHourPlugIn.cs
```diff
@@ -0,0 +1,220 @@
...
+        await state.Context.ForEachPlayerAsync(p => this.TrySendStartMessageAsync(p)).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Calls after the state changed to Started.
+    /// </summary>
+    /// <param name="state">State.</param>
+    protected virtual async ValueTask OnStartedAsync(GameServerState state)
+    {
+        state.Context.Configuration.ExperienceRate = state.NewExperienceRate;
```
afrokick, line 209: @sven-n

I know several ways:
- GameConfiguration
- MapDefiniton
- Character expRate attribute

Which one more suitable for this task?

src/GameLogic/PlugIns/InvasionEvents/HappyHourPlugIn.cs
```diff
@@ -0,0 +1,220 @@
...
+
+namespace MUnique.OpenMU.GameLogic.PlugIns.InvasionEvents;
+
+using System.Collections.Concurrent;
+using System.Runtime.InteropServices;
+using MUnique.OpenMU.GameLogic.NPC;
+using MUnique.OpenMU.GameLogic.Views;
+using MUnique.OpenMU.PlugIns;
+
+public class HappyHourConfiguration : PeriodicInvasionConfiguration
```
afrokick, line 13: Will refactor it, because we need base class like `PeriodicEventConfiguration`

Another part - GameServerState

afrokick:
@sven-n Need your help!

I have an issue: When player enters the map, he receives the message, because the plugin handle `ObjectAddedToMap`. 
Screenshot 2023-07-27 at 20 16 21

Which event/interface should I implement to send message only when player enter the game(first map)?

sven-n:
Hey, I'm not at my machine atm, but probably IPlayerStateChangedplugin or similar. You could then check if the old state was the character selection and the new one is EnteredWorld

afrokick:
I found `IPlayerStateChangedPlugIn` Thank you!

afrokick:
@sven-n ready for review

sven-n:
Thanks, I made some minor changes, can you check if they make sense for you? :)

afrokick:
@sven-n Yeah, I see. I think it is more clear now with the element. Thanks for the cool solution!

PR #303, opened 2022-12-08
#268

### LINKED ISSUE #268: Character respawns at wrong place
**Describe the bug**
When a character dies, it respawns at the wrong place. For example at Devias 2 instead of Devias town.

**To Reproduce**
1. Enter the game with a (weak) character
2. Go to Icarus or Devias and let it die
3. It respawns at Devias 2

**Expected behavior**
It should respawn at Devias town

**Additional context**
To determine where a character respawns, the ExitGate has a property called "IsSpawnGate". The problem is, for the respawn logic it's basically true for too many gates - all which are listed in the warp menu.

Possible solution: Set it to true just for the safezone gates at the towns. Change the comment which describes the property.
They're initialized at _MUnique.OpenMU.Persistence.Initialization.Version*.Gates_.

Mosch0512:
Close #268 

I think the respawn with Devias 2 should be fixed now.
As of everything i was able to reproduce.

### DIFF
--- a/src/Persistence/Initialization/Version075/Gates.cs
+++ b/src/Persistence/Initialization/Version075/Gates.cs
@@ -109,27 +109,33 @@ public class Gates : InitializerBase
     {
         var targetGates = new Dictionary<short, ExitGate>();
 
+        // Lorencia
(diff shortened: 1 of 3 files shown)


sven-n:
Look good, thank you :)

PR #305, opened 2022-12-08
Add Custom init spawns for Icarus/Karutan2

### LINKED ISSUE #304: Wrong Respawn Locations
**Describe the bug**
If you die in Icarus you will be warped to Lorencia instead of Lost Tower 1.
If you die in Karutan 2 you will be warped to Lorencia after #268 or Karutan 2 again with current Master branch.

Wrong spawns

1. Santa Village
2. Raklion Boss
3. Icarus
4. Karutan 2

**To Reproduce**
1. Go to Icarus/Karutan 2
2. Get Killed
5. Respawned at Lorencia

**Expected behavior**
Respawn at Lost Tower 1/ Karutan 2

**Additional context**
#268 has nothing to do with that

Mosch0512:
Pls check if code is okay i am not used to C#.

Custom spawnpoint initalisation for
1. Icarus
2. Karutan2
3. Santa Village
4. Raklion Boss

close #304

### DIFF
--- a/src/Persistence/Initialization/BaseMapInitializer.cs
+++ b/src/Persistence/Initialization/BaseMapInitializer.cs
@@ -32,7 +32,7 @@ internal abstract class BaseMapInitializer : IMapInitializer
     {
         this.Context = context;
         this.GameConfiguration = gameConfiguration;
-        this._mapDefinition = this.GameConfiguration.Maps.FirstOrDefault(map => map.Number == this.MapNumber && map.Discriminator == this.MapDiscriminator);
+        this._mapDefinition = this.GameConfiguration.Maps.FirstOrDefault(map => map.Number == this.MapNumber && map.Discriminator == this.Discriminator);
     }
 
     /// <summary>
@@ -77,9 +77,9 @@ internal abstract class BaseMapInitializer : IMapInitializer
     protected virtual string TerrainVersionPrefix => string.Empty;
 
     /// <summary>
-    /// Gets the MapDiscriminator of the map definition.
+    /// Gets the discriminator of the map definition.
     /// </summary>
-    protected virtual byte MapDiscriminator { get; }
+    protected virtual byte Discriminator { get; }
 
     /// <summary>
     /// Gets the map number of the safezone map where a player respawns after death.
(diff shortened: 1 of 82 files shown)


src/Persistence/Initialization/BaseMapInitializer.cs
```diff
@@ -127,6 +127,18 @@ public void SetSafezoneMap()
         }
 
         var safezoneMap = this.GameConfiguration.Maps.FirstOrDefault(map => map.Number == this.SafezoneMapNumber);
+        switch (this._mapDefinition.GetName())
```
sven-n, line 130: Well, it will work, but there is a cleaner approach.
Just override the SafezoneMapNumber property in the Icarus and Karutan2 classes ;-)

Mosch0512: thx will do that instead makes more sense xD.

Didnt realize that there is specific map classes yet xD

src/Persistence/Initialization/Version095d/Maps/Icarus.cs
```diff
@@ -30,6 +30,9 @@ public Icarus(IContext context, GameConfiguration gameConfiguration)
     /// <inheritdoc/>
     protected override string MapName => "Icarus";
 
+    /// <inheritdoc/>
+    protected override byte SafezoneMapNumber => 4;
```
sven-n, line 34: Please replace these "magic numbers" by constants.
For example, Lorencia or Devias have internal consts for their numbers. Add constants to LostTower and the other safezone maps as well, and use them instead of these "magic numbers".

Mosch0512: okay makes total sense i am sry <3.

Shouldnt this be checked by the code style analyes also? xD

Mosch0512: hmm i am wondering right now if should use const or static readonly. But i th ink it should be static readonly because it might be inizialised from config files or db or sth xD.

Mosch0512:
If i die in Kanturu Ruins3 where should i get spawned Kanturu Ruins 1 or Kanturu Ruins 3 safe zone?

I think this needs an extra handeling because i dont think there is anything handling death in specific area of specific map to cause different spawn points to get used :)

src/Persistence/Initialization/Version095d/Maps/Icarus.cs
sven-n: I suggest "internal const" in that case, because these values never change ;-)
Usually, you use "static readonly" only for public "constants", or for stuff which can't be defined as a const.

sven-n:
Oh, you're right... I think we might just ignore that for now, until we come up with a good solution. As far as I know, only Kanturu is such a map.

sven-n:
I had a quick look at other server sources, and it seems that if you die in kanturu ruins 3, you'll get respawned at the safezone of kanturu 1, too - so all is good ;-)

Mosch0512:
okay haha :D

Mosch0512:
@sven-n Finished pls have a look <3

Mosch0512:
Okay i checked everything after your changes seems to be fine. Is something still missing? I think its mergable now :D

sven-n:
Yes, it looks good now. Thank you!

Mosch0512:
> Yes, it looks good now. Thank you!

No THANK YOU <3

PR #310, opened 2022-12-23
Randomizing monster attacks to not be simultaneous

Mosch0512:


Mosch0512:
@sven-n i think it is okay like that <3 feels very very good <3

sven-n:
Thanks :)

PR #314, opened 2023-01-25
Added plugins to chat commands to ban by acc and char

marcosdenoni:
/banacc
/unbanacc
/banchar
/unbanchar


sven-n (commented):
Nice, this looks good! One minor thing could be improved:
The code in the plugins looks pretty similar. I think it could be possible to abstract this logic into one or two (abstract) base classes.

sven-n:
Thank you!

PR #316, opened 2023-02-03
Player is online - inform everyone

recone83:
Nice to see when someone joins the game...


sven-n:
I refactored a bit, so it's a plugin which can be deactivated.

PR #320, opened 2023-02-20
Fix for masterlevel level report on client

urshulgi:
I noticed that a fix was implemented on `Player` for this issue, but it wasn't until I added this in `Party` that I started to get the regular level to show up while leveling up; the client wasn't reporting the level up without this fix, it was staying at the level the character had when I logged in.

Example:
I login with MG level 200; I level up to 300 but character screen still shows level 200. I log out, log back in, now it shows level 300 but it stays there.

New:
I login with MG level 200; I level up to 300 and level keeps showing up with experience as I kill monsters.

### DIFF
--- a/src/GameLogic/Party.cs
+++ b/src/GameLogic/Party.cs
@@ -232,10 +232,13 @@ public sealed class Party : Disposable
         var randomizedTotalExperiencePerLevel = randomizedTotalExperience / totalLevel;
         foreach (var player in this._distributionList)
         {
-            if (player.SelectedCharacter?.CharacterClass?.IsMasterClass ?? false)
+            if ((short)player.Attributes![Stats.Level] == player.GameContext.Configuration.MaximumLevel)
             {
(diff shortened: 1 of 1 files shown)


src/GameLogic/Party.cs
```diff
@@ -232,10 +232,13 @@ private async ValueTask<int> InternalDistributeExperienceAfterKillAsync(IAttacka
         var randomizedTotalExperiencePerLevel = randomizedTotalExperience / totalLevel;
         foreach (var player in this._distributionList)
         {
-            if (player.SelectedCharacter?.CharacterClass?.IsMasterClass ?? false)
+            if ((short)player.Attributes![Stats.Level] == this._distributionList.First().GameContext.Configuration.MaximumLevel)
```
sven-n, line 235: Instead of ```this._distributionList.First()``` you could simply use ```player```.

sven-n (changes requested):

src/GameLogic/Party.cs
urshulgi: Thank you! I just added that change and is working correctly

sven-n:
Thanks again :)

PR #321, opened 2023-02-22
Feat/centralize commom dependencies

MarcusViniciusSS:
The goal is to reduce repeat dependencies like StyleCop and Analyzers.

image

Reference: https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022


sven-n:
Thank you :)

PR #322, opened 2023-02-23
Feat/improvement persistence layer

MarcusViniciusSS:
Some points fixed:

1.  reduce nested if
2.  simplify the logic method
3.  reduce the responsibility of context.
4.  add Guard clauses.
5.  remove magic numbers.
6.  use principle clean code.


sven-n:
Thanks :)

PR #323, opened 2023-02-23
Feat/improvement chat server

MarcusViniciusSS:
Some improvements:

- reduce nested if.
- Use template log instead interpolation.
- Use parallel for each instead of linear for each.
- add guard clauses.

### DIFF
--- a/src/ChatServer/ChatServer.cs
+++ b/src/ChatServer/ChatServer.cs
@@ -292,14 +292,12 @@ public sealed class ChatServer : IChatServer, IDisposable
         e.Cancel = this.CurrentConnections >= this.Settings.MaximumConnections;
     }
 
-    private ValueTask ChatClientAcceptedAsync(ClientAcceptedEventArgs e)
+    private async ValueTask ChatClientAcceptedAsync(ClientAcceptedEventArgs e)
     {
         var chatClient = new ChatClient(e.AcceptedConnection, this._manager, this._loggerFactory.CreateLogger<ChatClient>());
         this._connectedClients.Add(chatClient);
         this.RaisePropertyChanged(nameof(this.CurrentConnections));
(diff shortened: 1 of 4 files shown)


src/ChatServer/ChatServer.cs
```diff
@@ -292,12 +292,14 @@ private async ValueTask ChatClientAcceptingAsync(CancelEventArgs e)
         e.Cancel = this.CurrentConnections >= this.Settings.MaximumConnections;
     }
 
-    private async ValueTask ChatClientAcceptedAsync(ClientAcceptedEventArgs e)
```
sven-n, line 295: it's async by purpose, please revert

src/ChatServer/ChatServer.cs
```diff
@@ -316,15 +318,20 @@ private async void ClientCleanupInactiveClients(object? sender, ElapsedEventArgs
         try
         {
             var bottomDateTimeMargin = DateTime.Now.Subtract(this.Settings.ClientTimeout);
-            for (int i = this._connectedClients.Count - 1; i >= 0; i--)
```
sven-n, line 319: Please revert this loop as well.
The parallel for will cause issues, because client.LogOffAsync mutates the list. The for loop is running backwards for this reason as well :)

sven-n (changes requested):

src/ChatServer/ChatServer.cs
MarcusViniciusSS: What's the reason? 

The reason for changes is "This async method lacks 'await' operators and will run synchronously"

sven-n: First, it doesn't hurt... Second, exception handling might be better - I can look that up or think about that, but I'm too tired right now 😅

sven-n:
Thanks again :)

PR #328, opened 2023-03-12
Refactor/gamelogic

MarcusViniciusSS:
Goal:
    - SImplify logic of ChatMessage Action.
    - Reduce Cognitive complexity,
    - Reduce nested if.
    -  Add Guard Clause.
    -  Remove magic numbers.


src/GameLogic/PlayerActions/Chat/IChatProcessorMessage.cs
```diff
@@ -0,0 +1,19 @@
+﻿// <copyright file="IChatProcessorMessage.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic.PlayerActions.Chat;
+
+/// <summary>
+/// Interface for a chat message processor.
+/// </summary>
+public interface IChatProcessorMessage
```
sven-n, line 10: Can you rename these types to "chat message processor", and not "chat processor message" ?

src/GameLogic/PlayerActions/Chat/IChatProcessorMessage.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Sends a chat message from the player to other players.
+    /// </summary>
+    /// <param name="sender" cref="Player"></param>
+    /// <param name="content"></param>
+    /// <returns></returns>
+    ValueTask ConsumeAsync(Player sender, (string Message, string PlayerName) content);
```
sven-n, line 18: I would call it "ProcessMessageAsync" instead of "ConsumeAsync"

src/GameLogic/PlayerActions/Party/PartyRequestAction.cs
```diff
@@ -20,26 +20,29 @@ public class PartyRequestAction
     /// <param name="toRequest">The player which receives the request.</param>
     public async ValueTask HandlePartyRequestAsync(Player player, Player toRequest)
     {
+        var playerInParty = player.Party != null && !Equals(player.Party.PartyMaster, player);
```
sven-n, line 23: The variable name is misleading. It also checks if the player is the party master.
Suggestion: isPartyMember

src/GameLogic/PlayerActions/Quests/QuestCompletionAction.cs
```diff
@@ -35,21 +35,25 @@ public async ValueTask CompleteQuestAsync(Player player, short group, short numb
 
         foreach (var requiredItem in activeQuest.RequiredItems)
         {
-            if (requiredItem.MinimumNumber > player.Inventory?.Items.Count(i => i.Definition == requiredItem.Item))
+            if (!(requiredItem.MinimumNumber > player.Inventory?.Items.Count(i => i.Definition == requiredItem.Item)))
```
sven-n, line 38: Comparison with ```<=``` would be nicer

src/GameLogic/PlayerActions/Skills/AreaSkillAttackAction.cs
```diff
@@ -15,6 +15,8 @@ namespace MUnique.OpenMU.GameLogic.PlayerActions.Skills;
 /// </summary>
 public class AreaSkillAttackAction
 {
+    private const int DefaultTarget = 0xFFFF;
```
sven-n, line 18: The value doesn't represent a "default target", but an "undefined target"

sven-n (changes requested):
Hey, thanks again for the PR. Please change these things I commented, then it's fine 😅

src/GameLogic/PlayerActions/Chat/IChatProcessorMessage.cs
sven-n: Some comments are missing as well here

sven-n (approved):

PR #343, opened 2023-04-19
fix arrows and bolts consumption issue

terry83299387:
When an elf player logs in to the game, only the arrows or bolts equipped at the beginning will be consumed with the attack, and will no longer be consumed after replacing it with a new one.


sven-n:
Looks perfect, thank you! :)

PR #347, opened 2023-04-22
Feat/api provider example

recone83:
Small HTTP API provider demo.
Easy way to get server state or to controll server.
It is not "ready to publish" rather starting point to discussion.

Endpoints:
http://localhost/api/send?msg=ssss%20aaa%20bbbb - send global message
http://localhost/api/send/0?msg=ssss%20aaa%20bbbb
http://localhost/api/status - print server status/ppl online


sven-n:
Several issues here:
* Adding another host with another port is not required. Your controller can be added to the existing host. In the Startup project, simply call ```builder.Services.AddControllers().AddApplicationPart(typeof(HomeController).Assembly)``` instead of creating your host in the Program.cs of the Startup project.
* The static variable for the game server is not a good idea. It would only work for one of all game servers. Better inject the ```GameServerContainer``` or the ```IDictionary<int, IGameServer>``` by adding them to the constructor of the controller.

PR #348, opened 2023-04-23
fixed warp issues

terry83299387:
This PR fixed following issues:

1. Fixed a bug in calculating level. It should be `levelRequirement * reduction / 100`, not `levelRequirement * 100 / reduction`.
2. Optimized the calculation for levels. (Compared with the previous code, the result might differ by 1~2 levels now.)

The motivation for this modification is:

As an example, in the case of map Tarkan, according to the previous calculation, a MG player should reach level 94 to enter the map using the /move command or warp menu. But when the player was supposed to be able to enter the map at level 93 (which the original server required), he was told that level 94 is required, which will make the player feel confused.

Therefore, for the above reason, I modified the calculation method so that the required level is the same as the original server, or 1~2 levels lower, so that the player's experience may be better.

3. Fixed the description of the required level in the error message when a player's level was lower than requirement.
4. Fixed some incorrect definitions of name, level, and money in Gates.cs.

### DIFF
--- a/src/GameLogic/CharacterExtensions.cs
+++ b/src/GameLogic/CharacterExtensions.cs
@@ -96,7 +96,7 @@ public static class CharacterExtensions
 
         if (character.CharacterClass?.LevelWarpRequirementReductionPercent is { } reduction and > 0)
         {
-            levelRequirement -= levelRequirement * 100 / reduction;
+            levelRequirement = levelRequirement * (100 - reduction) / 100;
         }
 
         return levelRequirement;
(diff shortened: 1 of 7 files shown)


sven-n:
Really nice, thank you!

Just one thing is missing: When we change non-trivial things in the initialization logic (MUnique.OpenMU.Persistence.Initialization), it would be great to have an update plugin, which updates the existing configurations.

One example: https://github.com/MUnique/OpenMU/blob/master/src/Persistence/Initialization/Updates/FixSetBonusesPlugIn.cs

If you could add that, it would be great - if not, I'll do it when I have some time.

terry83299387:
OK, that's better, I will do it later.

sven-n:
Thanks :)

PR #354, opened 2023-06-06
Added Traefik as an option to handle SSL

antonioanerao:
alt text

[Traefik](https://doc.traefik.io/traefik/) is the modern standard for Routing, Load Balancing, and Proxies for the Cloud and Local environments.

Here we are using Traefik to work as a Reverse Proxy to handle SSL to production environment and a simple local dns when working in a local environment.

I added a new folder to deployment directory called all-in-one-traefik with all instructions needed.


sven-n:
Thanks, I just had a quick look at the changes.
So, Traefik is handling getting the SSL certificates from *Let's Encrypt* (or others) by itself? That's pretty nice, so we can remove the certbot stuff 👍
I'll test it in the next few days.

antonioanerao:
@sven-n feel free to send a DM if you need any information about how Traefik works or how to deploy the all-in-one-traefik stack

PR #366, opened 2023-07-24
Attribute System extensions

sven-n:
1. Added the possibility to define account-wide StatAttributes. One example usage: VIP-Flag for accounts.
2. Added the possibility to define power ups which apply only to players on a certain map. With this, you could make players weaker or stronger on a per-map basis.

### DIFF
--- a/src/GameLogic/Player.cs
+++ b/src/GameLogic/Player.cs
@@ -1687,7 +1687,7 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
 
     private async ValueTask OnPlayerEnteredWorldAsync()
     {
-        this.Attributes = new ItemAwareAttributeSystem(this.SelectedCharacter!);
+        this.Attributes = new ItemAwareAttributeSystem(this.Account!, this.SelectedCharacter!);
         this.Inventory = new InventoryStorage(this, this.GameContext);
         this.ShopStorage = new ShopStorage(this);
         this.TemporaryStorage = new Storage(InventoryConstants.TemporaryStorageSize, new TemporaryItemStorage());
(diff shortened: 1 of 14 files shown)


src/GameLogic/Player.cs
```diff
@@ -1685,9 +1707,37 @@ private async ValueTask AfterKilledPlayerAsync()
...
+                if (powerUpDefinition.TargetAttribute is not { } targetAttribute)
+                {
+                    continue;
+                }
+
+                var powerUps = PowerUpWrapper.CreateByPowerUpDefinition(powerUpDefinition, attributes);
+                powerUps.ForEach(p =>
+                {
+                    this.Attributes?.AddElement(p, targetAttribute);
+                    this.PlayerLeftMap += (_, _) => attributes.RemoveElement(p, targetAttribute);
```
afrokick, line 1727: Is it necessary to unsubscribe?

sven-n: yes, of course 😅

afrokick: @sven-n I mean PlayerLeftMap event. Is it automatically GC? AFAIK we need to manually unsubscribe from it like `this.PlayerLeftMap -= someHandler`.

sven-n: oops, I thought i commited that change...

afrokick: Another place without unsubscribe https://github.com/MUnique/OpenMU/blob/master/src/GameLogic/MiniGames/MiniGameContext.cs#L170C9-L170C69

sven-n: Thanks, fixed with #373

PR #385, opened 2023-10-27
🔥 Configuration Hot Reload

sven-n:
When a config is changed and saved by an entity framework context:
* it will forward these changes to a IConfigurationChangeListener which is specific to the Persistence.EntityFramework.
* The implementation of that will
  * update the cached configuration objects and their parent objects, in case it's an collection
  * notify an instance of a ConfigurationChangeHandler which also existed before
* The change handler will notify the change mediator (game logic), on which several logic objects will subscribe for changes
  * currently, there are two places which subscribe to changes: monster/npc spawns and merchant npc items.

To make this work, especially updating the cached objects, I had to implement something like ICloneable<T> and Assignable<T>. I did this by implementing a code generator and a CloneableAttribute.


sven-n:
I'll merge that for now. If we find any configuration not live-updating, it'll be extended on-demand.

PR #393, opened 2023-12-07
Prevent equipping two-handed items with shields

halflumi:
Prevent equipping a two-handed item to the left hand slot when the right hand slot has a one-handed item or shield, and vice versa.


halflumi:
I should've cherry-picked instead of manually copying n' pasting. Now it's fixed.

PR #395, opened 2023-12-10
Compensate level-up points when gaining Hero Status beyond level 220

### LINKED ISSUE #394: Level-up points should be compensated when gaining Hero Status beyond level 220
**Describe the bug**
Gaining Hero Status beyond level 220 would normally get compensated for 1 level-up point per level-up after level 220. The points are added correctly on the client side upon quest completion, but they can't be spent and disappear on logout.

**To Reproduce**
1. Reset server to 1.04 with test accounts.
2. Log in to character `quest2Dk` of the test account `quest2`.
3. Gain some levels, in this case, 5.
4. Dump all 420 (initial 395 + 25 from level-ups) points to `Energy`.
5. Complete the 2 class change quests. Now there are 20 level-up points from the quest reward.
6. Complete the `Hero Status` quest. Now there are 25 level-up points, 5 more from the `Hero Status` quest, compensated for level-ups from level 221 to 225.
7. Spend the points. Notice that the last 5 points can't be spent and disappear on logout.

**Expected behavior**
The compensated points can be spent as usual.

halflumi:
Closes #394

### DIFF
--- a/src/GameLogic/PlayerActions/Quests/QuestCompletionAction.cs
+++ b/src/GameLogic/PlayerActions/Quests/QuestCompletionAction.cs
@@ -86,14 +86,14 @@ public class QuestCompletionAction
 
         foreach (var reward in activeQuest.Rewards)
         {
-            await AddRewardAsync(player, reward).ConfigureAwait(false);
+            await AddRewardAsync(player, reward, activeQuest).ConfigureAwait(false);
         }
 
         await questState.ClearAsync(player.PersistenceContext).ConfigureAwait(false);
         await player.InvokeViewPlugInAsync<IQuestCompletionResponsePlugIn>(p => p.QuestCompletedAsync(activeQuest)).ConfigureAwait(false);
     }
 
-    private static async ValueTask AddRewardAsync(Player player, QuestReward reward)
+    private static async ValueTask AddRewardAsync(Player player, QuestReward reward, QuestDefinition quest)
     {
         switch (reward.RewardType)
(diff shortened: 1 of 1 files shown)


sven-n:
Thanks for your PR, but I think it could be improved, so it doesn't depend on magic values (level 220, 1 as reward) and is more tolerant to changes to the quest data:
```csharp
            case QuestRewardType.Attribute:
                var attribute = player.SelectedCharacter!.Attributes.FirstOrDefault(a => a.Definition == reward.AttributeReward);
                if (attribute is null)
                {
                    attribute = player.PersistenceContext.CreateNew<StatAttribute>(reward.AttributeReward, 0);
                    player.SelectedCharacter.Attributes.Add(attribute);
                }

                attribute.Value += reward.Value;

                // Compensate level-up points when doing a quest at a later level.
                if (attribute.Definition == Stats.PointsPerLevelUp)
                {
                    var playerLevel = (int)player.Attributes![Stats.Level];
                    player.SelectedCharacter.LevelUpPoints += (playerLevel - quest.MinimumCharacterLevel) * reward.Value;
                }

                await player.InvokeViewPlugInAsync<ILegacyQuestRewardPlugIn>(p => p.ShowAsync(player, QuestRewardType.Attribute, reward.Value, attribute.Definition)).ConfigureAwait(false);
                break;
```
What do you think?

halflumi:
That's a good idea! I've adopted your code to the PR.

Issue #400, opened 2024-01-10
A Fenrir cannot be repaired with a Jewel of Bless


sven-n:
**Describe the bug**
A Fenrir cannot be repaired with a Jewel of Bless.

**To Reproduce**
1. Prepare a test account, e.g. `testgm` or `test400` on the admin panel, so that the fenrir has about 100 durability.
2. Enter the game, login with the account.
3. Take a _Jewel of Bless_ and drop it on the Fenrir
4. See that the fenrir didn't get repaired and the _Jewel of Bless_ is still there.

**Expected behavior**
The fenrir should be repaired to 255 durability and the _Jewel of Bless_ should be consumed.

**Additional context**
See https://github.com/MUnique/OpenMU/blob/a8703d1dce1f500277696316b2fed30e2f326164/src/GameLogic/PlayerActions/ItemConsumeActions/BlessJewelConsumeHandlerPlugIn.cs#L18
Currently, it just works by upgrading an item level.
Hard coding the fenrir into the `ModifyItem` method to get the behavior is possible. However, it would be great to be able to define which Item can be repaired by a bless in a configuration. A plugin configuration could be added by implementing `ISupportCustomConfiguration<TConfig>` and `ISupportDefaultCustomConfiguration`, but I'm not sure if a lookup for items is possible with the current system.

PR #403, opened 2024-03-13
fix: Repair Fenrir with Jewel of Bless

radzki:
This PR aims to solve the Fenrir Repair bug, but with a different approach than the one suggested in the issue. Instead of adding a plugin configuration, I've added a new column to the Item itself. A new field will appear when configuring Items:

image

Tested it in-game and it's working pretty well. Is it good enough, though? hahahah

Possibly closes https://github.com/MUnique/OpenMU/issues/400


radzki:
I'm currently trying to add a plugin to the Updates folder, which would change the fenrir value to true, but it doesn't seem to be running (?)
image

radzki:
Closed based on Discord's review 😆

Issue #392, opened 2023-12-06
Prevent equipping weapons and shields when not enough hands are free


sven-n:
**Describe the bug**
Two handed weapons can be equipped, even if a shield is already equipped.

**To Reproduce**
1. Enter the game with a wizard character which has a `Chaos Lightning Staff` and a `Skull Shield`.
2. Equip the shield
3. Equip the staff
4. See that both are equipped, even if the staff requires both hands to wear it.

**Expected behavior**
The staff should not be able to be equipped.
Also, if you equip the staff first, the shield should not be able to be equipped.

**Additional context**
We can determine if a weapon requires two hands based on the width of the item. Every weapon which needs two hands is at least 2 fields wide.
The behavior is wrongly implemented here, as it only considers to prevent equipping, when the item to equip is a weapon which can be equipped on both hands (one-handed swords), and there is already an two-handed item equipped on the left hand:
https://github.com/MUnique/OpenMU/blob/6b933062f3de667ac67c279355e15fbc6640812b/src/GameLogic/PlayerActions/Items/MoveItemAction.cs#L237-L243

This logic needs to consider more cases, probably even preventing to equip a shield when an elf already equipped a bow or crossbow.
For elfs there is an exception for arrows and bolts - they occupy a slot, but not use any hand. However, when arrows or bolts are equipped it makes to sense to allow shields.

halflumi:
Thank you for the excellent summary.

After discovering the staff-shield situation, I did some tests. It appears that the client already prevents most invalid combinations, with shields being the primary exceptions.

I've opened Issue #393 to address these shield cases.

sven-n:
Yea, however, we should also make sure that the server prevents invalid things - we don't want to trust the client :)

halflumi:
Indeed. It seems plausible to expand the `group` checks to include more invalid cases. Just gotta list all of them.

radzki:
@sven-n I was looking through issues where I could help, and it seems that this is already solved (?)

https://github.com/MUnique/OpenMU/blob/29229f6d673c937dfbf71406a255942a02d78e9e/src/GameLogic/PlayerActions/Items/MoveItemAction.cs#L235-L251

Should we close the issue?

PR #404, opened 2024-03-15
fix: Adds conditions to avoid equipping two-handed item when all slots are occupied

radzki:
Closes https://github.com/MUnique/OpenMU/issues/392

This PR goes one step further after https://github.com/MUnique/OpenMU/pull/393 and solves the case for one-handed items (such as shields and swords) being able to be equipped in the second slot.

This doesn't change the already working behavior for elves (bolts and arrows)


sven-n:
Thank you and sorry that it took so long 😅

PR #415, opened 2024-06-08
ADD reset points can be configured by character class

Niter88:
Note that the 'GetResetPoints' method will support changing on final result by VIP status.


Niter88:
This solves the issue #416  for now

sven-n:
Hey, I did some more changes here:
1. Renamed the attribute to be more generic. It's not tied to a character class.
2. To get the attribute value, you can simply call `player.Attributes[Stats...]`.
3. Added configuration update plugin to update existing databases.

sven-n:
Thanks @Niter88 👍

PR #422, opened 2024-07-05
Small changes & fixes

recone83:
- Admin Panel - left menu was broken on laptop(fixed)
- Guard was not able to walk on Safe Zone(fixed)
- Chaos Castle, minimum players 1 - standard now(jkok25 proposed)
- FenrirUpgradeCrafting (jkok25 proposed)
- Rage-fighter enabled
- Create-monster command fix
- Golden invasion fix
- Red invasion fix
- NPC random move range fixed(was hard-coded to 2)
- disable speed hack validation 
- POST command - everyone can write post - fix


src/GameLogic/NPC/Monster.cs
```diff
@@ -227,23 +227,38 @@ internal async ValueTask RandomMoveAsync()
         {
             return;
         }
+        
+        byte randx = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.X - 1), Math.Min(0xFF, this.Position.X + this.Definition.MoveRange));
```
sven-n, line 231: The 2 is correct. It's randomly moving by one step when it has no target and is watched by a player.
I'm note sure what the MoveRange actually means.
If you change it like that, the monsters will all walk towards max x and y

recone83: Rand(0, max) - sounds good?

sven-n: Okay, but when you add MoveRange just to the second parameter of `Rand.NextInt`, the character will more likely move to a x coordinate which is higher than the current. How about:
```
var moveByMaxX = GameLogic.Rand.NextInt(1, this.Definition.MoveRange + 1);
var moveByMaxY = GameLogic.Rand.NextInt(1, this.Definition.MoveRange + 1);
byte randx = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.X - moveByMaxX), Math.Min(0xFF, this.Position.X + moveByMaxX + 1));
byte randy = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.Y - moveByMaxY), Math.Min(0xFF, this.Position.Y + moveByMaxY + 1));
```

recone83: added

src/GameLogic/MiniGames/ChaosCastleContext.cs
```diff
@@ -89,7 +89,7 @@ public ChaosCastleContext(MiniGameMapKey key, MiniGameDefinition definition, IGa
     protected override TimeSpan RemainingTime => this._remainingTime;
 
     /// <inheritdoc />
-    protected override int MinimumPlayerCount => 2;
+    protected override int MinimumPlayerCount => 1;
```
sven-n, line 92: I accept this change for now, but I'd like to add this as setting to the MiniGameDefinition, so everyone can decide how it should be on his server. The normal behavior on the official server is, that it requires two players to start the event.

src/GameLogic/NPC/Monster.cs
```diff
@@ -228,22 +228,39 @@ internal async ValueTask RandomMoveAsync()
...
 
-        byte randx = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.X - 1), Math.Min(0xFF, this.Position.X + 2));
-        byte randy = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.Y - 1), Math.Min(0xFF, this.Position.Y + 2));
-        if (this.CurrentMap.Terrain.AIgrid[randx, randy] == 1)
+        var moveByMaxX = GameLogic.Rand.NextInt(1, this.Definition.MoveRange + 1);
+        var moveByMaxY = GameLogic.Rand.NextInt(1, this.Definition.MoveRange + 1);
+        byte randx = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.X - moveByMaxX), Math.Min(0xFF, this.Position.X + moveByMaxX + 1));
+        byte randy = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.Y - moveByMaxY), Math.Min(0xFF, this.Position.Y + moveByMaxY + 1));
+
+        switch (this.Definition.ObjectKind)
```
sven-n, line 236: Can you move this special treatment to the GuardIntelligence?
Simply add a virtual method like that to the MonsterIntelligenceBase and overwrite it on the GuardIntelligence:
```csharp
public virtual bool CanWalkOn(Point target)
{
    return this.Monster.CurrentMap.Terrain.AIgrid[randx, randy] == 1;
}
```

This way, the code in the `RandomMoveAsync()` looks cleaner and it doesn't have to be extended at this place again when a new kind needs special treatment.

src/GameLogic/PlayerActions/Craftings/FenrirUpgradeCrafting.cs
sven-n, line 1: Can you explain what was wrong before? 😅

sven-n (changes requested):

src/Web/AdminPanel/Styles/Navigation.scss
```diff
@@ -15,7 +15,7 @@ app > nav > div:first-child {
 @media (min-width: $widthThreshold) {
     app > nav {
         width: 250px;
-        height: 100vh;
+        height: 100%;
```
sven-n, line 18: This looks good when the page is long, but worse when it's short

sven-n: I'll keep it and add round edges and shadows to compensate that 😅

sven-n: I found a way to get the best of both solutions... it's now always fully visible whether scrolled or not.

src/GameLogic/PlayerActions/Craftings/FenrirUpgradeCrafting.cs
sven-n: I think I understood the code now... You want to allow to create a golden fenrir.
That's a good idea, however, I want to keep the original behavior. What we can do is, to add your version of this crafting side-by-side, so it can be changed by editing the `ItemCraftingHandlerClassName` in the crafting configuration.

sven-n (approved):

PR #424, opened 2024-07-05
Feat/small changes

recone83:


src/GameLogic/NPC/Monster.cs
```diff
@@ -227,9 +227,9 @@ internal async ValueTask RandomMoveAsync()
         {
             return;
         }
-
-        byte randx = (byte)GameLogic.Rand.NextInt(Math.Max(0, this.Position.X - 1), Math.Min(0xFF, this.Position.X + 2));
```
sven-n, line 231: The 2 is correct. It's randomly moving by one step when it has no target and is watched by a player.
I'm note sure what the MoveRange actually means.
If you change it like that, the monsters will all walk towards max x and y

PR #439, opened 2024-07-20
Adding more skills

sven-n:
Added:
* The three buffs of ragefighter: ignore defense, increase health, increase block
* wizardry enhance, including the master skill stages


sven-n:
Decided to merge that for now, so I can continue to work on some fixes which need database updates

PR #445, opened 2024-08-10
Fix all-in-one deployment bug

Kishoft:
**Description**
When i run "All in one deployment with 'Option B'", nginx cant find one file

**To Reproduce**
1. Go to OpenMu/deploy/all-in-one
2. Run "docker compose -f docker-compose.yml -f docker-compose.prod.yml up -d"
3. Run "docker ps"
4. Copy nginx 80 container hash
5. Run "docker logs (hash here)"
6. Logs will show "2024/08/10 20:37:36 [emerg] 1#1: open() "/etc/nginx/conf.d/nginx.server_name.conf" failed (2: No such file or directory) in /etc/nginx/nginx.conf:9"

**Expected behavior**
Nginx should work but it isn't

**Additional data:**
 - OS: Ubuntu 22.04 LTS local deployment on real PC

### DIFF
--- a/deploy/all-in-one/docker-compose-all-in-one.dcproj
+++ b/deploy/all-in-one/docker-compose-all-in-one.dcproj
@@ -20,9 +20,11 @@
     <None Include="docker-compose.yml" />
     <None Include=".dockerignore" />
     <None Include=".htpasswd" />
+    <None Include="nginx.prod.certificates.conf.template" />
     <None Include="nginx.prod80.conf" />
     <None Include="nginx.dev.conf" />
     <None Include="nginx.prod443.conf" />
+    <None Include="nginx.server_name.conf.template" />
     <None Include="README.md" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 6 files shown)


sven-n:
Thank you for your fix. Before nginx version 1.19, you couldn't use environment variables in nginx config files. Nginx was scanning the config files for `*.template`, replaced env variables and stored it without the `.template` suffix in the same folder. So at first, I was surprised that your fix works 😅
Since nginx improved at this point, I think this can be further improved: We can remove this template files and directly add their contents to the places where they were included.

sven-n:
I did the changes, can you confirm that it's working?

sven-n:
Oh nevermind, I was wrong... they didn't change a thing.

https://www.baeldung.com/linux/nginx-config-environment-variables

Kishoft:
Hi! Don’t worry :D, I’ll be working on better parameterizing the environment variables within the container with Nginx so that everything works without needing any modifications (except hard coding non-params env vars). The previous issue was resolved by mounting a file inside the container as read-only. Once I have everything set up with best practices, I’ll be making another request. Thank you for your feedback !

sven-n:
This was handled with #447

PR #466, opened 2024-08-20
Implement Blood Castle start time system 

bernatvadell:
Based on the implementation of Chaos Castle & Devil Square, I have implemented the timed startup system for Blood Castle.
Please review it carefully, I have tested it in the game and everything seems to work correctly.

### DIFF
--- a/src/GameLogic/PlugIns/PeriodicTasks/BloodCastleStartPlugIn.cs
+++ b/src/GameLogic/PlugIns/PeriodicTasks/BloodCastleStartPlugIn.cs
@@ -1,0 +1,146 @@
+﻿// <copyright file="BloodCastleStartPlugIn.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic.PlugIns.PeriodicTasks;
+
+using System.Runtime.InteropServices;
+using MUnique.OpenMU.GameLogic.MiniGames;
+using MUnique.OpenMU.PlugIns;
+
+/// <summary>
+/// This plugin enables the start of the blood castle.
+/// </summary>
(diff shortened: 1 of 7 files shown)


src/GameLogic/PlugIns/PeriodicTasks/BloodCastleStartPlugIn.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Runtime.InteropServices;
+using MUnique.OpenMU.GameLogic.MiniGames;
+using MUnique.OpenMU.PlugIns;
+
+/// <summary>
+/// This plugin enables the start of the blood castle.
+/// </summary>
+[PlugIn(nameof(BloodCastleStartPlugIn), "Blood Castle event")]
+[Guid("95E68C14-AD87-4B3C-AF46-45B8F1C3BC2A")]
+public sealed class BloodCastleStartPlugIn : PeriodicTaskBasePlugIn<BloodCastleStartConfiguration, BloodCastleGameServerState>, ISupportDefaultCustomConfiguration, IPeriodicMiniGameStartPlugIn
```
sven-n, line 16: Since we have now three very similar Start-classes, can we introduce a new generic base class which implements the common code?

sven-n (changes requested):
Thanks again, great work!
Can you do the base class or should I work on that later?

src/GameLogic/PlugIns/PeriodicTasks/BloodCastleStartPlugIn.cs
bernatvadell: Sure, let me check it out.

bernatvadell:
Please take a look to see if this is what you had in mind.
If you have any suggestions, let me know and I'll adjust it.

sven-n (approved):
Perfect! :)

PR #472, opened 2024-08-24
Fix wizardry dmg calc

bernatvadell:
Based on the files opened in the sven client, I have been able to confirm that the "maximumBaseDamage" value should also be equal to half of the "skillDamage" value.

client info:
image

server info:
image


sven-n:
Thank you :)

PR #473, opened 2024-08-25
Fix trade button state

bernatvadell:
We fixed an issue where it was not indicated when the user accepted a trade:

image


sven-n:
Thank you :)

PR #474, opened 2024-08-26
Jewels in a different drop item group

bernatvadell:
I add the jewels in a different drop item group to adjust chance individually.


bernatvadell:
@sven-n for this task, I had to modify the drop generator, since the Ferris wheel monsters could drop Lifes or Creations.

Please keep that in mind in case there is anything you need to modify.

src/Persistence/Initialization/Updates/AddItemDropGroupForJewelsSeason6Update.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <inheritdoc />
+    public override string DataInitializationKey => VersionSeasonSix.DataInitialization.Id;
+
+    /// <inheritdoc />
+    protected override async ValueTask ApplyAsync(IContext context, GameConfiguration gameConfiguration)
+    {
+        await base.ApplyAsync(context, gameConfiguration).ConfigureAwait(false);
+        this.AddJewelToItemDrop(gameConfiguration, "Jewel of Creation");
+        this.AddJewelToItemDrop(gameConfiguration, "Jewel of Guardian");
```
sven-n, line 30: 'Jewel of Guardian' only drops in Land of Trials, the map where only castle owners can grant access to.

bernatvadell: Previously it was enabled to be dropped by any mob with a level higher than 75.

I don't know if the only mobs that exceed that level are those on the Land of Trials map.

In any case, in my mind I have two options:
A) Leave the Jewel of Guardian in the jewel group, with the change I made so that only items from monsters with a sufficient level would drop, the same logic that was there before should be maintained.

https://github.com/MUnique/OpenMU/pull/474/files#diff-04390984d3c073558c83c0fa8e6c67c09da572f39be91c23e77c79cf9825c8d6R409

B) Remove the item from the jewel group, and add it to a new drop group for that specific map.

sven-n: I'd vote for B 😅

bernatvadell: Much clearer, really. So I'll prepare it :)

bernatvadell: @sven-n done!

sven-n (approved):

sven-n:
Thanks again 🙏

PR #477, opened 2024-09-01
Implement skill sleep

bernatvadell:
Implement Summoner Sleep Skill

https://github.com/orgs/MUnique/projects/3/views/1?pane=issue&itemId=67145119

# Tasks

- [x] Add new stat IsAsleep
- [x] Implement logic for new stat IsAsleep
- [x] Update Skills initializer to fix Sleep
- [x] Create update plugin to fix sleep into existing databases

I have implemented a new RandomElement class that is not being used.
We would have to analyze the best way to make the sleep have a random value for the duration of the magic effect.

### DIFF
--- a/src/AttributeSystem/RandomElement.cs
+++ b/src/AttributeSystem/RandomElement.cs
@@ -1,0 +1,69 @@
+﻿// <copyright file="RandomElement.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.AttributeSystem;
+
+/// <summary>
+/// An element with a constant value.
+/// </summary>
+public class RandomElement : IElement
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RandomElement" /> class.
+    /// </summary>
+    /// <param name="minValue">The min value.</param>
+    /// <param name="maxValue">The max value.</param>
+    /// <param name="aggregateType">Type of the aggregate.</param>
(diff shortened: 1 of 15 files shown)


sven-n:
Why do we need a random sleep duration? I couldn't find anything about that in the web or other server sources

bernatvadell:
True, it's not random, I was checking the source of a GS and I saw this. I noticed that it doesn't always "sleep", it has a success ratio.

image

How do we implement this? Is there a similar example?

On the other hand, should I delete the new RandomElement class? Or do you think it could be useful at some point?

sven-n:
I don't think we need the RandomElement then.
About the attack success rate: I think it could make sense to add a "Stats.SleepAttackRate" (expected value between 0 and 1) and some AttributeRelationships for that to the character class. It will require some code changes in GameLogic to apply that, too.

bernatvadell:
From what I can see, depending on whether you're attacking a monster or a player, the ratio changes. I understand that we can't manage this use case with a Stat, can we?

sven-n:
Oh yes, well we could add two stats for that. For example, the attack rate has one for pvp and one for pvm as well.

bernatvadell:
I was looking at the architecture of MagicEffectDefinition and I'm thinking about how to make it generic.

Many Summoner effects involve a success ratio when the "Buff" type magic is applied to the enemy.

Do you think it would be convenient to modify this class to optionally include a success ratio specification?

If so, should we use PowerUpDefinition?

On the other hand, do you think it would be interesting to extend AttributeRelationship to be able to indicate if that attribute applies under certain restrictions? For example, being able to restrict that relationship by Target.

This would allow us to modify the formula, without needing to add new "Stats".

PR #483, opened 2024-09-02
Added PVP option to GameServerDefinition and AdminPanel

nReCon:
This adds the ability to toggle PVP on individual game servers.
PVP is enabled by default as that's the default behavior of OpenMU before this change.

I'm not sure how battle zones should be handled.
This PR allows them on non-PVP servers, let me know if this shouldn't be the case.

This is my first PR for this project, and this is also my first time using C#, so please let me know if I'm doing something wrong here.

It also adds a checkbox in the admin panel:
image

### DIFF
--- a/src/Dapr/GameServer.Host/Pages/Index.razor
+++ b/src/Dapr/GameServer.Host/Pages/Index.razor
@@ -21,6 +21,9 @@
         <span class="badge badge-primary">
             Experience Rate: <span class="badge badge-light">@_gameServerContext.ExperienceRate</span>
         </span>
+        <span class="badge badge-primary">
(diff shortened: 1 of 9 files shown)


sven-n:
Thanks, it looks very good!
One thing is missing: A migration for the entity framework ;-)
See also https://github.com/MUnique/OpenMU/blob/master/src/Persistence/EntityFramework/Migrations/Cheatsheet.md

nReCon:
I've added the migration, and I set the default value to true as well because apparently I forgot to do that originally.

sven-n (approved):

sven-n:
Thanks, it looks perfect!

PR #486, opened 2024-09-03
Improvements for the connect server ip resolve

sven-n:
We always have the issue that the server doesn't magically work after it has been installed. One major thing is that the game client gets disconnected after selecting a server in the server list.
I had the idea that, if the client connects, we can look at the Socket.LocalEndPoint to which IP it initially connected. If the ip address of the connect server is the same as the ip of the choosen game server we can assume that the client has also access to the game server by connecting to the same ip address.

Probably needs some additional tests for the different scenarios, so I leave it as draft for now.


sven-n:
The same logic should be added to the chat server as well, but in a new PR.

PR #491, opened 2024-09-09
Fixed exit gate's coordinates to walkable coordinates

ze-dom:
Dungeon 240/148 is a non-walkable coordinate (map decoration), which would cause a player entering gate from D2->D1 and being placed there to be warped to Lorencia (evidenced in video).

https://github.com/user-attachments/assets/503555db-b72a-4bb1-ab8b-5561cb77b9bd


sven-n:
Thanks :)

Issue #460, opened 2024-08-18
Adding/removing GameServers


sven-n:
**Is your feature request related to a problem? Please describe.**
No user-friendly way to add more game-servers to existing installation.

**Describe the solution you'd like**
It would be nice if we could add or remove gameservers over the admin panel.
This could also be possible at run-time, so that no server restart is required.

The server list of the admin panel should be extended with some buttons for that purpose.

PR #492, opened 2024-09-09
Improved game server management through admin panel

sven-n:
Game Servers can now be added and removed at runtime of the server.


sven-n:
Fixes #460.

PR #496, opened 2024-09-24
Fixed max element jewelry resistance

ze-dom:
Fixes:

- Resistance base power ups missing from jewelry items
- Max element resistance given by jewelry - only the highest level pendant or ring applies => max element resistance = 0.5, which implies an equipped pendant or ring level +4 (upon investigation)
- Jewelry items max option level from +4% to +3% (upon investigation)
- Excellent and ancient ring items giving additional base defense
- Excellent wizardry attack options giving excellent physical damage options

### DIFF
--- a/src/GameLogic/Attributes/Stats.cs
+++ b/src/GameLogic/Attributes/Stats.cs
@@ -553,37 +553,37 @@ public class Stats
     public static AttributeDefinition IsStunned { get; } = new (new Guid("22C86BAF-7F27-478D-8075-E4465C2859DD"), "Is stunned", "The player is poisoned and loses health");
 
     /// <summary>
-    /// Gets the ice resistance attribute definition. Value range from 0 to 1.
+    /// Gets the ice resistance attribute definition. Value range from 0 to 0.5.
     /// </summary>
     public static AttributeDefinition IceResistance { get; } = new (new Guid("47235C36-41BB-44B4-8823-6FC415709F59"), "Ice Resistance", string.Empty);
 
     /// <summary>
-    /// Gets the fire resistance attribute definition. Value range from 0 to 1.
+    /// Gets the fire resistance attribute definition. Value range from 0 to 0.5.
     /// </summary>
     public static AttributeDefinition FireResistance { get; } = new (new Guid("9AE4D80D-5706-48B9-AD11-EAC4FE088A81"), "Fire Resistance", string.Empty);
 
     /// <summary>
-    /// Gets the water resistance attribute definition. Value range from 0 to 1.
+    /// Gets the water resistance attribute definition. Value range from 0 to 0.5.
     /// </summary>
     public static AttributeDefinition WaterResistance { get; } = new (new Guid("3AF88672-D8DB-44E1-937A-7E6484134C39"), "Water Resistance", string.Empty);
 
     /// <summary>
-    /// Gets the earth resistance attribute definition. Value range from 0 to 1.
+    /// Gets the earth resistance attribute definition. Value range from 0 to 0.5.
     /// </summary>
     public static AttributeDefinition EarthResistance { get; } = new (new Guid("4470890F-00CE-44A6-BADB-203684B6014D"), "Earth Resistance", string.Empty);
 
     /// <summary>
-    /// Gets the wind resistance attribute definition. Value range from 0 to 1.
+    /// Gets the wind resistance attribute definition. Value range from 0 to 0.5.
     /// </summary>
     public static AttributeDefinition WindResistance { get; } = new (new Guid("03A29C46-7B7E-424D-8325-8390692570C3"), "Wind Resistance", string.Empty);
 
     /// <summary>
-    /// Gets the poison resistance attribute definition. Value range from 0 to 1.
+    /// Gets the poison resistance attribute definition. Value range from 0 to 0.5.
     /// </summary>
     public static AttributeDefinition PoisonResistance { get; } = new (new Guid("3D50D0B7-63A2-4DA9-8855-12173EAE6B39"), "Poison Resistance", string.Empty);
 
     /// <summary>
-    /// Gets the lightning resistance attribute definition. Value range from 0 to 1.
+    /// Gets the lightning resistance attribute definition. Value range from 0 to 0.5.
     /// </summary>
     public static AttributeDefinition LightningResistance { get; } = new (new Guid("3E339393-2D17-452E-81D9-3987947A407F"), "Lightning Resistance", string.Empty);
 
--- a/src/GameLogic/ItemPowerUpFactory.cs
+++ b/src/GameLogic/ItemPowerUpFactory.cs
@@ -27,7 +27,7 @@ public class ItemPowerUpFactory : IItemPowerUpFactory
     }
 
     /// <inheritdoc/>
-    public IEnumerable<PowerUpWrapper> GetPowerUps(Item item, AttributeSystem attributeHolder)
+    public IEnumerable<PowerUpWrapper> GetPowerUps(Item item, AttributeSystem attributeHolder, bool skipBasePowerUps = false, bool skipOptionPowerUps = false)
     {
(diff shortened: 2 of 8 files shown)


src/Persistence/Initialization/Items/OptionExtensions.cs
```diff
@@ -80,6 +80,6 @@ public static ItemOptionDefinition ExcellentPhysicalAttackOptions(this GameConfi
     /// <returns>The excellent wizardry attack options.</returns>
     public static ItemOptionDefinition ExcellentWizardryAttackOptions(this GameConfiguration gameConfiguration)
     {
-        return gameConfiguration.ItemOptions.First(o => o.Name == ExcellentOptions.PhysicalAttackOptionsName);
+        return gameConfiguration.ItemOptions.First(o => o.Name == ExcellentOptions.WizardryAttackOptionsName);
```
sven-n, line 83: Good catch! 😅

src/GameLogic/Attributes/Stats.cs
```diff
@@ -553,37 +553,37 @@ public class Stats
     public static AttributeDefinition IsStunned { get; } = new (new Guid("22C86BAF-7F27-478D-8075-E4465C2859DD"), "Is stunned", "The player is poisoned and loses health");
 
     /// <summary>
-    /// Gets the ice resistance attribute definition. Value range from 0 to 1.
+    /// Gets the ice resistance attribute definition. Value range from 0 to 0.5.
```
sven-n, line 556: Why 0.5?
I understand that there are only Rings and Pendants up to level 4 in the official game, but that could be higher for another configuration

sven-n:
Thanks for the work, I need some time to think this through 😅
I think it could make some sense to make this logic more generic, not limited to rings/pendants and resistancies. But I have no idea yet how to add this, maybe as an additional setting in the AttributeDefinition.

sven-n:
💡 Just had an idea... how about adding a new 'AggregateType.Maximum' which takes the highest of the available values? 😅
This could simplify the code a lot.

src/GameLogic/Attributes/Stats.cs
ze-dom: Yes, I didn't think of that. I'll change this and leave a "default = 0.5/max level = 4" note.

ze-dom:
@sven-n , looks much cleaner now :smile: 
Tested with Master Skill Tree points element resistance increase and seems to work fine (adds MST resistance (`AggregateType.AddRaw`) to the highest jewelry item resistance (`AggregateType.Maximum`)).

src/GameLogic/ItemPowerUpFactory.cs
```diff
@@ -184,12 +184,25 @@ private IEnumerable<PowerUpWrapper> GetBasePowerUpWrappers(Item item, AttributeS
         attribute.ThrowNotInitializedProperty(attribute.BaseValueElement is null, nameof(attribute.BaseValueElement));
         attribute.ThrowNotInitializedProperty(attribute.TargetAttribute is null, nameof(attribute.TargetAttribute));
 
-        yield return new PowerUpWrapper(attribute.BaseValueElement, attribute.TargetAttribute, attributeHolder);
-
-        var levelBonus = (attribute.BonusPerLevelTable?.BonusPerLevel ?? Enumerable.Empty<LevelBonus>()).FirstOrDefault(bonus => bonus.Level == item.Level);
-        if (levelBonus is not null)
+        var levelBonusElmt = (attribute.BonusPerLevelTable?.BonusPerLevel ?? Enumerable.Empty<LevelBonus>())
```
sven-n, line 187: Are these changes still required?

ze-dom: Yes. 187-190 deleted lines went into the 200-205 added lines in the `else` statement.

sven-n: I mean, is the special handling for jewellery still required? Shouldn't it work already without it? 😀

ze-dom: The way I see it, yes it is.
Without this special handling, if you equipped, for example, two rings of ice, one +2, one +3, you would have a `CompositeAttribute` with the the elements list: [0.1, 0.2, 0.1, 0.3], thus, when getting the maximum value, you would get 0.3. With this special handling your list now becomes [0.3, 0.4], thus the maximum value is 0.4, which is the correct resistance. 
Do you see a better way?

sven-n: Ah, now I understand. Thank you for the explaination.
I think it could make sense here to apply the new logic for all items, not just jewellery. What do you think?

ze-dom: Do you mean always making a `CombinedElement` for the base + level bonus elements whenever there is a level bonus element? Yes, that seems to make sense.

sven-n: So, can you change this, when you have time? Thank you :)

ze-dom: @sven-n , PR updated.
For this change the wings' dmg absorption and increase level bonus tables also needed to be updated. There was also a bug here, as the values of level bonus were being compounded instead of added together as a unique multiplier.
For example, with the following wings:
Screenshot from 2024-10-08 10-22-26

**Before changes**:
- for the dmg increase you would have ~81% (1.39*1.3) instead of 69%:
Screenshot from 2024-10-08 10-35-30
- for the dmg absorb you would have ~57% (1 - 0.61*0.7) instead of 69%:
Screenshot from 2024-10-08 10-33-54

**After changes,** with a `CombinedElement`:
- now actual 69% increase:
Screenshot from 2024-10-08 10-20-41
- now actual 69% absorption
Screenshot from 2024-10-08 10-21-29

src/Persistence/Initialization/Updates/FixWingsDmgRatesUpdatePlugIn095d.cs
```diff
@@ -0,0 +1,24 @@
+﻿// <copyright file="SystemConfigurationAddedPlugIn095d.cs" company="MUnique">
```
sven-n, line 1: file name

src/Persistence/Initialization/Updates/FixWingsDmgRatesUpdatePlugInBase.cs
```diff
@@ -0,0 +1,59 @@
+﻿// <copyright file="SystemConfigurationAddedPlugInBase.cs" company="MUnique">
```
sven-n, line 1: file name

src/Persistence/Initialization/Updates/FixWingsDmgRatesUpdatePlugInSeason6.cs
```diff
@@ -0,0 +1,24 @@
+﻿// <copyright file="SystemConfigurationAddedPlugInSeason6.cs" company="MUnique">
```
sven-n, line 1: file name

src/GameLogic/ItemPowerUpFactory.cs
```diff
@@ -253,7 +261,7 @@ private IEnumerable<PowerUpWrapper> CreateExcellentAndAncientBasePowerUpWrappers
         var baseDropLevel = item.Definition!.DropLevel;
         var ancientDropLevel = item.Definition!.CalculateDropLevel(true, false, 0);
 
-        if (InventoryConstants.IsDefenseItemSlot(item.ItemSlot))
+        if (InventoryConstants.IsDefenseItemSlot(item.ItemSlot) && !item.IsJewelry())
```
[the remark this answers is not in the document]
ze-dom: I put this here because exc rings and pendants were giving `additionalDefense` > 0 (see line 259/267), even if `baseDefense` = 0.

sven-n (approved):

Issue #511, opened 2024-10-24
Appearance Serializer for character does not include ancient options


nitoygo:
**Describe the bug**
Appearance Serializer for own character and other players in scope does not include ancient items

**To Reproduce**
1. Login 'ancient'
2. Select any character
3. Check character appearance
4. // Notice that the armor does not have ancient set aura
5. Switch/reload character
6. // Notice that the armor now show ancient set aura

**Expected behavior**
Ancient set aura should be shown always if complete

**Additional context**
* This also happen for new characters entering your scope of vicinity that has ancient sets equipped
* From initial investigation, newPlayer.AppearanceData.EquippedItems does not have any ancient option included (data fetch issue?)

sven-n:
Thanks for pointing that out - I wasn't aware of this issue 🙈

nitoygo:
Sure thing. I'm just familiarizing myself with the source code :D
Is it somewhere related to here?
![image](https://github.com/user-attachments/assets/91c5eb36-48d4-4df0-bd31-3fc650ea92db)

Issue #508, opened 2024-10-23
Repair Not Working on Gywen Pendant of Ability


nitoygo:
**Describe the bug**
Repairing Gywen Pendant of Ability does not work

**To Reproduce**
1. Login using 'ancient' and select 'ancientElf'
2. Inspect pendant (initial durability of Gywen Pendant is 50/70)
3. Click Repair Icon (Hammer) and apply action to the pendant
4. Gywen Pendant remains damaged (no effect)

**Expected behavior**
Gywen Pendant should be 70/70 after repair

**Screenshots**
![image](https://github.com/user-attachments/assets/2240f40d-2a9b-462c-995c-973b64226f79)

**Additional context**
* Tried on Broy's Pendant: OK
* Tried on Vicious Ring: OK

sven-n:
Gywen Pendant is somehow special, because it is ancient, but has no ancient bonus option. However, the ItemExtensions.IsAncient() is checking the bonus :-/

sven-n:
Current:
```csharp
    public static bool IsAncient(this Item item)
    {
        return item.ItemOptions.Any(link => link.ItemOption?.OptionType == ItemOptionTypes.AncientBonus);
    }
```

Fix:
```
    public static bool IsAncient(this Item item)
    {
        return item.ItemSetGroups.Any(itemSet => itemSet.AncientSetDiscriminator > 0);
    }
```

You can create a PR if you want. I'll try to fix this as soon as I can

nitoygo:
How about making our own business rule that gywen will have a bonus? 😂

PR #512, opened 2024-10-24
Fixed serialization of ancient item appearance

sven-n:


sven-n:
Fixes #511

sven-n:
Fixes #508, too

PR #515, opened 2024-10-28
Multiple warnings fix in GameLogic project

ze-dom:
This addresses most of the intellisense warnings for the GameLogic project.

If this is OK, then I can either do the rest of projects in this PR or do a different one for each project.


src/GameLogic/DuelRoom.cs
```diff
@@ -139,6 +219,11 @@ public async ValueTask RemoveSpectatorAsync(Player spectator)
         await spectator.RemoveInvisibleEffectAsync().ConfigureAwait(false);
     }
 
+    /// <summary>
+    /// Starts the duel.
+    /// </summary>
+    /// <returns>A <see cref="Task"/>.</returns>
```
sven-n, line 225: You can remove <returns> for Task and ValueTask.

src/GameLogic/DuelRoom.cs
```diff
@@ -69,29 +127,44 @@ public byte ScoreRequester
...
+            this._scoreOpponent = value;
             if (value >= this._maximumScore)
             {
                 this.State = DuelState.DuelFinished;
             }
         }
     }
 
-    //public bool IsAccepted { get; set; }
+    // public bool IsAccepted { get; set; }
```
sven-n, line 146: commented code can be removed.

sven-n:
Thanks for cleaning up :)

Issue #518, opened 2024-11-02
An easier way to browse configurations


Niter88:
Problem example:
Today, let's say I want to configure the Item Drop Chance for every quest item.
I have to: open "Full Configuration" > scroll to "Drop item Groups" > browse and find the item I'll edit > clicking edit will open it on the same page, forcing me to go through every step again for the next item. The same applies to other configurations.

Someone has suggested open edits in a modal but I think it'll be great to open it as a new tab, then you can open various tabs and even compare them side by side. A new tab will also be simpler to implement because we'll only alter the button to open it in a new tab, no need to alter the edit page itself. Even enable the middle mouse click as a new tab and leaving everything else as it was would be a great solution.

I'd like us to discuss this feature, if you had any other ideas that would be amazing.

Niter88:
just to clarify, I'll start working on this issue

PR #519, opened 2024-11-02
Better navigation on "Edit"

Niter88:
it was made for issue #518 

The "Edit" are now links, they can be opened in the current tab or in a new tab.
PS. I didn't alter the Merchants "Edit" button since it was not simply an URL, I'll have to look further into it.

I also added a "Drop Item Groups" in the navbar. It'll be easier to use it this way.

### DIFF
--- a/src/Web/AdminPanel/Components/Form/ItemTable.razor
+++ b/src/Web/AdminPanel/Components/Form/ItemTable.razor
@@ -56,7 +56,7 @@
                                         urlPrefix = "edit-account";                                        
                                     }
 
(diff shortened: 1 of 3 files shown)


Niter88:
Hey @sven-n , don't mark the issue as solved yet, I'll look into the breadcrumb later

nitoygo:
If everyone's ok with this new tab implementation, can we merge it for now, and do another PR for the breadcrumbs? xD

sven-n:
Okay, I'll take a look as soon as I can :)

PR #540, opened 2024-11-28
DL Summon Party Skill

nitoygo:
- [x] Implement core skill (DL can teleport all party members to his current location)
- [x] Notify members that summon is ongoing
- [x] Exclude members that are trading or interacting with NPC
- [x] Exclude members that are not yet qualified to the current map of the summoner DL
- [x] Exclude members that are still in a mini-game / Disable summon if the DL is still in a mini-game

Other to-dos:
- [ ] (Optional) Party member moved during summoning countdown


src/GameLogic/PlayerActions/Skills/NovaSkillStartAction.cs
```diff
@@ -11,62 +11,50 @@ namespace MUnique.OpenMU.GameLogic.PlayerActions.Skills;
 using MUnique.OpenMU.GameLogic.Views.World;
 
 /// <summary>
-/// The nova skill action.
+/// The start nova skill action.
 /// </summary>
-public class NovaSkillAction
+public class NovaSkillStartAction : TargetedSkillActionBase
```
sven-n, line 16: That's a bit strange, because nova start is not a targetted skill

src/GameServer/MessageHandler/TargetedSkillHandlerPlugIn.cs
```diff
@@ -13,21 +13,41 @@ namespace MUnique.OpenMU.GameServer.MessageHandler;
 using MUnique.OpenMU.PlugIns;
 
 /// <summary>
-/// Handler for targeted skill packets.
+/// Reprersents the targeted skill packet handler.
+/// </summary>
+internal partial class TargetedSkillHandlerPlugIn
```
sven-n, line 18: Why partial in the same file?

src/GameServer/MessageHandler/TargetedSkillHandlerPlugIn.cs
```diff
@@ -13,21 +13,41 @@ namespace MUnique.OpenMU.GameServer.MessageHandler;
...
+            .SelectMany(a => a.GetTypes())
+            .Where(t => pluginType.IsAssignableFrom(t) && !t.IsInterface && !t.IsAbstract)
+            .Select(t => (ITargetedSkillAction)Activator.CreateInstance(t)!)
+            .ToList();
+
+        foreach (var plugin in pluginInstances)
+        {
+            this._handlers[plugin.Key] = plugin;
+        }
+    }
```
sven-n, line 40: The plugin framework can do similar stuff already. You'd probably want to implement strategy plugins. To get the plugin manager, you can go over `player.GameContext.PlugInManager`. I think there is similar stuff for the area skills.

src/GameLogic/PlayerActions/Skills/NovaSkillStartAction.cs
sven-n: I think I got it... Nova uses packets for targetted skills

src/GameLogic/PlayerActions/Skills/TargetedSkillActionBase.cs
```diff
@@ -0,0 +1,22 @@
+﻿// <copyright file="TargetedSkillActionBase.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic.PlayerActions.Skills;
+
+/// <summary>
+/// Action to perform a skill which is explicitly aimed to a target.
+/// </summary>
+public abstract class TargetedSkillActionBase : ITargetedSkillAction
```
sven-n, line 10: Why an abstract class without any implementation?

sven-n (changes requested):

src/GameServer/MessageHandler/TargetedSkillHandlerPlugIn.cs
nitoygo: I was trying to segregate the interface implementation, and own implementation.
I think it was because of the code compliance warnings about arrangement of fields that i decided to do partials

src/GameLogic/PlayerActions/Skills/TargetedSkillActionBase.cs
nitoygo: This was intended to be a pure virtual class, the default implementation is in the TargetedSkillActionDefault (this is the old one, just added interface)

src/GameServer/MessageHandler/TargetedSkillHandlerPlugIn.cs
nitoygo: Now removed.

src/GameServer/MessageHandler/TargetedSkillHandlerPlugIn.cs
nitoygo: Updated to use the PlugInManager as suggested.

src/GameLogic/Player.cs
```diff
@@ -922,6 +956,35 @@ public async ValueTask WarpToAsync(ExitGate gate)
         // Then ClientReadyAfterMapChange is called.
     }
 
+    /// <summary>
+    /// Moves the player to the specified map and coordinates.
+    /// </summary>
+    /// <param name="map">The map to which the player should be warped.</param>
+    /// <param name="position">The coordinates on the target map.</param>
+    public async ValueTask WarpToMapAsync(GameMap map, Point position)
```
sven-n, line 964: It looks similar to WarpToAsync. Can you refactor WarpToAsync to call this method?
Also, WarpToAsync calls PlaceAtGate which also moves the summoned monster to the new coordinates. We'll need that for the summon party skill as well.
Alternatively: refactor TeleportToMapAsync to smth. like TeleportToGateAsync. Then remove WarpToMapAsync, define a temporary ExitGate object for the area of the dark lord and do something like player.TeleportToGateAsybc -> WarpToAsync

nitoygo: Yeah, I actually copied this from WarpToAsync. But instead of calling PlaceAtGate(), I added them manually here which is also what PlaceAtGate() does (except it does not randomize anymore).
```
        this.SelectedCharacter!.CurrentMap = map.Definition;
        this.SelectedCharacter.PositionX = position.X;
        this.SelectedCharacter.PositionY = position.Y;
```
The problem I have with PlaceAtGate() is that it will try to randomize another set of X,Y when I already have a prepared X,Y beforehand.

nitoygo: Anyway, I just realized I could make X1 = X2 Y1 = Y2 so the randomization in PlaceAtGate won't matter.

nitoygo: Updated as suggested.

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// <inheritdoc />
+    public override async ValueTask PerformSkillAsync(Player player, IAttackable target, ushort skillId)
+    {
+        var skillEntry = player.SkillList?.GetSkill(skillId);
+        if (skillEntry?.Skill is null)
+        {
+            return;
+        }
+
+        if (player.Party == null)
```
sven-n, line 35: Please check with `is null`

nitoygo: Updated. Also, latest commit already has map requirement checking before warping the target member.

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
```diff
@@ -0,0 +1,147 @@
...
+
+                if (!player.IsAlive || player.IsAtSafezone())
+                {
+                    await player.Party.SendChatMessageAsync("Summoning canceled.", player.Name).ConfigureAwait(false);
+                    return;
+                }
+
+                await Task.Delay(1000, cancellationToken).ConfigureAwait(false);
+
+                targetPlayers = targetPlayers.Where(this.CanTargetBeTeleported);
```
sven-n, line 79: Sorry to bother you again 😅
The check `this.CanTargetBeTeleported` is not actually be executed inside of the loop, but when enumerating the players.
To fix that, `targetPlayers` should be a List, so just add a ToList() in the first line.
Then in the loop, call `targetPlayers.RemoveAll(t => !this.CanTargetBeTeleported(t))`

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
```diff
@@ -0,0 +1,147 @@
...
+                }
+            }
+
+            if (player.CurrentMap!.Definition.TryGetRequirementError(targetPlayer, out var errorMessage))
+            {
+                await targetPlayer.InvokeViewPlugInAsync<IShowMessagePlugIn>(p => p.ShowMessageAsync(errorMessage, Interfaces.MessageType.BlueNormal)).ConfigureAwait(false);
+                continue;
+            }
+
+            _ = Task.Run(() => targetPlayer.TeleportToMapAsync(player.CurrentMap!, targetPoint));
```
sven-n, line 132: No need for Task.Run here. But a try catch around that would be great, so that an error from one player is not affecting the next.

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            return;
+        }
+
+        await player.ForEachWorldObserverAsync<IShowSkillAnimationPlugIn>(p => p.ShowSkillAnimationAsync(player, player, skillEntry.Skill.Number, true), true).ConfigureAwait(false);
+
+        var cancellationTokenSource = new SkillCancellationTokenSource();
+        player.SkillCancelTokenSource = cancellationTokenSource;
+
+        _ = this.RunSummonPartyAsync(player, cancellationTokenSource);
```
sven-n, line 50: You could add a Task.Run here

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
nitoygo: I think it's good to check every countdown (check per one second) in the loop.  
This will update the targetPlayers that can be summoned.

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
nitoygo: Remvoed Task.Run. I did not add try catch since the TeleportToMapAsync already catches possible exceptions.

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
nitoygo: Added.

src/GameLogic/PlayerActions/Skills/SummonPartySkillPlugin.cs
nitoygo: It's possible that for every second, a party member is excluded, then for the next second, another target is excluded, then after the timer expires, only the party members that are not violating any rules will be summoned.

nitoygo: Filtering updated as suggested.

sven-n (approved):

sven-n:
Thank you again 🚀

PR #550, opened 2024-12-09
Fixed nova skill not activating

nitoygo:
Server side fix (see client side fix in https://github.com/sven-n/MuMain/pull/79/commits/9234a479f7a789cc1f96676f140dc4b5887dbe4a)

### DIFF
--- a/src/GameLogic/PlayerActions/Skills/NovaSkillStartPlugin.cs
+++ b/src/GameLogic/PlayerActions/Skills/NovaSkillStartPlugin.cs
@@ -19,6 +19,8 @@ using MUnique.OpenMU.PlugIns;
 [Guid("e966e7eb-58b8-4356-8725-5da9f43c1fa4")]
 public class NovaSkillStartPlugin : TargetedSkillPluginBase
 {
+    private const ushort NovaEndSkillId = 40;
+
     private static readonly TimeSpan NovaStepDelay = TimeSpan.FromMilliseconds(500);
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/GameLogic/PlayerActions/Skills/NovaSkillStartPlugin.cs
```diff
@@ -38,7 +40,7 @@ public override async ValueTask PerformSkillAsync(Player player, IAttackable tar
             return;
         }
 
-        var skillEntry = player.SkillList?.GetSkill(skillId);
+        var skillEntry = player.SkillList?.GetSkill(NovaEndSkillId);
```
nitoygo, line 43: The nova end skill should be used for attacking the targets.

src/GameServer/MessageHandler/TargetedSkillHandlerPlugIn.cs
```diff
@@ -45,11 +45,6 @@ public virtual async ValueTask HandlePacketAsync(Player player, Memory<byte> pac
     /// <param name="targetId">The target identifier.</param>
     protected async ValueTask HandleAsync(Player player, ushort skillId, ushort targetId)
     {
-        if (player.SkillList is null || !player.SkillList.ContainsSkill(skillId))
```
nitoygo, line 48: This check is already done in the targeted skill plugins (and the default/base plugin).  
Reason for removing this check is that the nova start skill is not registered in the user skills - only the nova end)

Issue #534, opened 2024-11-24
Chaos Machine Item Recovery


nitoygo:
**Describe the bug**
Items left in chaos machine can not be recovered when game crashes or on disconnection

**To Reproduce**
1. Place items for chaos mixing
2. // Disconnect or crash happens
3. Reconnect 
4. // Items are lost

**Expected behavior**
Server has a way or an option for the player to recover items left unintentionally in the chaos machine

sven-n:
I think the best would be to handle this like the trade. When a disconnect happens, the items should be moved back to the original position before saving the account.

nitoygo:
Agreed. However, for the case where the mix already happened, it might need special handling (?). No idea yet what happens to the persistence of that new/upgraded item.

sven-n:
Yea, that needs some extra logic for sure

PR #557, opened 2024-12-21
Recover items placed in NPC when player is disconnected

nitoygo:
Resolves: https://github.com/MUnique/OpenMU/issues/534

Notes:
- [x] Chaos Goblin Failed mix - can not be recovered
- [x] Chaos Goblin Successful mix - new/upgraded item can be recovered
- [x] Jerridon (JOH restore)


nitoygo:
When trying to disconnect after a failed/successful mix:
```
2024-12-22 03:08:23.255 +08:00 [Error] [MUnique.OpenMU.GameServer.GameServer] Couldn't Save at Disconnect. Player: "0 - [Server 0]"
Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException: The database operation was expected to affect 1 row(s), but actually affected 0 row(s); data may have been modified or deleted since entities were loaded.
```

PR #579, opened 2025-01-14
Template Accounts

sven-n:
TODO:
 - [x] Add EF Core migration (currently testing in demo mode)
 - [x] ~~Testing: Add a big amount of players with a test client~~

Limitations:
* The player store state cannot be restored for the original season 6 client. The client crashes if we try that.


sven-n:
Merging this for now - testing the performance shouldn't be a task of the PR.

PR #586, opened 2025-02-13
Add dark horse options & major dmg calcs refactoring

ze-dom:
To do:
- [ ] Add Update Plugins

PR #587, opened 2025-02-23
Added item drop configuration page

sven-n:


ze-dom:
Note: there are some bugs on some options (excellent, particularly) which I will address on the [PR#586](https://github.com/MUnique/OpenMU/pull/586). I hope to finish it in the next few days. Just to let you know because some rework might come up if you do this first.

PR #589, opened 2025-03-04
Generalize level-adding jewels plugin

DavidTs93:
Adds an option to choose how many levels will be added per jewel - useful for new possible jewels (like Dark Bless, Dark Soul, Kundun, Level +10~15, etc.).
The new option - `JewelAddLevels` - allows to choose how many levels to add per usage, along with a change to the main plugin - `UpgradeItemLevelJewelConsumeHandlerPlugIn` - which adds the amount of levels specified.
The change also ensures that the minimum level is enforced.

### DIFF
--- a/src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelConfiguration.cs
+++ b/src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelConfiguration.cs
@@ -7,12 +7,15 @@
 namespace MUnique.OpenMU.GameLogic.PlayerActions.ItemConsumeActions;
 
 using System.ComponentModel.DataAnnotations;
+using MUnique.OpenMU.DataModel.Configuration.Items;
 
 /// <summary>
 /// Configuration for a <see cref="UpgradeItemLevelJewelConsumeHandlerPlugIn{TConfig}"/>.
 /// </summary>
 public class UpgradeItemLevelConfiguration
 {
+    private byte _levelAmount = 1;
+
     /// <summary>
     /// Gets or sets the success rate percentage.
     /// </summary>
@@ -36,17 +39,35 @@ public class UpgradeItemLevelConfiguration
     [Display(Name = "Item Maximum Level", Description = "The maximum item level which the item has to have before applying the jewel.")]
     public byte MaximumLevel { get; set; }
 
-    private byte _levelAmount = 1;
-
     /// <summary>
     /// Gets or sets the amount of levels which the item will be upgraded by.
     /// </summary>
-    [Display(Name = "Add Levels Amount", Description = "How many times to add a level to the item (will become MaximumLevel + 1).")]
+    [Display(Name = "Add Levels Amount", Description = "How many times to add a level to the item (will become maximum allowed level).")]
     public byte LevelAmount
     {
         get => this._levelAmount;
         set => this._levelAmount = value;
     }
+
+    /// <summary>
+    /// Gets the items which are allowed to be upgraded. If empty, all items are allowed except those in <see cref="DisallowedItems"/>.
+    /// </summary>
+    public ICollection<ItemDefinition> AllowedItems { get; set; } = new List<ItemDefinition>();
+
+    /// <summary>
+    /// Gets the items which are not allowed to be upgraded.
+    /// </summary>
+    public ICollection<ItemDefinition> DisallowedItems { get; set; } = new List<ItemDefinition>();
+
+    /// <summary>
+    /// Gets the drop item groups which are allowed to be upgraded. If empty, all drop item groups are allowed except those in <see cref="DisallowedDropItemGroups"/>.
+    /// </summary>
+    public ICollection<DropItemGroup> AllowedDropItemGroups { get; set; } = new List<DropItemGroup>();
+
+    /// <summary>
+    /// Gets the drop item groups which are not allowed to be upgraded.
+    /// </summary>
+    public ICollection<DropItemGroup> DisallowedDropItemGroups { get; set; } = new List<DropItemGroup>();
 
     /// <summary>
     /// Gets or sets the item level after which the item will drop to level 0 when it fails.
--- a/src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelConfiguration.cs
+++ b/src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelConfiguration.cs
@@ -37,6 +37,12 @@ public class UpgradeItemLevelConfiguration
     public byte MaximumLevel { get; set; }
 
     /// <summary>
+    /// Gets or sets the amount of levels which the item will be upgraded by.
+    /// </summary>
+    [Display(Name = "Jewel Add Levels", Description = "The amount of levels to add to the item (up to the Maximum Level).")]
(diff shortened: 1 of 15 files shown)


DavidTs93:
What failed? Can't view the error

sven-n:
`D:\a\1\s\src\GameLogic\PlayerActions\ItemConsumeActions\UpgradeItemLevelJewelConsumeHandlerPlugIn.cs(49,56): error CS8602: Dereference of a possibly null reference. [D:\a\1\s\src\GameLogic\MUnique.OpenMU.GameLogic.csproj]`

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelJewelConsumeHandlerPlugIn.cs
```diff
@@ -52,8 +52,13 @@ protected override bool ModifyItem(Item item, IContext persistenceContext)
         }
 
         this.Configuration ??= (TConfig)this.CreateDefaultConfig();
+        if (item.Level < this.Configuration.MinimumLevel)
+        {
+            return false;
+        }
 
-        if (item.Level > this.Configuration.MaximumLevel)
+        var levelsAdd = Math.Min(this.Configuration.LevelsAdd, this.Configuration.MaximumLevel - item.Level);
```
sven-n, line 60: I've run the unit tests `ItemConsumptionTest` to validate the changes and I think there is a bug. `this.Configuration.MaximumLevel` describes the maximum level of the source item.
E.g. a bless can be applied to items until level 5 to increase them to 6. With this code, the `levelsAdd` will be 0 in this case, and the method returns false.
Additionally, you should consider, that the `Configuration.LevelsAdd` can be 0, because this property was added to an existing configuration class. In this case, we should assume it's 1 by default.

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelConfiguration.cs
```diff
@@ -36,6 +36,12 @@ public class UpgradeItemLevelConfiguration
     [Display(Name = "Item Maximum Level", Description = "The maximum item level which the item has to have before applying the jewel.")]
     public byte MaximumLevel { get; set; }
 
+    /// <summary>
+    /// Gets or sets the amount of levels which the item will be upgraded by.
+    /// </summary>
+    [Display(Name = "Jewel Add Levels", Description = "The amount of levels to add to the item (up to the Maximum Level).")]
+    public byte LevelsAdd { get; set; }
```
sven-n, line 43: Can you rename this to `LevelAmount`?

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelJewelConsumeHandlerPlugIn.cs
```diff
@@ -64,9 +69,9 @@ protected override bool ModifyItem(Item item, IContext persistenceContext)
             percent += this.Configuration.SuccessRateBonusWithLuckPercentage;
         }
 
-        if (this._randomizer.NextRandomBool(percent))
+        if (percent >= 100 || this._randomizer.NextRandomBool(percent))
```
sven-n, line 72: The change is unneeded, because NextRandomBool will also return `true` when `percent` is `100`.
However, we could change `Rand.NextRandomBool` so that value 'higher than' `100` and `lower than` 0 is considered, too

sven-n (changes requested):

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelJewelConsumeHandlerPlugIn.cs
DavidTs93: Fixed

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelJewelConsumeHandlerPlugIn.cs
DavidTs93: Removed (will updated the randomizer in later PR)

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelConfiguration.cs
DavidTs93: Done

DavidTs93:
@sven-n please don't merge yet, need to add some stuff

src/Persistence/EntityFramework/ConnectionSettings.xml
sven-n, line 1: please revert this file

src/Startup/Properties/launchSettings.json
sven-n, line 1: please revert this file

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelJewelConsumeHandlerPlugIn.cs
```diff
@@ -52,8 +52,34 @@ protected override bool ModifyItem(Item item, IContext persistenceContext)
...
+
+        if (this.Configuration.DisallowedDropItemGroups.Any(dig => dig.PossibleItems?.Contains(item.Definition!) == true))
+        {
+            return false;
+        }
+
+        if (this.Configuration.AllowedDropItemGroups.Any() && !this.Configuration.AllowedDropItemGroups.Any(dig => dig.PossibleItems?.Contains(item.Definition!) == true))
+        {
+            return false;
+        }
```
sven-n, line 78: Can you give an example when you'd need this with DropItemGroups?

sven-n (changes requested):

src/Persistence/EntityFramework/ConnectionSettings.xml
DavidTs93: Yeah, sorry, forgot to add them to ignore...

src/Startup/Properties/launchSettings.json
DavidTs93: All good

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelJewelConsumeHandlerPlugIn.cs
sven-n: Okay, I get it. However, "DropItemGroups" are really only for drops. Other usages would be confusing, imho.
To use it in other places, like chaos goblin, it would have to be added to even more places which makes it even more confusing.
Can we just leave that out. for now?
I'm open to add something like more general "ItemGroup" later in another PR.

sven-n (changes requested):

src/GameLogic/PlayerActions/ItemConsumeActions/UpgradeItemLevelJewelConsumeHandlerPlugIn.cs
DavidTs93: That is in fact a better option, I'll remove it

sven-n (approved):

sven-n:
Thank you :)

PR #599, opened 2025-03-29
Fix add comboStep

schec285:
Add some code to accept combo skills by default in the combo state machine. 

Discord
https://discord.com/channels/669595902750490698/959730931395076116/1353744025974214656

### DIFF
--- a/src/GameLogic/ComboStateMachine.cs
+++ b/src/GameLogic/ComboStateMachine.cs
@@ -68,7 +68,7 @@ public sealed class ComboStateMachine : StateMachine
             await this.TryAdvanceToAsync(this.InitialState).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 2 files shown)


sven-n:
Hey, thanks for the PR.
You've added the Master skills to the combo skill combinations list. However, I wanted to fix this in another way. It would be better to handle combo skills in the combo state machine, so that it accepts the master skill if the replaced skill is included in the combo skill combination list.
That way, it's cleaner in the data with less repetitions. Also, a player cannot do the normal skill and then the same enhanced skill to complete the combo.

schec285:
I see, thank you! I'll give that method a try.

schec285:
@sven-n 
I have revised and committed the changes again. I appreciate it if you could review them at your convenience.

sven-n (changes requested):

schec285:
@sven-n 
Thank you!
With my current knowledge, I couldn't have gotten that far. I really learned a lot.
I've submitted the pull request again, so please take a look when you have a chance.

Issue #617, opened 2025-06-14
Excellent Drop sword no skill


nitoygo:
**Describe the bug**
Excellent weapon drop (Rune Blade from BOK) did not contain skill

**To Reproduce**
(Note: might be difficult to replicate since it's a random drop)
1. Hunted golden tantalos, pick up BOK
2. Drop BOK
3. Picked up Excellent Rune Blade with no skill

**Expected behavior**
Excellent weapons, if can contain skill, should have skill 

**Screenshots**
![Image](https://github.com/user-attachments/assets/95cbacde-c9ae-45c1-b1bc-fbda2272f985)

PR #621, opened 2025-06-25
Issue #617 - Excellent weapon W/O Skill from item box drop 

ze-dom:
Fix for reported bug in issue #617.

Also fixes the appropriate drop sound effect which was not being bypassed and not sent to client.


sven-n:
Thanks :)

PR #636, opened 2025-08-13
Tests update and fix

ze-dom:
Remembered to run the tests and found some issues.

Due to previous changes the SkillListTests.cs tests were raising an exception [here](https://github.com/MUnique/OpenMU/blob/a4fad3273fddfd7da76338be427a0e23aad4a1d0/src/GameLogic/SkillList.cs#L125).

### DIFF
--- a/tests/MUnique.OpenMU.Tests/SkillListTest.cs
+++ b/tests/MUnique.OpenMU.Tests/SkillListTest.cs
@@ -41,8 +41,8 @@ public class SkillListTest
     {
         var player = await TestHelper.CreatePlayerAsync().ConfigureAwait(false);
         var skillList = player.SkillList as SkillList;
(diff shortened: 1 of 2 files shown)


tests/MUnique.OpenMU.Tests/SkillListTest.cs
```diff
@@ -32,30 +34,18 @@ public async ValueTask LearnedSkillAsync()
...
     [Test]
-    public async ValueTask ItemSkillAsync()
-    {
-        var player = await TestHelper.CreatePlayerAsync().ConfigureAwait(false);
-        var item = this.CreateItemWithSkill();
-        item.Durability = 1;
-        await player.Inventory!.AddItemAsync(0, item).ConfigureAwait(false);
-        var skillList = new SkillList(player);
-        Assert.That(skillList.ContainsSkill(ItemSkillId), Is.True);
-    }
```
ze-dom, line 46: I didn't find any difference between this test and the one right after. But please confirm :wink:

PR #650, opened 2025-10-07
Added a max value of 70% for soul barrier

ze-dom:
For reset chars it was possible to go above 100% damage reduction, which could also cause negative damage values and instantly kill the char.

### DIFF
--- a/src/GameLogic/Attributes/Stats.cs
+++ b/src/GameLogic/Attributes/Stats.cs
@@ -791,7 +791,10 @@ public class Stats
     /// <summary>
     /// Gets the soul barrier skill damage receive decrement attribute definition.
     /// </summary>
-    public static AttributeDefinition SoulBarrierReceiveDecrement { get; } = new(new Guid("CBC5404A-6232-4BF3-9B85-9AA0AE0F9BA4"), "Soul Barrier Damage Receive Decrement", "The soul barrier skill receive damage multiplier which is multiplied with the final damage and subtracted from it.");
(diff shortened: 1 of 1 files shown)


src/GameLogic/Attributes/Stats.cs
```diff
@@ -791,7 +791,10 @@ public class Stats
     /// <summary>
     /// Gets the soul barrier skill damage receive decrement attribute definition.
     /// </summary>
-    public static AttributeDefinition SoulBarrierReceiveDecrement { get; } = new(new Guid("CBC5404A-6232-4BF3-9B85-9AA0AE0F9BA4"), "Soul Barrier Damage Receive Decrement", "The soul barrier skill receive damage multiplier which is multiplied with the final damage and subtracted from it.");
+    public static AttributeDefinition SoulBarrierReceiveDecrement { get; } = new(new Guid("CBC5404A-6232-4BF3-9B85-9AA0AE0F9BA4"), "Soul Barrier Damage Receive Decrement", "The soul barrier skill receive damage multiplier which is multiplied with the final damage and subtracted from it.")
+    {
+        MaximumValue = 0.7f,
+    };
```
ze-dom, line 797: https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/BuffSkillEffect.xml#L19

sven-n:
It makes sense, but it's questionable if webzen itself has implemented such a limit 😅
Can you add a configuration update plugin for this change, please?

ze-dom:
No, I don't think WZ had a need for it :stuck_out_tongue: 
I have more stuff in the works, so I think it's better to not do an update plugin just for this?

Issue #645, opened 2025-09-04
Spawn locked after dying


santo998:
After dying in Tarkan, my wizard spawned in a locked area. He can't walk.
Only way to solve this manually is to use the move command.

**To Reproduce**
You have to be lucky to spawn there. But you can check spawn areas and remove (148, 52) coordinates

**Expected behavior**
Being able to walk.

**Screenshots**
<img width="802" height="627" alt="Image" src="https://github.com/user-attachments/assets/a550bef7-a9c6-48c5-be98-087edcf9be37" />

**Desktop (please complete the following information):**
 - OS: Windows 11

Niter88:
I'm looking into it

Niter88:
I've made the PR #651 that fixes it in the default configuration. New servers will have this fixed but you'll still need to change your current configs.

On the [admin panel](http://localhost/), go to "Game Configuration" > "Maps"
Find Tarkan and click "Edit" on it, then click on "Map Editor" at the top of the page

Select the spawn region and change Y1 from 54 to 63 and hit save. Reload your server to apply all configs.

PR #651, opened 2025-10-15
fix for the issue #645 where the player would spawn in an incorrect…

Niter88:
…location and be stuck inside the collisions.
The Spawn region's Y needs to be at least 62 to avoid being stuck inside a part of the map, changed to 63.
This alters the default value for this, for anyone who's already running the server, change the Y1 location for Tarkan's spawn from 54 to 63 and save.


sven-n:
Thanks

PR #658, opened 2025-11-16
Fix Docker deployment issues for all-in-one setup

xdohko:
This PR addresses several critical issues in the Docker deployment that prevent the all-in-one setup from working properly on fresh installations.

### 1. Database Initialization Race Condition
**Problem**: The `openmu-startup` container was starting before PostgreSQL was ready to accept connections, resulting in errors:
```
57P03: the database system is starting up
```

**Solution**: Added a health check to the PostgreSQL service and configured `openmu-startup` to wait for `service_healthy` condition instead of just the container being started.

### 2. Missing Entity Framework Migrations
**Problem**: The Docker image didn't include EF migration files, causing the database schema creation to fail with errors like:
```
42P01: relation "config.GameConfiguration" does not exist
```

**Solution**: Modified the Dockerfile to explicitly copy the `Migrations` directory from the build stage to the final image.

### 3. PostgreSQL Version Compatibility
**Problem**: Using `postgres:latest` could pull incompatible versions.

**Solution**: Pinned to `postgres:16` for consistent behavior.

### 4. Obsolete Docker Compose Syntax
**Problem**: Warnings about deprecated `version` attribute in docker-compose files.

**Solution**: Removed the obsolete `version: '3'` declarations.

## Testing
These changes have been tested with a fresh deployment and successfully:
- Create the database schema using migrations
- Initialize the database with default data
- Start all services without errors

## Files Changed
- `src/Startup/Dockerfile`: Added migration files copy
- `deploy/all-in-one/docker-compose.yml`: Added health check, updated postgres version, removed obsolete version
- `deploy/all-in-one/docker-compose.override.yml`: Removed obsolete version attribute

### DIFF
--- a/deploy/all-in-one/docker-compose.yml
+++ b/deploy/all-in-one/docker-compose.yml
@@ -8,7 +8,8 @@ services:
       - ./nginx/nginx.dev.conf:/etc/nginx/nginx.conf:ro
       - ./.htpasswd:/etc/nginx/.htpasswd
     depends_on:
-      - openmu-startup
+      openmu-startup:
+        condition: service_healthy
 
   openmu-startup:
     build:
@@ -36,6 +37,12 @@ services:
     depends_on:
       database:
         condition: service_healthy
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8080/health"]
+      interval: 10s
+      timeout: 5s
+      retries: 10
+      start_period: 60s
 
   database:
     image: postgres:16
--- a/src/Startup/Dockerfile
+++ b/src/Startup/Dockerfile
@@ -26,6 +26,8 @@ WORKDIR /app
 COPY --from=publish /app/publish .
 COPY --from=build /src/Persistence/EntityFramework/Migrations ./Persistence/EntityFramework/Migrations
 
+RUN apk add --no-cache curl
+
 RUN mkdir /app/logs
 RUN chmod 777 /app/logs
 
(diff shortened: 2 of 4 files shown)


r7nd7:
Thanks for the excellent suggestions! I've applied all the recommended improvements to the health check:

✅ Changed from `CMD-SHELL` to `CMD` exec form for better security and efficiency
✅ Added `start_period: 30s` to prevent premature health check failures during bootstrap
✅ Adjusted timing parameters (10s interval, 5 retries) for better reliability

The updated health check is now more robust and follows Docker best practices.

r7nd7:
## Additional Improvements: Health Check Endpoint

I've added a new commit that further improves the Docker setup with proper health checks.

### Changes Added

**1. ASP.NET Core Health Check Endpoint**
- Added `/health` endpoint in `Program.cs` and `WebApplicationExtensions.cs`
- Returns `200 OK` with "Healthy" response when the application is ready
- Standard endpoint for monitoring and orchestration

**2. Docker Health Checks**
- Installed `curl` in Alpine image for health check testing
- OpenMU container now uses `curl -f http://localhost:8080/health`
- Nginx waits for OpenMU to be `healthy` before starting

**3. Proper Dependency Chain**
```
database (healthy) → openmu-startup (healthy) → nginx-80 (starts)
```

### Benefits
- Eliminates nginx "Connection refused" errors during startup
- Provides standard `/health` endpoint for monitoring
- Proper container orchestration with health checks
- No more initialization errors in logs

### Testing
Tested successfully with:
```bash
curl http://localhost:8080/health
# Returns: Healthy
```

All containers now start cleanly with no errors.

PR #659, opened 2025-11-18
feat: add max magic speed

5Miro:
Previously, AttackSpeed was capped at 200 while MagicSpeed had no cap,
causing an imbalance where physical skills (like Twisting Slash, Power Slash)
stopped scaling at high agility levels while magic spells (like Evil Spirit)
continued to scale indefinitely.

Changes:
- Added MagicSpeed MaximumValue of 200 in Stats.cs (previously uncapped)
- Added MagicSpeed MaximumValue initialization in FixAttackSpeedCalculationUpdate

This ensures both physical skills and magic spells scale consistently, balancing gameplay at high agility levels. Both values remain editablein the admin panel through the AttributeDefinition MaximumValue property.

### DIFF
--- a/src/GameLogic/Attributes/Stats.cs
+++ b/src/GameLogic/Attributes/Stats.cs
@@ -365,7 +365,7 @@ public class Stats
     /// </summary>
     public static AttributeDefinition AttackSpeed { get; } = new(new Guid("BACC1115-1E8B-4E62-B952-8F8DDB58A949"), "Attack Speed", string.Empty)
     {
-        MaximumValue = 1500,
+        MaximumValue = 200,
     };
 
     /// <summary>
@@ -404,7 +404,7 @@ public class Stats
     /// </summary>
     public static AttributeDefinition MagicSpeed { get; } = new(new Guid("AE32AA45-9C18-43B3-9F7B-648FD7F4B0AD"), "Magic Speed", string.Empty)
     {
-        MaximumValue = 1500,
+        MaximumValue = 200,
     };
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


sven-n:
> There is currently (hard code) a limit in Stat.cs: AttackSpeed MaximumValue = 200. If we limit MagicSpeed MaximumValue = 200 in Stat.cs, it will not be possible to increase it in the AdminPanel - Stats.MagicSpeed.MaximumValue. I think the limit in Stat.cs: AttackSpeed MaximumValue = 200 should be removed, and the Stats.MagicSpeed.MaximumValue and Stats.AttackSpeed.MaximumValue limits can be limited in the AdminPanel -> Full Configuration -> Attributes -> Edit Attributes: 'Attack Speed'; 'Magic Speed'

It's a bug, see #663 and #664

src/Persistence/Initialization/Updates/FixAttackSpeedCalculationUpdate.cs
```diff
@@ -137,6 +137,7 @@ protected override async ValueTask ApplyAsync(IContext context, GameConfiguratio
         AddStatIfNotExists(context, gameConfiguration, Stats.WalkSpeed);
 
         Stats.AttackSpeed.GetPersistent(gameConfiguration).MaximumValue = Stats.AttackSpeed.MaximumValue;
+        Stats.MagicSpeed.GetPersistent(gameConfiguration).MaximumValue = Stats.MagicSpeed.MaximumValue;
```
sven-n, line 140: This will not work when someone has already run this update.
To make this work 100% correct, you'd need to add a new separate update plugin.
However, for such a small change, we can also ignore it.

sven-n:
Thank you :)

PR #661, opened 2025-11-18
feat: money drop rate

5Miro:
Add MoneyDropRate configuration to allow independent control of money drops

This addresses the issue where increasing XP rate also increases money drops proportionally.

Adds a new MoneyDropRate property to GameConfiguration that acts as a multiplier for all money drops, allowing server administrators to balance money drops independently from the experience rate.

Changes:
- Add MoneyDropRate property to GameConfiguration (default: 1.0f)
- Initialize MoneyDropRate in GameConfigurationInitializerBase
- Update DefaultDropGenerator to apply MoneyDropRate multiplier to money drops
- Store GameConfiguration reference in DefaultDropGenerator for access to rate

### DIFF
--- a/deploy/all-in-one-traefik/docker-compose.prod.yml
+++ b/deploy/all-in-one-traefik/docker-compose.prod.yml
@@ -40,7 +40,7 @@ services:
     ports:
       - "5432"
     volumes:
-      - dbdata:/var/lib/postgresql/data #store data on volume
+      - dbdata:/var/lib/postgresql #store data on volume
 
   traefik:
     image: traefik
--- a/deploy/all-in-one-traefik/docker-compose.yml
+++ b/deploy/all-in-one-traefik/docker-compose.yml
@@ -40,7 +40,7 @@ services:
     ports:
       - "5432:5432"
     volumes:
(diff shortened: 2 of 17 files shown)


sven-n:
Hi, thanks for the PR. I'd like to implement this a bit different to be even more flexible for more options in the future.

My idea is to add two collections to the `GameConfiguration`, like we already have in `CharacterClass`:
```csharp
    /// <summary>
    /// Gets or sets the attribute combinations.
    /// </summary>
    [MemberOfAggregate]
    public virtual ICollection<AttributeRelationship> GlobalAttributeCombinations { get; protected set; } = null!;

    /// <summary>
    /// Gets or sets the base attribute values.
    /// For example the amount of health a character got without any added stat point.
    /// </summary>
    [MemberOfAggregate]
    public virtual ICollection<ConstValueAttribute> GlobalBaseAttributeValues { get; protected set; } = null!;
```

The idea is to apply both globally, for all characters, no matter which character class. With that we can easily add a `ConstValueAttribute` for `Stats.MoneyAmountRate`.

Do you think you can do that? 😅

PR #667, opened 2025-11-26
refactor: Replace invasion spawn tuples with InvasionMobSpawn record class

5Miro:
Refactor invasion plugin system to use a dedicated record class instead of tuple types, improving type safety and code maintainability.

This also includes optional spawn coordinates, instead of a default global random spawn.

Changes:
- Add InvasionMobSpawn record class with optional spawn coordinates
- Replace tuple types in BaseInvasionPlugIn with InvasionMobSpawn[]
- Add support for optional x1, x2, y1, y2 coordinates per spawn entry
- Update SpawnMobsAsync to accept IEnumerable<InvasionMobSpawn>
- Update GoldenInvasionPlugIn and RedDragonInvasionPlugIn to use new class

Benefits:
- Single class type can be used throughout codebase
- Named parameters improve code readability
- Easier to extend with additional properties in the future
- Enables grouped spawning by specifying custom coordinates

The InvasionMobSpawn class supports:
- Default behavior: spreads monsters across entire map (10-240, 10-240)
- Custom coordinates: specify exact spawn area for grouped spawning
- Map selection: MapId null = selected map, MapId set = specific map


sven-n:
Thanks :)

PR #694, opened 2026-01-29
Fix localized string converter for reading plugin configuration

ze-dom:
After the multilanguage support PR, I was having exceptions when starting up the DB. It turned out `LocalizedString`s within `PlugInConfiguration.CustomConfiguration`  couldn't be read by the `LocalizedStringJsonConverter` because those are kept as objects within the JSON data:
Screenshot from 2026-01-29 11-40-46

The `LocalizedString`s were also not showing up on the AdminPanel's edit modal.


sven-n:
Thanks! I somehow didn't see this problems during testing, but I can reproduce it when I reinitialize the database.
To fix the Json serialization, it's enough to add the JsonConverter to the JsonSerializerOptions. Unfortunately, I missed 3 out of 4 places in the `PlugInConfigurationExtensions`.
Additionally, I probably need some additional checks in the game logic code to check for empty LocalizedStrings before sending messages to players 😅
I'll add a fresh PR to address these problems.

PR #704, opened 2026-03-04
fix: map editor monster spawn

eduardosmaniotto:
Fix MapEditor delete operations and validation error messages

- Fixed EntityFrameworkContextBase.DeleteAsync to properly attach entities before deletion when they're not tracked by EF Core (occurred in non-caching contexts used by the MapEditor)
- Modified RemoveFocusedObjectAsync in MapEditor to automatically save changes after removing a spawn area, gate, or other map object
- Fixed LessThanOrEqualToAttribute error message formatting that caused a FormatException when validation failed
- Adjusted the layout to display the map and form as 2 columns on the map editor


sven-n:
Thanks :)

PR #719, opened 2026-03-24
feat: Add Kanturu Refinery Tower event implementation

apraxico:
## Summary

Implements the full server-side logic for the **Kanturu Refinery Tower** mini-game event (maps 38-40), following the same plugin patterns already used by BloodCastle and ChaosCastle.

The upstream already contained the three map definitions (`KanturuEvent.cs`, `KanturuRelics.cs`, `KanturuRuins.cs`) and the two client-side packet docs (`C1-D1-00`, `C1-D1-01`). This PR adds the missing logic layer that wires those maps into a playable event.

### New files

| File | Purpose |
|---|---|
| `KanturuContext` | Event game loop — 8 sequential phases (wave clears → Maya L/R Hand → Nightmare → Tower of Refinement) |
| `IKanturuEventViewPlugIn` | View-plugin interface for client packet dispatch |
| `KanturuEventViewPlugIn` | Sends opening-state and monster-count packets to clients |
| `KanturuGatewayPlugIn` | NPC gateway that controls event entry |
| `KanturuStartPlugIn` | Periodic task that starts the event on a configurable interval |
| `KanturuStartConfiguration` | Configuration class for the periodic start task |
| `KanturuGameServerState` | Tracks per-server event state (open / running / closed) |
| `KanturuEnterRequestHandlerPlugIn` | Handles `C1-D1-01` entry packet |
| `KanturuInfoRequestHandlerPlugIn` | Handles `C1-D1-00` info packet |
| `KanturuGroupHandlerPlugIn` | Registers the D1 packet group handler |
| `KanturuInitializer` | Seeds `MiniGameDefinition` and spawn waves at startup |

### Modified files

- `MiniGameType.cs` — added `Kanturu` enum value
- `GameConfigurationInitializer.cs` — registers `KanturuInitializer`
- `KanturuEvent.cs` — added full wave monster spawn definitions (Blade Hunters, Dreadfear, Twin Tale, Genocider, Persona, Maya hands, Nightmare)

## Event flow

```
Phase 1 → Wave of Blade Hunters + Dreadfear → Maya Left Hand
Phase 2 → Wave of Blade Hunters + Dreadfear → Maya Right Hand
Phase 3 → Wave of Dreadfear + Twin Tale → Both Maya Hands
Nightmare Prep → Genocider + Dreadfear + Persona
Nightmare → Boss fight (teleports at 75/50/25% HP)
Victory → Elphis barrier opens → Tower of Refinement
```

Players who die respawn at Kanturu Relics (Map 38) via the map's `SafezoneMap` setting.

## Test plan

- [x] `dotnet build src/` compiles without errors
- [x] `dotnet test tests/MUnique.OpenMU.Persistence.Initialization.Tests/` passes
- [x] In-game: talk to Kanturu NPC → info dialog appears
- [ ] In-game: enter with valid character (lv 350-400, Moonstone Pendant equipped)
- [x] Event progresses through all 8 phases
- [ ] Defeat: player respawns at Kanturu Relics
- [x] Victory: Tower of Refinement is accessible


src/GameLogic/MiniGames/IKanturuEventViewPlugIn.cs
sven-n, line 1: 1. Convert static classes to enums, remove numerical values. In the packet definitions, similar enums are required, where the values should be specified. Also mapping code is required to map between the enum types.
2. Use the enum types for method parameters
3. Change time parameters like `remainTimeSec` to be `TimeSpan`. The conversion to seconds can be done in the view plugin implementation.
4. Change method names from `Send...` to `Show...`. The game logic should not be aware that it's actually sending something over the network.
5. Change counts from byte to int. Conversion to byte can be done inside the view plugin implementation.

src/GameLogic/MiniGames/KanturuContext.cs
```diff
@@ -0,0 +1,943 @@
...
+    // Wave numbers matching MonsterSpawnArea.WaveNumber in KanturuEvent.cs
+    private const byte WaveMayaAppear = 0;   // Maya body rises at battle start
+    private const byte WavePhase1Monsters = 1;
+    private const byte WavePhase1Boss = 2;
+    private const byte WavePhase2Monsters = 3;
+    private const byte WavePhase2Boss = 4;
+    private const byte WavePhase3Monsters = 5;
+    private const byte WavePhase3Bosses = 6;
+    private const byte WaveNightmarePrep = 7;
+    private const byte WaveNightmare = 8;
```
sven-n, line 52: Usually, I'm not a fan of hardcoding stuff. Since it's inside a plugin it's acceptable.

src/GameLogic/MiniGames/KanturuContext.cs
```diff
@@ -0,0 +1,943 @@
...
+    // after Nightmare is defeated so players can walk to Elpis.
+    private static readonly (byte StartX, byte StartY, byte EndX, byte EndY)[] ElphisBarrierAreas =
+    [
+        (73, 144, 90, 195),
+    ];
+
+    private readonly IMapInitializer _mapInitializer;
+    private readonly TimeSpan _towerOfRefinementDuration;
+
+    private volatile KanturuPhase _phase = KanturuPhase.Open;
```
sven-n: Why volatile?

src/GameLogic/MiniGames/KanturuContext.cs
```diff
@@ -0,0 +1,943 @@
...
+    private int _waveKillCount;
+    private int _waveKillTarget;
+    private TaskCompletionSource _phaseComplete = new(TaskCreationOptions.RunContinuationsAsynchronously);
+    private volatile bool _isVictory;
```
sven-n: Why volatile?

src/GameLogic/MiniGames/KanturuContext.cs
```diff
@@ -0,0 +1,943 @@
...
+    protected override async ValueTask OnGameStartAsync(ICollection<Player> players)
+    {
+        await base.OnGameStartAsync(players).ConfigureAwait(false);
+
+        // Maya rises from the depths when the battle begins.
+        await this._mapInitializer.InitializeNpcsOnWaveStartAsync(this.Map, this, WaveMayaAppear).ConfigureAwait(false);
+
+        await this.ForEachPlayerAsync(player =>
+            player.InvokeViewPlugInAsync<IShowMessagePlugIn>(p =>
+                p.ShowMessageAsync("Maya rises from the depths of the Refinery Tower!", MessageType.GoldenCenter)).AsTask()).ConfigureAwait(false);
```
sven-n: Please move messages like this to the `PlayerMessage` resources. Then please also use the `Player.ShowLocalizedBlueMessageAsync` and `Player.ShowLocalizedGoldenMessageAsync` for this and all other player messages.

src/GameLogic/MiniGames/KanturuContext.cs
```diff
@@ -0,0 +1,943 @@
...
+
+        _ = Task.Run(() => this.RunKanturuGameLoopAsync(this.GameEndedToken), this.GameEndedToken);
+    }
+
+    /// <inheritdoc/>
+    protected override void OnMonsterDied(object? sender, DeathInformation e)
```
sven-n: 1. Make the method async void and wrap it with try-catch.
2. Then you can remove all these Task.Runs, and await the calls, e.g. to SendMonsterUserCountAsync.
3. Make sure that SendMonsterUserCountAsync doesn't throw exceptions.

src/GameServer/RemoteView/MiniGames/KanturuEventViewPlugIn.cs
sven-n, line 1: Instead of constructing every packet manually here, define them in the ServerToClientPackets.xml. Otherwise, they're not documented and not tested.
Then you can easily use the automatically generated structs to send the packets.

sven-n (changes requested):

apraxico:
> @apraxico Did you actually test the event? If my memory is right, a "Moonstone Pendant" is required to enter it, but it wasn't added 😅

Yes, I did quite a lot of testing actually. After your observations I started making adjustments, but unfortunately I haven't had time to continue with it. Keep in mind the testing I did was before the changes, so when I pick this back up during the week I'll run full tests again and make whatever adjustments are needed.

Before the adjustments you requested, the event was working end-to-end. The only missing pieces were: generating a proper event state after it ends (to prevent re-entry), the ability to loop into the event indefinitely, and the ability to enter the Refinery Tower directly once it concluded.

Here's what was already working at that point:
- Event entry, validating that the Moonstone Pendant was equipped (the validation sent a system message rather than a direct gateway message)
- Maya's appearance animation
- All waves
- Monster kill counter
- Maya Hands and Maya skills
- Fall animation
- Nightmare fight, including teleports
- Map transition to enter Elphis

I'm hoping to pick this back up during the week. I also went through some PC changes recently, which has slowed things down a bit on my end.

didiconcs:
Tested this branch hands-on (isolated worktree off current `origin/master`, `d91c7038f` + rebasing on top of the latest master). Great effort on the overall design (the phase-based `KanturuContext` mirrors the BloodCastle/ChaosCastle pattern nicely), but it doesn't currently build clean from scratch, so I wasn't able to get to the in-game testing steps that are still unchecked in the description.

## Build failure (reproducible)

```
dotnet build src/Startup/MUnique.OpenMU.Startup.csproj -c Debug
```

fails with 7x `CS1061`, e.g.:

```
KanturuEventViewPlugIn.cs(44,39): error CS1061: 'IConnection' does not contain a definition for 'SendKanturuStateInfoAsync' ...
```

(same for `SendKanturuEnterResultAsync`, `SendKanturuStateChangeAsync`, `SendKanturuBattleResultAsync`, `SendKanturuTimeLimitAsync`, `SendKanturuMonsterUserCountAsync`, `SendKanturuMayaWideAreaAttackAsync`)

## Root cause

`src/Network/Packets/MUnique.OpenMU.Network.Packets.csproj` has a `PreBuild` target that regenerates `ServerToClient/ConnectionExtensions.cs` from `ServerToClient/ServerToClientPackets.xml` via XSLT (`GenerateExtensions.xslt`) on every non-CI build. The PR adds the 9 new `SendKanturuXxxAsync` methods directly to the committed `ConnectionExtensions.cs`, and the corresponding `<Packet>` entries do exist in `ServerToClientPackets.xml` (e.g. `KanturuStateInfo`, `KanturuStateChange`), but after a clean build, the regenerated file doesn't contain those methods (`grep -c SendKanturuStateInfoAsync ConnectionExtensions.cs` → `0` post-build). So the checked-in `.cs` is currently out of sync with what the project's own generator produces from the XML that ships in the same PR.

My guess is this was tested with an incremental build (which may skip the `PreBuild` regeneration if the IDE doesn't trigger it), so it never surfaced locally. Might be worth checking the new `<Packet>` entries against `GenerateExtensions.xslt`'s expectations, or just running a fully clean build (`git clean -fdx` first) before the next push to confirm it regenerates correctly.

Happy to help dig further into the XSLT side if useful, this looks close, just needs that sync issue sorted out before the in-game testing checklist items can be verified.

foxtnaider:
Hi @apraxico - I've been running this branch on a Season 6 server and found the root cause of the `ConnectionExtensions.cs` desync @sven-n flagged, plus a second bug that keeps the event from doing anything at runtime. The event now runs end to end.

**The desync starts in the XML.** The seven Kanturu `<Packet>` elements are direct children of the root `<PacketDefinitions>`, outside `<Packets>`. `GenerateExtensions.xslt` only iterates the children of `<Packets>`, so the generator emits nothing for them - which is why the methods ended up added to the generated file by hand. Moving the seven blocks inside `<Packets>` makes a clean build produce them, and nothing has to be hand-written.

**Second bug, only visible once it compiles.** `GameContext` has no `case MiniGameType.Kanturu` in the mini game factory, so the event falls through to `default` and gets a plain `MiniGameContext`. All the logic lives in `KanturuContext` - Maya, the eight phases, the waves - so the event would start, open the map, accept players, and then do nothing. `new KanturuContext` does not appear anywhere in the tree.

With those two, plus the `ObjClassH`/`ObjClassL` arguments on `SendKanturuMayaWideAreaAttackAsync` (CS7036), the branch builds clean and the event runs: Maya rises, all three phases, both hands, Nightmare, victory screen, tower opens. Tested with two characters, full clear.

I also added two update plug-ins so existing databases get the data, since the initializers only run on a fresh DB.

Branch: https://github.com/foxtnaider/OpenMU/tree/kanturu-fixes

Happy to hand this over however is most useful - a PR into your branch, a patch, or just take the diff. Written with AI assistance and then tested in-game, per CONTRIBUTING.md.

foxtnaider:
Screen(08_30-02_15)-0001
Screen(08_30-02_16)-0002
Screen(08_30-03_19)-0000
Screen(08_30-03_22)-0001
Screen(08_30-03_42)-0002
Screen(08_30-03_43)-0003
Screen(08_30-03_44)-0005

PR #747, opened 2026-04-05
fix(inventory): prevent phantoms, wipes and dupes of inventory items

olisikh:
This PR hardens item movement and rollback behavior to prevent inventory wipes, dupes, and desyncs during failed operations and edge-case states.

**What changed**

- Added move-state validation so item moves are only processed in valid gameplay states (EnteredWorld, NpcDialogOpened, TradeOpened).
- Strengthened rollback for failed moves:
  -   first restore to original slot,
  -   then fallback to any free slot,
  -   log critical context if both fail,
  -   trigger inventory refresh to resync client/server state.
-   Fixed complete-stack handling so the source item is actually removed from its storage after stacking.
-   Fixed slot-blocking logic that used the wrong index while scanning occupied slots.
-   Corrected IItemMovingPlugIn call to report the actual target storage.
-   Hardened trade/disconnect backup restore paths with per-item exception handling, fallback placement, and detailed logging so one bad item doesn’t abort full restore.

**Effect**

- Lower risk of full inventory wipe on restore/cancel flows.
- Lower risk of duplication from stack/move edge cases.
- Better consistency between in-memory inventory, persisted state, and client view.
- Better diagnostics when recovery paths are hit.

I have been running these changes on my local server and didn't see an issue, but this PR should be taken with a grain of salt.


src/GameLogic/PlayerActions/Items/MoveItemAction.cs
```diff
@@ -324,15 +346,7 @@ private Movement ItemFitsAtNewLocation(Player player, Item item, byte toSlot, by
...
-        {
-            if (this.IsTargetSlotBlocked(item, toSlot, toStorage, usedSlots))
-            {
-                return true;
-            }
-        }
-
-        return false;
+        return this.IsTargetSlotBlocked(item, toSlot, toStorage, usedSlots);
     }
```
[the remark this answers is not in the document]
olisikh: good point, cleaned up.

src/GameLogic/PlayerActions/Items/MoveItemAction.cs
```diff
@@ -92,9 +99,10 @@ public async ValueTask MoveItemAsync(Player player, byte fromSlot, Storages from
         }
     }
 
-    private async ValueTask FullStackAsync(Player player, Item sourceItem, Item targetItem)
+    private async ValueTask FullStackAsync(Player player, IStorage sourceStorage, Item sourceItem, Item targetItem)
     {
         targetItem.Durability += sourceItem.Durability;
+        await sourceStorage.RemoveItemAsync(sourceItem).ConfigureAwait(false);
         await player.InvokeViewPlugInAsync<IItemMoveFailedPlugIn>(p => p.ItemMoveFailedAsync(sourceItem)).ConfigureAwait(false);
```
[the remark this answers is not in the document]
olisikh: addressing this would take a lot of effort which is beyond the scope of this PR

src/GameLogic/PlayerActions/Items/MoveItemAction.cs
```diff
@@ -379,5 +388,18 @@ private void SetUsedSlots(StorageInfo toStorage, Item blockingItem, bool[,] used
         }
     }
 
+    private bool IsMoveAllowed(Player player)
```
sven-n: We could even go further an check the source/target `Storages`

src/GameLogic/PlayerActions/Items/MoveItemAction.cs
```diff
@@ -379,5 +388,18 @@ private void SetUsedSlots(StorageInfo toStorage, Item blockingItem, bool[,] used
...
+    {
+        if (player.SelectedCharacter is null)
+        {
+            return false;
+        }
+
+        return player.PlayerState.CurrentState == PlayerState.EnteredWorld
+               || player.PlayerState.CurrentState == PlayerState.NpcDialogOpened
+               || player.PlayerState.CurrentState == PlayerState.TradeOpened
+               || player.PlayerState.CurrentState == PlayerState.TradeButtonPressed;
```
sven-n: Is `TradeButtonPressed` really a valid state under which item movements should happen?

sven-n (changes requested):

src/GameLogic/PlayerActions/Items/MoveItemAction.cs
olisikh: Yeah, it doesn't make sense to allow that if server already has TradeButtonPressed state. I realised I can't reliably test it, it makes sense to not allow item movements when state is `TradeButtonPressed`

olisikh: now it should:
- allow moves in TradeOpened state
- reject moves in TradeButtonPressed state

src/GameLogic/PlayerActions/Items/MoveItemAction.cs
olisikh: now it should test player state and both source/target storage states before allowing item moves.

sven-n:
Thank you :)

PR #748, opened 2026-04-05
fix(inventory): Daybreak weapon dimensions (2 x 4 squares, instead of 2 x 2)

olisikh:
looting Daybreak in the last 4 squares of the inventory would leave them as phantom for the client, because client expects a 2 x 4 (8 squares) to be occupied, not 2 x 2. This issue would persist through relogins until item is moved / deleted through Admin UI.


sven-n (approved):

sven-n:
Usually we do data changes through configuration update plugins, but for this simple fix we can live with a migration.

Thanks :)

olisikh:
@sven-n thanks for guidance, that's really good to know. I would ensure to do it as you suggested from now on.

I was also thinking that general configuration page is bloated with very "deep" configuration options, that you normally don't wanna see all the time. Those would sit well in dedicated plugins. 👍

PR #749, opened 2026-04-08
Fix Decay and Ice Storm area skill settings

itayalroy:
Fix AreaSkillSettings of a few mage skills:

1. Fix decay's target area so it no longer hits monsters far outside the intended impact zone
2. Remove Ice Storm's deferred-hit timing
3. Fix hellfire & hellfire strengthener target areas

### DIFF
--- a/src/Persistence/Initialization/Updates/FixAreaSkillsUpdatePlugIn.cs
+++ b/src/Persistence/Initialization/Updates/FixAreaSkillsUpdatePlugIn.cs
@@ -1,0 +1,101 @@
+// <copyright file="FixAreaSkillsUpdatePlugIn.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.Persistence.Initialization.Updates;
+
+using System.Runtime.InteropServices;
+using MUnique.OpenMU.DataModel.Configuration;
+using MUnique.OpenMU.Persistence.Initialization.Skills;
+using MUnique.OpenMU.PlugIns;
+
+/// <summary>
+/// This update fixes area skills' range, effect radius, and delay issues.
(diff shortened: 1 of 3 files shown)


sven-n:
Thanks, it looks fine. I'll need to test it when I have time

itayalroy:
@sven-n  I've added a fix to hellfire (target area was too big) and hellfire strengthener (target area was 0, skill did not hit any monster)

itayalroy:
@sven-n Actually, I've been comparing skill effect area implementation in OpenMU vs MuAlfa open source (https://github.com/creadormu/GameServer), and they use euclidean distance to check which monsters are in range, while OpenMU is using manhattan distance (which creates a square effect area)- is it intentional?

In any case, it seems that on legacy MU implementations, the client chooses which monsters are hit, and the server only validates it against some sanity value. I think it's worth to thoroughly align OpenMU skill effect areas with the legacy MU client, because it is very likely that we have additional skills with similar issues

sven-n:
Okay, I'll think about reverting it. Which checks do you do on server side? IMHO, it's way to easy to cheat with the original system.

itayalroy:
@sven-n I followed most GameServer implementations out there, they define a valid radius per skill (which is an approximation, not exact effect area), then verify that each monster the client reports as hit is within that radius from the player.

Although this might allow for some level of manipulation, as long as we enforce player position on server side (which OpenMU does) I can't really think of a game breaking cheat to implement here, correct me if you think otherwise

ze-dom:
From the sources I use ([zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/), [emu](https://github.com/kyleruss/emu-server/)), sometimes it's Manhattan, sometimes Euclidean - it depends on the skill.
OpenMU can currently accommodate both:
For Manhattan use `effectRange`
For Euclidean use `useTargetAreaFilter` and `targetAreaDiameter`

sven-n (approved):
Looks fine, thank you :)

PR #760, opened 2026-04-26
Add max drop level item property and remove hardcoded check from DefaultDropGenerator

eduardosmaniotto:
- Created new column for item MaximumDropLevel
- Created new project for GameLogic benchmark


sven-n:
Looks good, thank you

PR #766, opened 2026-05-02
DK Master Tree Finishup

ze-dom:
**To-do**

- [x] Comment changes with source references
- [x] UpdatePlugin

**Developments**
- Moved _Cold_ magic effect to `SkillsInitializerBase.ApplyElementalModifier()` as it's shared with Chain Drive and Strike of Destruction
- Added _Swell Life_ magic effect maximum values
- Added _Swell Life Proficiency_ magic effect
- Added _Twisting Slash Mastery_, _Rageful Blow Mastery_, _Spear Mastery_, _Mace Mastery_, _Swell Life Strengthener_, and _Swell Life Proficiency_ skill effects
- Replaced `Stats.PollutionMoveTargetChance` and `Stats.StunChance` for the more generic and descriptive `Stats.MasteryMoveTargetChance` and `Stats.MasteryStunChance`, which can serve several classes

**Bugfixes**

- Map names in logs
- Changed Chain Drive magic effect number from _Iced_ to _Cold_
- Fixed _Life Swell_ magic effect boost calcs
- Fixed MagicEffect comparison in `SkillsInitizalizerBase.CreateEffect()` (bug introduced in previous PR)
- Fixed double wield damage calcs (averaged the weapons DMG + item options and excellent options)

PR #770, opened 2026-05-09
Fix player position desync during movement

itayalroy:
The current Walker implementation is not aligned to client-side movement- it uses wrong movement speeds and does not account for many parameters that client takes into account when calculating player/monster movement speed (safezones, boots, wings, pets, mounts, diagonal tiles, skill effects such as ice/strike of destruction etc). As a result, server and client positions of players/monsters frequently go out of sync, which causes the server to reject valid walk requests, and send forced move corrections back, causing a flaky movement experience overall.

This PR refines server-side player/monster movement by calculating walk delays per step accounting for movement speed modifiers (safe zones, wings, pets, boots, gloves, Fenrir, ice, strike of destruction, etc) instead of using one fixed inaccurate delay.

It also includes 2 other movement related minor fixes:
- Blocks player attacks when either side is protected by a safe zone.
- Stops sending walk direction data by default, since MuMain no longer uses it.

### DIFF
--- a/src/GameLogic/Attributes/Stats.cs
+++ b/src/GameLogic/Attributes/Stats.cs
@@ -440,6 +440,21 @@ public class Stats
     public static AttributeDefinition WalkSpeed { get; } = new(new Guid("9CDDC598-E5F3-4372-9294-505455E4A40B"), "Walk Speed", string.Empty);
 
     /// <summary>
+    /// Gets the maximum movement speed attribute definition.
+    /// </summary>
+    public static AttributeDefinition MaxMovementSpeed { get; } = new(new Guid("E29301BE-626B-4B42-9F68-0DFAC18B3856"), "Maximum Movement Speed", "The maximum movement speed of a character on regular terrain.");
+
+    /// <summary>
+    /// Gets the maximum underwater movement speed attribute definition.
+    /// </summary>
+    public static AttributeDefinition MaxMovementSpeedUnderwater { get; } = new(new Guid("12128DC7-0740-48A5-A653-E546191CD7E0"), "Maximum Underwater Movement Speed", "The maximum movement speed of a character on underwater terrain.");
+
+    /// <summary>
+    /// Gets the movement speed factor attribute definition.
+    /// </summary>
+    public static AttributeDefinition MovementSpeedFactor { get; } = new(new Guid("003E1F2E-661D-4258-BEF0-33111D5F4AD2"), "Movement Speed Factor", "The factor which is applied to the final movement speed of a character.");
+
+    /// <summary>
     /// Gets the attack damage increase attribute definition.
     /// </summary>
     /// <remarks>Includes wings, imp, dinorant, black fenrir, and infinite arrow strengthener.</remarks>
@@ -1330,6 +1345,11 @@ public class Stats
     /// Gets the <see cref="IsInSafezone"/> attribute which defines if the character is located in a safezone of a game map.
     /// </summary>
     public static AttributeDefinition IsInSafezone { get; } = new(new Guid("82044DF9-F528-4AD6-9AAA-6FEAA4C786E7"), "Flag, if the character is located in a safezone of a game map", "Characters at the safezone recover additional health and shield.");
+
+    /// <summary>
+    /// Gets the <see cref="IsUnderwater"/> attribute which defines if the character is located on an underwater game map.
+    /// </summary>
+    public static AttributeDefinition IsUnderwater { get; } = new(new Guid("72A684C1-102B-4FDE-B637-2665ADD5F4AE"), "Flag, if the character is located on an underwater game map", "Characters on underwater maps use underwater movement speed attributes.");
 
     /// <summary>
     /// Gets the attribute definition, which defines if a player has MU Helper activated.
--- a/src/GameLogic/Attributes/Stats.cs
+++ b/src/GameLogic/Attributes/Stats.cs
@@ -440,14 +440,14 @@ public class Stats
     public static AttributeDefinition WalkSpeed { get; } = new(new Guid("9CDDC598-E5F3-4372-9294-505455E4A40B"), "Walk Speed", string.Empty);
 
     /// <summary>
-    /// Gets the maximum movement speed attribute definition.
+    /// Gets the movement speed attribute definition.
     /// </summary>
-    public static AttributeDefinition MaxMovementSpeed { get; } = new(new Guid("E29301BE-626B-4B42-9F68-0DFAC18B3856"), "Maximum Movement Speed", "The maximum movement speed of a character on regular terrain.");
+    public static AttributeDefinition MovementSpeed { get; } = new(new Guid("E29301BE-626B-4B42-9F68-0DFAC18B3856"), "Movement Speed", "The movement speed of a character on regular terrain.");
 
     /// <summary>
-    /// Gets the maximum underwater movement speed attribute definition.
+    /// Gets the underwater movement speed attribute definition.
     /// </summary>
-    public static AttributeDefinition MaxMovementSpeedUnderwater { get; } = new(new Guid("12128DC7-0740-48A5-A653-E546191CD7E0"), "Maximum Underwater Movement Speed", "The maximum movement speed of a character on underwater terrain.");
+    public static AttributeDefinition MovementSpeedUnderwater { get; } = new(new Guid("12128DC7-0740-48A5-A653-E546191CD7E0"), "Underwater Movement Speed", "The movement speed of a character on underwater terrain.");
 
     /// <summary>
     /// Gets the movement speed factor attribute definition.
--- a/src/GameLogic/NPC/Monster.cs
+++ b/src/GameLogic/NPC/Monster.cs
@@ -18,6 +18,11 @@ using Nito.AsyncEx;
 /// </summary>
 public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISupportWalk, IMovable, ISummonable
 {
+    private const short IcedEffectNumber = 0x38;
+    private const short BlowOfDestructionEffectNumber = 0x56;
+    private const double IcedMovementSpeedFactor = 0.5;
+    private const double BlowOfDestructionMovementSpeedFactor = 0.33;
+
     private readonly AsyncLock _moveLock = new();
     private readonly INpcIntelligence _intelligence;
     private readonly Walker _walker;
@@ -59,7 +64,7 @@ public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISuppor
         : base(spawnInfo, stats, map, eventStateProvider, dropGenerator, plugInManager)
     {
         this._pathFinderPool = pathFinderPool;
-        this._walker = new Walker(this, () => this.StepDelay);
+        this._walker = new Walker(this, this.GetStepDelay);
         this._intelligence = npcIntelligence;
 
         (this._skillPowerUp, this._skillPowerUpDuration, this._skillPowerUpTarget) = this.CreateMagicEffectPowerUp();
@@ -87,7 +92,7 @@ public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISuppor
     public Point WalkTarget => this._walker.CurrentTarget;
 
     /// <inheritdoc/>
-    public TimeSpan StepDelay => this.Definition.MoveDelay;
+    public TimeSpan StepDelay => this.GetStepDelay(null);
 
     /// <inheritdoc/>
     /// <remarks>Monsters don't do combos.</remarks>
@@ -354,6 +359,30 @@ public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISuppor
             From = node.PreviousPoint,
             To = new Point(node.X, node.Y),
         };
+    }
+
+    private TimeSpan GetStepDelay(WalkingStep? step)
+    {
+        var tileDistance = step is { } walkingStep ? walkingStep.From.EuclideanDistanceTo(walkingStep.To) : 1.0;
+        var delayMilliseconds = this.Definition.MoveDelay.TotalMilliseconds * Math.Max(1.0, tileDistance);
+        delayMilliseconds /= this.GetMovementSpeedFactor();
+
+        return TimeSpan.FromMilliseconds(delayMilliseconds);
+    }
+
+    private double GetMovementSpeedFactor()
+    {
+        if (this.MagicEffectList.ActiveEffects.ContainsKey(IcedEffectNumber))
+        {
+            return IcedMovementSpeedFactor;
+        }
+
+        if (this.MagicEffectList.ActiveEffects.ContainsKey(BlowOfDestructionEffectNumber))
+        {
+            return BlowOfDestructionMovementSpeedFactor;
+        }
+
+        return 1.0;
     }
 
     /// <summary>
--- a/src/GameLogic/NPC/Monster.cs
+++ b/src/GameLogic/NPC/Monster.cs
@@ -18,30 +18,20 @@ using Nito.AsyncEx;
 /// </summary>
 public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISupportWalk, IMovable, ISummonable
 {
-    private const short IcedEffectNumber = 0x38;
-    private const short BlowOfDestructionEffectNumber = 0x56;
-    private const double IcedMovementSpeedFactor = 0.5;
-    private const double BlowOfDestructionMovementSpeedFactor = 0.33;
-
     private readonly AsyncLock _moveLock = new();
     private readonly INpcIntelligence _intelligence;
     private readonly Walker _walker;
 
     /// <summary>
-    /// The power up element of the monster skill.
-    /// It is a "cached" element which will be created on demand and can be applied multiple times.
+    /// The power up elements of the monster skill.
+    /// These are "cached" elements which will be created on demand and can be applied multiple times.
     /// </summary>
-    private readonly IElement? _skillPowerUp;
+    private readonly (AttributeDefinition Target, IElement Boost)[] _skillPowerUps;
 
     /// <summary>
-    /// The duration of the <see cref="_skillPowerUp"/>.
+    /// The duration of the <see cref="_skillPowerUps"/>.
     /// </summary>
     private readonly IElement? _skillPowerUpDuration;
-
-    /// <summary>
-    /// The target attribute of the <see cref="_skillPowerUp"/>.
-    /// </summary>
-    private readonly AttributeDefinition? _skillPowerUpTarget;
 
     private readonly IObjectPool<PathFinder> _pathFinderPool;
 
@@ -67,7 +57,7 @@ public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISuppor
         this._walker = new Walker(this, this.GetStepDelay);
         this._intelligence = npcIntelligence;
 
-        (this._skillPowerUp, this._skillPowerUpDuration, this._skillPowerUpTarget) = this.CreateMagicEffectPowerUp();
+        (this._skillPowerUps, this._skillPowerUpDuration) = this.CreateMagicEffectPowerUps();
 
         this._intelligence.Npc = this;
     }
@@ -127,7 +117,7 @@ public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISuppor
         await this.ForEachWorldObserverAsync<IShowAnimationPlugIn>(p => p.ShowMonsterAttackAnimationAsync(this, target, this.GetDirectionTo(target)), true).ConfigureAwait(false);
         if (this.Definition.AttackSkill is { } attackSkill)
         {
-            await target.TryApplyElementalEffectsAsync(this, attackSkill, this._skillPowerUp, this._skillPowerUpDuration, this._skillPowerUpTarget, hitInfo).ConfigureAwait(false);
+            await target.TryApplyElementalEffectsAsync(this, attackSkill, this._skillPowerUps, this._skillPowerUpDuration, hitInfo).ConfigureAwait(false);
 
             await this.ForEachWorldObserverAsync<IShowSkillAnimationPlugIn>(p => p.ShowSkillAnimationAsync(this, target, attackSkill, true), true).ConfigureAwait(false);
         }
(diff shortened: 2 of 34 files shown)


src/GameLogic/NPC/Monster.cs
```diff
@@ -18,6 +18,11 @@ namespace MUnique.OpenMU.GameLogic.NPC;
 /// </summary>
 public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISupportWalk, IMovable, ISummonable
 {
+    private const short IcedEffectNumber = 0x38;
+    private const short BlowOfDestructionEffectNumber = 0x56;
```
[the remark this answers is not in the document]
itayalroy: MagicEffectNumber enum lives in Persistence.Initialization, inaccessible to GameLogic. I leave moving effect numbers to shared code to a follow-up PR.

src/GameLogic/Player.cs
```diff
@@ -41,6 +42,9 @@ namespace MUnique.OpenMU.GameLogic;
 /// </summary>
 public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacker, ITrader, IPartyMember, IRotatable, IHasBucketInformation, ISupportWalk, IMovable, ILoggerOwner<Player>
 {
+    private const short IcedEffectNumber = 0x38;
+    private const short BlowOfDestructionEffectNumber = 0x56;
```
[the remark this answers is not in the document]
itayalroy: MagicEffectNumber enum lives in Persistence.Initialization, inaccessible to GameLogic. I leave moving effect numbers to shared code to a follow-up PR.

src/GameLogic/NPC/Monster.cs
```diff
@@ -18,6 +18,11 @@ namespace MUnique.OpenMU.GameLogic.NPC;
...
+    private const double DelayedMovementAdditionalMilliseconds = 300.0;
+    private const double DiagonalStepDelayFactor = 1.3;
```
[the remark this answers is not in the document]
itayalroy: Fixed (just for context, this 1.3 was taken from Pegasus server source, but indeed inaccurate, better use sqrt(2))

src/GameServer/RemoteView/World/ObjectMovedPlugInExtended.cs
```diff
@@ -40,7 +40,7 @@ protected override async ValueTask SendWalkAsync(IConnection connection, ushort
     {
         int Write()
         {
-            var stepsSize = steps.Length == 0 ? 1 : (steps.Length / 2) + 2;
+            var stepsSize = stepsLength == 0 ? 0 : (steps.Length / 2) + 2;
```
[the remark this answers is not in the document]
itayalroy: Fixed

src/GameServer/RemoteView/World/ObjectMovedPlugIn.cs
```diff
@@ -95,7 +95,7 @@ protected virtual async ValueTask SendWalkAsync(IConnection connection, ushort o
     {
         int Write()
         {
-            var stepsSize = steps.Length == 0 ? 1 : (steps.Length / 2) + 2;
+            var stepsSize = stepsLength == 0 ? 0 : (steps.Length / 2) + 2;
```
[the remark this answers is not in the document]
itayalroy: Fixed

src/GameLogic/NPC/Monster.cs
```diff
@@ -356,6 +361,30 @@ private static WalkingStep GetStep(PathResultNode node)
...
+    {
+        if (this.MagicEffectList.ActiveEffects.ContainsKey(IcedEffectNumber))
+        {
+            return IcedMovementSpeedFactor;
+        }
+
+        if (this.MagicEffectList.ActiveEffects.ContainsKey(BlowOfDestructionEffectNumber))
+        {
+            return BlowOfDestructionMovementSpeedFactor;
+        }
```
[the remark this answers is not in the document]
itayalroy: MuMain checks Iced first actually :)

src/GameLogic/Player.cs
```diff
@@ -2107,19 +2127,119 @@ private async ValueTask RegenerateHeroStateAsync()
...
+    {
+        if (this.MagicEffectList.ActiveEffects.ContainsKey(IcedEffectNumber))
+        {
+            return speed * 0.5;
+        }
+
+        if (this.MagicEffectList.ActiveEffects.ContainsKey(BlowOfDestructionEffectNumber))
+        {
+            return speed * 0.33;
+        }
```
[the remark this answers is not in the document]
itayalroy: MuMain checks Iced first actually :)

src/GameLogic/Player.cs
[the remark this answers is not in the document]
itayalroy: Done

ze-dom:
Hey @sven-n and @itayalroy!
I also added the strike of destruction magic effect (cold) in my [PR](https://github.com/MUnique/OpenMU/pull/766/changes#diff-e624c0fb034008b6843ec5f5ac27ee93022e6267a0e0ddee06ed2c084a49f10c). (It is also used for the RF's Chain Drive skill). I don't know what will be merged first, but this can create a conflict.

My suggestion is that my PR gets merged first. That way @itayalroy can drop the changes to the files in /Persistence. It will also save him the hassle of creating the update plugin :stuck_out_tongue:

src/GameLogic/NPC/Monster.cs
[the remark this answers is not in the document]
sven-n: You can check the iced-status by `this.AttributeSystem[Stats.IsIced]`.

src/GameLogic/NPC/Monster.cs
```diff
@@ -18,6 +18,11 @@ namespace MUnique.OpenMU.GameLogic.NPC;
 /// </summary>
 public sealed class Monster : AttackableNpcBase, IAttackable, IAttacker, ISupportWalk, IMovable, ISummonable
 {
+    private const short IcedEffectNumber = 0x38;
+    private const short BlowOfDestructionEffectNumber = 0x56;
+    private const double IcedMovementSpeedFactor = 0.5;
+    private const double BlowOfDestructionMovementSpeedFactor = 0.33;
```
sven-n: Please add a new `Stats.MovementSpeedFactor` (default value: 1.0), where the ice effect multiplies the value with 0.5, and blow of descruction with 0.33.

src/GameLogic/Player.cs
```diff
@@ -41,6 +42,17 @@ namespace MUnique.OpenMU.GameLogic;
 /// </summary>
 public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacker, ITrader, IPartyMember, IRotatable, IHasBucketInformation, ISupportWalk, IMovable, ILoggerOwner<Player>
 {
+    private const short IcedEffectNumber = 0x38;
+    private const short BlowOfDestructionEffectNumber = 0x56;
+    private const double IcedMovementSpeedFactor = 0.5;
+    private const double BlowOfDestructionMovementSpeedFactor = 0.33;
```
sven-n: Same as for the monster class.

src/GameLogic/Player.cs
```diff
@@ -2107,19 +2135,135 @@ private async ValueTask RegenerateHeroStateAsync()
...
+            || this.IsItem(pet, 13, 2)
+            || this.IsItem(pet, 13, 3))
+        {
+            return this.GetWingMovementSpeed(wings, runSpeed, fastWingSpeed);
         }
 
-        // TODO: Consider pets etc.
-        return TimeSpan.FromMilliseconds(500);
+        return this.HasRunningGear() ? runSpeed : walkSpeed;
+    }
```
sven-n: Imho, there are too many hardcoded values and special logic. The item definitions could hold something like a movement speed, e.g. in new `Stats.MaxMovementSpeed`/`Stats.MaxMovementSpeedUnderwater` attributes. Then the players attribute system can be asked about it.

And about atlans/kalima: You could add another new `Stats.IsUnderwater` which can be added to the `GameMapDefinition.CharacterPowerUpDefinitions` with value 1. So when the character enters the map, it gets this attribute value in it's attribute system, too.

sven-n (changes requested):

src/GameLogic/NPC/Monster.cs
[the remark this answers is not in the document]
itayalroy: done

src/GameLogic/NPC/Monster.cs
itayalroy: done

src/GameLogic/Player.cs
itayalroy: done

src/GameLogic/Player.cs
itayalroy: Agreed, I added MaxMovementSpeed/MaxMovementSpeedUnderwater attributes and movement speed power ups to the relevant items. I also added an isUnderwater attribute which is updated by underwater map powerups.

itayalroy:
@sven-n Fixed according to your comments. Since we now modify the DB structure, this resulted in a fairly large DB update script. I tested this PR with both a fresh DB and an existing DB updated via the script, and both work properly. All relevant changes are in 401f123955d68dd08bbbed5b74dc9808f2abbe53.

itayalroy:
@sven-n pinging about this one :)

sven-n:
Checking it soon 😅

src/GameLogic/Walker.cs
```diff
@@ -180,15 +181,18 @@ public void Dispose()
...
         var lastOffset = TimeSpan.Zero;
         while (!cancellationToken.IsCancellationRequested)
         {
             var sw = Stopwatch.StartNew();
-            await this.WalkStepAsync(cancellationToken).ConfigureAwait(false);
+            var step = await this.WalkStepAsync(cancellationToken).ConfigureAwait(false);
+            if (step is null)
+            {
+                continue;
+            }
```
sven-n, line 198: Potential hot loop: when `step` is `null`, this `continue`s without any delay. Most null paths cancel the token or call `StopAsync()` first, so they terminate — but the generic `catch (Exception)` in `WalkStepAsync` returns `null` with only a `Debug.Fail` (a no-op in release builds). If that exception repeats, this loop spins at 100% CPU with no backoff.

Consider a small fallback delay (or breaking out of the loop) when `step is null` and the token isn't cancelled.

---

src/Persistence/Initialization/Skills/ColdEffectInitializer.cs
```diff
@@ -0,0 +1,47 @@
...
+        magicEffect.StopByDeath = true;
+        magicEffect.Duration = this.Context.CreateNew<PowerUpDefinitionValue>();
+        magicEffect.Duration.ConstantValue.Value = (float)TimeSpan.FromSeconds(10).TotalSeconds;
+
+        var movementSpeedFactorPowerUp = this.Context.CreateNew<PowerUpDefinition>();
+        magicEffect.PowerUpDefinitions.Add(movementSpeedFactorPowerUp);
+        movementSpeedFactorPowerUp.TargetAttribute = Stats.MovementSpeedFactor.GetPersistent(this.GameConfiguration);
+        movementSpeedFactorPowerUp.Boost = this.Context.CreateNew<PowerUpDefinitionValue>();
+        movementSpeedFactorPowerUp.Boost.ConstantValue.Value = MovementSpeedConstants.BlowOfDestructionMovementSpeedFactor;
+        movementSpeedFactorPowerUp.Boost.ConstantValue.AggregateType = AggregateType.Multiplicate;
```
sven-n, line 45: Question on effect stacking: in an earlier thread you mentioned MuMain checks Iced *first*, i.e. only one slow factor applies at a time. With the attribute model, Iced (×0.5) and Blow of Destruction (×0.33) are both `Multiplicate` elements on `MovementSpeedFactor`, so when both effects are active the server computes 1 × 0.5 × 0.33 = 0.165 while the client would use 0.5 — the same class of desync this PR fixes.

Edge case, and it follows the requested attribute design — but does the client stack these or short-circuit?

---

src/GameLogic/Player.cs
```diff
@@ -2149,19 +2168,48 @@ private async ValueTask RegenerateHeroStateAsync()
...
         }
 
-        // TODO: Consider pets etc.
-        return TimeSpan.FromMilliseconds(500);
+        var speedAttribute = this.Attributes?[Stats.IsUnderwater] > 0
+            ? Stats.MaxMovementSpeedUnderwater
+            : Stats.MaxMovementSpeed;
+        var speed = this.Attributes?[speedAttribute] ?? 0;
+
+        return this.ApplyMovementSpeedFactor(Math.Max(WalkMovementSpeed, speed));
```
sven-n, line 2188: A player with no +5 boots/wings/mount gets `Max(12, 0)` = walk speed (12) outside safezones. That matches the old-school "need +5 boots to run" behavior — but does it also match the Season 6 client, where IIRC running doesn't require +5 boots? If S6 clients always run at 15, itemless players would desync.

---

src/GameServer/RemoteView/World/ObjectMovedPlugIn.cs
```diff
@@ -40,7 +40,7 @@ public class ObjectMovedPlugIn : IObjectMovedPlugIn
     /// Gets or sets a value indicating whether the directions provided by <see cref="ISupportWalk.GetDirectionsAsync"/> should be send when an object moved.
     /// This is usually not required, because the game client calculates a proper path anyway and doesn't use the suggested path.
     /// </summary>
-    public bool SendWalkDirections { get; set; } = true;
+    public bool SendWalkDirections { get; set; }
```
sven-n, line 43: This default change affects all configured clients globally (including the 0.75/0.97 walk-code paths), not just MuMain. If any supported older client actually consumes the step data, observed movement of other objects breaks for them. It's plugin-configurable, so this may be a deliberate trade-off — but it deserves a mention in the changelog/release notes.

---

src/Persistence/Initialization/Version095d/Items/Pets.cs
```diff
@@ -71,6 +78,11 @@ private ItemDefinition CreatePet(byte number, SkillNumber skillNumber, string na
                 var powerUpDefinition = this.Context.CreateNew<ItemBasePowerUpDefinition>();
                 powerUpDefinition.TargetAttribute = basePowerUp.Item1.GetPersistent(this.GameConfiguration);
                 powerUpDefinition.BaseValue = basePowerUp.Item2;
+                if (basePowerUp.Item1 == Stats.MaxMovementSpeed || basePowerUp.Item1 == Stats.MaxMovementSpeedUnderwater)
```
sven-n: `AggregateType` is honored only for the two movement-speed attributes here, while the other tuples (e.g. Dinorant's `DamageReceiveDecrement` ×0.9, `AttackDamageIncrease` ×1.15) still silently ignore their declared `Multiplicate` type — which looks like a pre-existing bug this conditional tiptoes around.

Either set `AggregateType` unconditionally (and verify 0.95d balance) or add a comment explaining why it's restricted; as written, the condition will confuse the next reader.

---

src/GameLogic/Player.cs
```diff
@@ -1450,6 +1458,17 @@ public async ValueTask WalkToAsync(Point target, Memory<WalkingStep> steps)
...
+    private bool IsAttackBlockedBySafezone(IAttacker attacker)
+    {
+        if (this.IsAtSafezone())
+        {
+            return true;
+        }
+
+        var attackerPlayer = attacker as Player ?? (attacker as IPlayerSurrogate)?.Owner;
+        return attackerPlayer?.IsAtSafezone() is true;
+    }
```
sven-n: Nice defense-in-depth, and the `IPlayerSurrogate` handling for summons is a good touch. One note: this only guards attacks *on players* (`Player.AttackByAsync`) — attacks on monsters from a safezone remain governed only by the action-layer checks (`HitAction`, `AreaSkillHitAction`, etc.). Intentional asymmetry?

---

sven-n (commented):
Solid, well-iterated PR — the attribute-based design is clean, the migration plugins are careful and idempotent, and along the way it fixes three real pre-existing bugs which I verified against the codebase:

- **Map power-up double-application** (`Player.RaisePlayerEnteredMap`): `PowerUpWrapper` already self-registers in its constructor, so the removed `AddElement` call was applying every map power-up twice, and the old cleanup removed only one of them.
- **`MonsterAttributeHolder.RemoveElement`**: the old condition was inverted — it removed the composed attribute while extra elements still existed and kept it when only the base remained.
- **Walk packet sizing**: the old code used the rented buffer length (16), so walk packets always carried 10 bytes of step data; also confirmed the tighter `steps.Span[..stepsLength]` slice can't go out of range in `SetStepData`.

Also verified that always calling `UpdateSetPowerUps()` in `InventoryStorage` is necessary (the Fenrir speed bonuses are `ItemOptionCombinationBonus`es, and Fenrir has no `PossibleItemSetGroups`, so the old guard skipped recalculation), and that fresh installs pick up the new attributes via the `typeof(Stats)` reflection in `GameConfigurationInitializerBase`.

I left six inline comments. The `Walker` spin guard is the only code change I'd consider blocking; the Iced+BoD stacking and default run speed ones are client-fidelity questions where you two know the domain best.

Minor, non-blocking:
- The Fenrir combination-bonus numbers 101–106 and construction logic are duplicated between `VersionSeasonSix/Items/Pets.cs` and `AddMovementSpeedAttributesPlugInBase` — the numbers could live in `MovementSpeedConstants` next to the speeds.
- The constants in `Player.GetStepDelay` (`terrainScale = 100`, `referenceFrameTimeMilliseconds = 40`, `WalkMovementSpeed = 12`) would benefit from a short comment stating the client formula they mirror — that provenance is the justification for the math.
- No tests were added; `Player.GetStepDelay` (safezone/underwater/factor/diagonal cases) and the `MonsterAttributeHolder.RemoveElement` fix are very testable, and a regression test for the latter would have caught the original inverted condition.

---

src/Persistence/Initialization/Skills/ColdEffectInitializer.cs
itayalroy: Right, the client ignores Blow of Destruction when Iced is present. There is currently no good way to replicate this behavior using the attribute system, as it requires two multiplicative attributes with priority rules between them. For now I left this edge case unresolved

src/GameLogic/Player.cs
itayalroy: At least with MuMain, players do need +5 boots to run (just verified)

src/GameServer/RemoteView/World/ObjectMovedPlugIn.cs
itayalroy: Isn't the step data an addition of OpenMU+MuMain? I don't think original MU clients ever used these.

src/GameLogic/Walker.cs
itayalroy: Fixed by stopping the walker when running into a null step

src/Persistence/Initialization/Version095d/Items/Pets.cs
itayalroy: Changed the 0.95d pet initializer to apply aggregate type for every power-up, matching 0.75/s6 initializers

src/GameLogic/Player.cs
itayalroy: Unintentional, I moved the safezone guard into a shared IAttackable helper and now use it for all targets

src/Persistence/Initialization/Skills/ColdEffectInitializer.cs
ze-dom: Maybe this works (didn't test):
```
movementSpeedFactorPowerUp.TargetAttribute = Stats.MovementSpeedFactor.GetPersistent(this.GameConfiguration);
movementSpeedFactorPowerUp.Boost = this.Context.CreateNew<PowerUpDefinitionValue>();
movementSpeedFactorPowerUp.Boost.ConstantValue.Value = 0;
movementSpeedFactorPowerUp.Boost.ConstantValue.AggregateType = AggregateType.Multiplicate;

var maxSpeedFactor = this.Context.CreateNew<AttributeRelationship>();
maxSpeedFactor.InputAttribute = Stats.IsIced.GetPersistent(this.GameConfiguration);
maxSpeedFactor.InputOperator = InputOperator.Maximum;
maxSpeedFactor.InputOperand = MovementSpeedConstants.BlowOfDestructionMovementSpeedFactor;
movementSpeedFactorPowerUp.Boost.RelatedValues.Add(maxSpeedFactor);
```
Also, could you change the "BlowOfDestruction" names to "Cold" in this PR? That's the name S6E3 client and server sources use. RF's _Chain Drive_ skill also uses this effect:
https://github.com/MUnique/OpenMU/pull/766/changes#r3176790807

itayalroy: Renamed the effect to cold. reg. your suggestion, I'm not sure about the 0 const value. some variation of this might work, no time to test it right now, perhaps we can do a follow up PR for this specific issue, I think it's probably very uncommon, and also, sounds like the client is bugged for ignoring cold when ice is active

src/Persistence/Initialization/VersionSeasonSix/Items/Wings.cs
```diff
@@ -79,7 +79,7 @@ public override void Initialize()
         var secondWingOptions = this.CreateSecondClassWingOptions();
         this.CreateWing(3, 5, 3, "Wings of Spirits", 150, 30, 200, 215, 0, 0, 2, 0, 0, 0, 0, this.BuildOptions((0b10, OptionType.HealthRecover), (0b00, OptionType.PhysDamage)), 32, 25, this._damageIncreaseByLevelTableSecond, secondWingOptions);
         this.CreateWing(4, 5, 3, "Wings of Soul", 150, 30, 200, 215, 2, 0, 0, 0, 0, 0, 0, this.BuildOptions((0b00, OptionType.HealthRecover), (0b10, OptionType.WizDamage)), 32, 25, this._damageIncreaseByLevelTableSecond, secondWingOptions);
-        this.CreateWing(5, 3, 3, "Wings of Dragon", 150, 45, 200, 215, 0, 2, 0, 0, 0, 0, 0, this.BuildOptions((0b00, OptionType.HealthRecover), (0b10, OptionType.PhysDamage)), 32, 25, this._damageIncreaseByLevelTableSecond, secondWingOptions);
+        this.CreateWing(5, 3, 3, "Wings of Dragon", 150, 45, 200, 215, 0, 2, 0, 0, 0, 0, 0, this.BuildOptions((0b00, OptionType.HealthRecover), (0b10, OptionType.PhysDamage)), 32, 25, this._damageIncreaseByLevelTableSecond, secondWingOptions, movementSpeed: MovementSpeedConstants.FastWingMovementSpeed);
```
ze-dom, line 82: This is very interesting. I didn't know BK wings were faster 😃

src/GameLogic/Attributes/Stats.cs
```diff
@@ -439,6 +439,21 @@ public class Stats
...
 
+    /// <summary>
+    /// Gets the maximum movement speed attribute definition.
+    /// </summary>
+    public static AttributeDefinition MaxMovementSpeed { get; } = new(new Guid("E29301BE-626B-4B42-9F68-0DFAC18B3856"), "Maximum Movement Speed", "The maximum movement speed of a character on regular terrain.");
+
+    /// <summary>
+    /// Gets the maximum underwater movement speed attribute definition.
+    /// </summary>
+    public static AttributeDefinition MaxMovementSpeedUnderwater { get; } = new(new Guid("12128DC7-0740-48A5-A653-E546191CD7E0"), "Maximum Underwater Movement Speed", "The maximum movement speed of a character on underwater terrain.");
```
ze-dom: Do you think `MovementSpeed` and `MovementSpeedUnderwater`are more appropriate names for char attributes/stats (which vary depending on items equipped)?

`MaxMovementSpeed` and `MaxMovementSpeedUnderwater` sounds more like hard constant names to me. (I understand you might have named like that due to these being comprised of `AggregateType.Maximum` values.)

src/GameLogic/Player.cs
ze-dom: Evidence from S6E3 client:
https://github.com/user-attachments/assets/559809fe-7d1f-43ca-92a9-00a9ba56b4d1

src/GameLogic/Attributes/Stats.cs
itayalroy: I don't have a strong preference but Max indeed seems redundant so renamed it

itayalroy:
@sven-n could you take another look on this one?

PR #772, opened 2026-05-12
Monster Attribute Scaler Plugin

nolt:
Monster Attribute Scaler Plugin
-------------------------------

What it does:
A feature plugin that scales monster base stats by a configurable percentage.
Each stat group (Damage, AttackRate, DefenseRate, Defense, Health) has its own
percentage setting. Default: +25% to MinimumPhysBaseDmg, MaximumPhysBaseDmg,
AttackRatePvm, DefenseRatePvm, DefenseBase, and MaximumHealth. Supports a
"Scale All" option that sets all stat groups at once. Starts disabled by default.

How it works:
The plugin uses shared multiplicative IElement instances, one per stat group,
added to all monsters at spawn time via IObjectAddedToMapPlugIn and removed
on despawn via IObjectRemovedFromMapPlugIn. Configuration changes in the
Admin Panel take effect immediately by mutating the shared element values,
which propagate to all active monsters through the attribute system's
ValueChanged event. The configuration class handles ScaleAll cascading and
individual field isolation at the property level.

Bug fix — ConfigurationChangeHandler (not plugin-specific, but required for live updates):
The OnPlugInConfigurationChanged method passed the entity primary key (Id)
instead of plugInConfiguration.TypeId to all PlugInManager calls. This caused
ActivatePlugIn, DeactivatePlugIn, and ConfigurePlugIn to silently no-op because
PlugInManager looks up plugins by their [Guid] TypeId, not by the database entity
PK. Fix: replaced 'id' with 'plugInConfiguration.TypeId' in all four calls.

Files modified for this plugin:
OpenMU/src/GameLogic/Properties/PlugInResources.resx
OpenMU/src/GameLogic/Properties/PlugInResources.Designer.cs

Files modified (bug fix, required for live updates):
OpenMU/src/Startup/ConfigurationChangeHandler.cs

Files added for this plugin:
OpenMU/src/GameLogic/PlugIns/MonsterAttributeScaler.cs
OpenMU/src/GameLogic/PlugIns/MonsterAttributeScalerConfiguration.cs

### DIFF
--- a/src/GameLogic/PlugIns/MonsterAttributeScaler.cs
+++ b/src/GameLogic/PlugIns/MonsterAttributeScaler.cs
@@ -1,0 +1,136 @@
+// <copyright file="MonsterAttributeScaler.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic.PlugIns;
+
+using System.Collections.Concurrent;
+using System.ComponentModel.DataAnnotations;
+using System.Runtime.InteropServices;
+using MUnique.OpenMU.AttributeSystem;
+using MUnique.OpenMU.GameLogic;
+using MUnique.OpenMU.GameLogic.Attributes;
+using MUnique.OpenMU.GameLogic.NPC;
+using MUnique.OpenMU.PlugIns;
+
+/// <summary>
+/// A feature plugin which scales all monster base stats by a configurable percentage.
+/// Uses multiplicative IElement instances applied per-monster. A server restart is
+/// required for configuration changes to take effect.
+/// </summary>
+[PlugIn]
+[Display(Name = "Monster Attribute Scaler", Description = "Increases all monster base stats by a configurable percentage.")]
+[Guid("59E5B45F-0A3D-4BAF-8B6C-9E1D2F3A4B5C")]
+public class MonsterAttributeScaler : IFeaturePlugIn, IObjectAddedToMapPlugIn, IObjectRemovedFromMapPlugIn,
+    ISupportCustomConfiguration<MonsterAttributeScalerConfiguration>,
+    ISupportDefaultCustomConfiguration,
+    IDisabledByDefault
+{
+    private MonsterAttributeScalerConfiguration? _configuration;
+
+    private readonly ConcurrentDictionary<IAttackable, List<(AttributeDefinition Attribute, IElement Element)>> _scaledMonsters = new();
+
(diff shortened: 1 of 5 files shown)


sven-n (changes requested):
> Configuration changes in the Admin Panel require a server restart to take effect.

Do you know why? Because the plugin creates one element per monster and never updates it again.
How about this solution:
* The plugin just holds one SimpleElement and shares it between all monsters by adding the same instance to all monsters. No tracking of monsters required.
* When configuration changed, simply apply the change to the one shared SimpleElement instance.

Can you try that?

It could also be interesting to add percentages for different kind of attributes. At the moment, it scales every attribute by the same percentage.

nolt:
Good idea I will try this your solution. I tried to make a plugin as simple as it possible and try to not change anything in source code only plugin files.
And yea I can add few options. Again the goal was to make it simple and scale all attributes at once.

nolt:
Plugin now allows setting % to each stat + option change to all. Also I've implemented your soulution.
NOTE: Admin Panel changing settings works and I tested this but in game tests are not done yet.

nolt:
Long tests ingame with debug and logging. But finally plugin works as it should - tested in game. (1st PR message updated).
Bug has been found in ConfigurationChangeHandler.cs
 
Here is an detailed explanation from AI.

---

Bug found and fixed: Live configuration updates from Admin Panel silently did nothing.

Root cause: In `ConfigurationChangeHandler.OnPlugInConfigurationChanged` (src/Startup/ConfigurationChangeHandler.cs), the `id` parameter (entity primary key `PlugInConfiguration.Id`, e.g. `331d9e01-...`) was passed to all `PlugInManager` methods — `IsPlugInActive`, `ActivatePlugIn`, `DeactivatePlugIn`, and `ConfigurePlugIn`. These methods expect a plugin `TypeId` (the `[Guid]` attribute on the plugin class, e.g. `59e5b45f-...`). The mismatch caused all four calls to silently no-op (no matching plugin found), so configuration saves to the database persisted correctly but never reached the runtime plugin instance.

Fix: Replaced every usage of the `id` parameter with `plugInConfiguration.TypeId` in the four `PlugInManager` calls at lines 114–126. After the fix, the full pipeline works end-to-end: Admin Panel save → EF change tracker → ConfigurationChangeHandler → PlugInManager.ConfigurePlugIn → PlugInConfigurationChanged event → plugin Configuration setter → SimpleElement.Value update → live stat change on all monsters.

sven-n (approved):

sven-n:
Thank you :)

PR #775, opened 2026-05-15
Reset info in character details added

nolt:
## Summary
This PR adds reset info in character details.

mumain_resets

## Related issues
https://github.com/sven-n/MuMain/issues/361

### DIFF
--- a/src/GameServer/RemoteView/Character/UpdateCharacterStatsPlugIn.cs
+++ b/src/GameServer/RemoteView/Character/UpdateCharacterStatsPlugIn.cs
@@ -65,8 +65,7 @@ public class UpdateCharacterStatsPlugIn : IUpdateCharacterStatsPlugIn
             (ushort)this._player.Attributes[Stats.BaseLeadership],
             (ushort)this._player.SelectedCharacter.UsedNegFruitPoints,
             this._player.SelectedCharacter.GetMaximumFruitPoints(),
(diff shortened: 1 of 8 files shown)


itayalroy:
I think we should probably make it optional in case OpenMU is configured with resets disabled

src/GameServer/RemoteView/Character/UpdateCharacterStatsPlugIn.cs
sven-n, line 1: The default plugin should not be changed.

sven-n (changes requested):
It looks like you modified the packet code manually. Please adjust the CharacterStatsExtended packet in the ServerToClientPackets.xml file. The code is generated based on this file.
And I think using `ushort` is enough for the number of resets, too.
See: https://github.com/MUnique/OpenMU/blob/f6b1943a33e932db792420b77fb85ac39c02e738/src/Network/Packets/ServerToClient/ServerToClientPackets.xml#L6320-L6483

nolt:
I will made necessary change in ServerToClientPackets.xml + revert to ```ushort```.

src/GameServer/RemoteView/Character/UpdateCharacterStatsPlugIn.cs
nolt: done

PR #785, opened 2026-05-27
Apply monster attribute changes instantly on the game server

### LINKED ISSUE #784: Apply changes to monster attributes instantly on the game server
**Is your feature request related to a problem? Please describe.**
When we change values in the attributes of monsters, they don't have immediate effect. You'd need to restart the server for that.

**Describe the solution you'd like**
Immediate effect of the changes on the game server.

**Additional context**
We already have code infrastructure for that, see `MUnique.OpenMU.GameLogic.ConfigurationChangeMediator`.
I think we somehow need to register about changes of `MUnique.OpenMU.DataModel.Configuration.MonsterAttribute` and update the `MUnique.OpenMU.GameLogic.Attributes.MonsterAttributeHolder` accordingly.

sven-n:
## Summary
Closes #784.

Changing a monster's attributes (e.g. via the admin panel) previously had no effect until a server restart, because `MonsterAttributeHolder` snapshots the monster's stats from its `MonsterDefinition` at construction time into a static cache and never refreshed them.

This change hooks into the existing `ConfigurationChangeMediator`: whenever a `MonsterAttribute` is added/removed/modified, the infrastructure publishes a parent-level change for its `MonsterDefinition`. Each spawned attackable NPC now registers for those changes and reloads its cached stat attributes in place — no re-spawn, so health and position are preserved.

## Changes
- **`MonsterAttributeHolder`**: added `ApplyChanges()`, which rebuilds the stat dictionary from the (already updated in place) `MonsterDefinition.Attributes`, refreshes the shared static cache, and atomically swaps in the new dictionary.
- **`AttackableNpcBase`**: added `ReloadAttributes()` forwarding to the holder.
- **`MapInitializer.RegisterForConfigChanges`**: each spawned attackable NPC registers for changes to its `MonsterDefinition`; the registration is disposed alongside the NPC in all existing teardown paths.

## Known limitation
Stats that currently have a per-instance composable element applied (e.g. an active magic effect, or the disabled-by-default `MonsterAttributeScaler`) won't reflect a base-value change until that element is removed/re-added, since composable attributes snapshot their base at add time. This matches existing system behavior.

## Test plan
- [x] New unit tests in `MonsterAttributeReloadTests` cover changing an existing attribute value and adding a new attribute; both confirm the value takes effect after `ReloadAttributes()`.
- [x] Manual verification: edit a monster's attribute in the admin panel on a running server and confirm immediate effect.

### DIFF
--- a/src/GameLogic/Attributes/MonsterAttributeHolder.cs
+++ b/src/GameLogic/Attributes/MonsterAttributeHolder.cs
@@ -35,9 +35,9 @@ public class MonsterAttributeHolder : IAttributeSystem
 
     private readonly AttackableNpcBase _monster;
 
-    private readonly IDictionary<AttributeDefinition, float> _statAttributes;
-
     private readonly object _attributesLock = new();
+
+    private IDictionary<AttributeDefinition, float> _statAttributes;
 
     /// <summary>
     /// Attribute dictionary of a monster instance.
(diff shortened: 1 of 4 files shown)


sven-n:
Thanks for the review! Addressed all three findings in 030bd554d:

- **Memory leak (high):** The `MonsterDefinition` change registration is now owned by the NPC. `AttackableNpcBase` gained a `RegisterDisposable(IDisposable)` method and disposes its registrations in `Dispose(bool)` (reliably invoked from both `Dispose()` and `DisposeAsync()`), so it's cleaned up even when the NPC is disposed outside the mediator callbacks (dying non-respawning/wandering monster, map cleanup, shutdown). The manual disposal calls in the spawn-area callbacks were removed since the NPC's disposal handles it (the mediator's disposable is idempotent).
- **Redundant rebuilds (medium):** `ApplyChanges()` now checks the shared cache first; the first instance of a monster type rebuilds the stat dictionary and the rest just adopt it.
- Added a test verifying all spawned instances of the same definition pick up a change (covers the optimization path). All tests pass.

ze-dom:
Tested it and it looks good :+1:

sven-n:
> Tested it and it looks good 👍

Thank you :)

PR #787, opened 2026-05-29
Dark mode for Admin Panel

nolt:
Dark mode for Admin Panel

---

## Summary

Adds a dark-mode toggle to the admin panel with persistence across sessions and page reloads. Implementation mirrors the existing `CultureSelector` pattern (cookie-based, no SignalR state), so the architectural footprint is minimal and the mental model is consistent with what already exists in the codebase.

The toggle lives in the panel header next to the "About" link. State is stored in an `OpenMU.Theme` cookie. A `[data-theme]` attribute on `<html>` drives all styling via CSS custom properties.

## What's in

**New files**
- `src/Web/Shared/Services/ThemeController.cs` — MVC controller endpoint `Theme/Set?theme=…&redirectUri=…` (1:1 with `CultureController`)
- `src/Web/Shared/Components/ThemeSelector.razor` + `.razor.cs` + `.razor.css` — toggle button with moon/sun icons from open-iconic
- `src/Web/Shared/wwwroot/themeSelector.js` — tiny `export function current()` reading `document.documentElement.dataset.theme` for hydration after the Blazor interactive circuit takes over
- `src/Web/Shared/wwwroot/css/theme.css` — theme tokens (`:root`/`[data-theme="dark"]`) and Bootstrap 4 surface overrides (`.bg-light`, tables, forms, dropdowns, modals, sidebar dropdown submenu)

**Modified files**
- `src/Web/AdminPanel/Components/App.razor` — sets `<html data-theme="…">` from the cookie via `[CascadingParameter] HttpContext`
- `src/Web/AdminPanel/Components/Layout/MainLayout.razor` + `.razor.css` — adds the `<ThemeSelector>` to the header, recolors sidebar gradient/breadcrumb/`#blazor-error-ui` via CSS variables
- `src/Web/AdminPanel/Components/Layout/CreationPanel.razor.css`, `ConfigurationSearch.razor.css` — surface colors switched to CSS variables
- `src/Web/Shared/Components/Form/AutoForm.razor.css` — Save/Refresh action bar background switched to CSS variables (fixes light panel under buttons on every config edit page)
- `src/Web/Shared/Exports.cs` — `theme.css` added to `Stylesheets`
- `src/Web/Shared/Properties/Resources.resx` + `Resources.Designer.cs` — three new strings: `ToggleTheme`, `SwitchToDarkMode`, `SwitchToLightMode`
- `src/Web/AdminPanel/_Imports.razor` — `@using Microsoft.AspNetCore.Http` so the cascading `HttpContext` resolves

## Design decisions

**Cookie + reload (not localStorage + JS swap).** Identical pattern to `CultureSelector`. Keeps SSR honest — the first paint matches the active theme, no flash-of-light-content. Cost: one full reload per toggle, accepted in admin tooling (low traffic, infrequent action).

**CSS variables in a separate `theme.css`, not in the SCSS pipeline.** The Dockerfile runs `dotnet build -p:ci=true`, which skips the `CompileSass` MSBuild target. SCSS edits never reach the published `shared.css` in containerized builds. `theme.css` is a plain static asset loaded via `Exports.Stylesheets` and works on every host.

**Bootstrap 4 surface overrides (not an upgrade to BS5).** Bootstrap 5 ships native dark-mode via `data-bs-theme`, but the migration is a separate, much larger effort. Overriding the specific BS4 classes the panel uses (`.bg-light`, `.table-striped`, `.dropdown-menu`, `.form-control`, `.modal-content`, …) is the smallest viable change.

**JS hydration via `wwwroot/themeSelector.js` as an ES module, not colocated `.razor.js`.** Colocated `*.razor.js` files are 404 in this project's build pipeline (verified — `ReconnectModal.razor.js` is also 404; see related issues). `wwwroot/` is the reliable static-asset path, same as `shared.css` and our new `theme.css`. The hydration reads `data-theme` from `<html>` after the interactive circuit connects, because `[CascadingParameter] HttpContext` is null in interactive renders — without this, the toggle would get stuck in dark mode (the icon would flip back to "moon" while the cookie stayed `dark`).

**Icon convention: action-affordance.** Moon icon shown in light mode (click → go dark), sun icon shown in dark mode (click → go light). Same convention as iOS, GitHub, Stripe Dashboard, Material Design. Tooltip ("Switch to dark mode" / "Switch to light mode") reinforces the affordance.

## Known limitations / out-of-scope

- **Brief icon flicker** on the first SSR→interactive transition (~50–200 ms) when reload happens while already in dark mode. The SSR renders the correct icon, the first interactive re-render briefly shows the wrong one (cascading HttpContext is null in interactive), then JS hydration corrects it. Fixable with `PersistentComponentState`; not worth the boilerplate for an admin tool.
- **Per-culture translations** of the three new resource strings — only English defaults added. Project doesn't have per-culture `.resx` files yet, so consistent with existing convention.
- **`MUnique.OpenMU.Web.AdminPanel.csproj` SDK reference chain** — not touched. Pre-existing static-web-asset path-doubling issue (`_content/AdminPanel/_content/<pkg>/…` 404s) and the `ReconnectModal.razor.js` 404 are pre-existing bugs unrelated to dark mode, surfaced in the same DevTools sweep. Tracked separately (see related issues).

## Test plan

- [x] Build the all-in-one Docker container, hit `localhost:8082`, confirm panel loads in light mode by default
- [x] Click the moon icon in the header → page reloads → background turns dark, icon becomes sun
- [x] Click the sun icon → page reloads → back to light mode, icon becomes moon
- [x] Navigate to **Game Configuration → General** — `AutoForm` panel and the sticky Save/Refresh bar are dark
- [x] Navigate to **Game Configuration → Monsters** — Edit button (rendered as `<a class="btn btn-info">`) has white text on teal background, readable
- [x] Navigate to **Game Configuration → Merchant Stores** — Edit button (rendered as `<button class="btn-info">`) is consistent
- [x] Open the Game Configuration dropdown in the sidebar — submenu has dark surface, links contrast cleanly
- [x] Open any modal (e.g. ItemEdit) — dark modal background, dark form fields
- [x] Open a table-based view (Accounts, Servers) — striped rows and hover state visible in dark mode
- [x] Verify cookie `OpenMU.Theme = dark` is set in DevTools → Application → Cookies after toggling
- [x] Verify `<html data-theme="dark">` is set in DevTools → Elements after toggling
- [x] Reload the page in dark mode → state persists, icon shows sun on first paint (no flicker on cold load)

Preview:

### DIFF
--- a/src/Web/AdminPanel/Components/App.razor
+++ b/src/Web/AdminPanel/Components/App.razor
@@ -1,7 +1,16 @@
 @using MUnique.OpenMU.Web.AdminPanel.Components.Layout
+@using MUnique.OpenMU.Web.Shared.Services
+
+@code {
+    [CascadingParameter]
+    public HttpContext? HttpContext { get; set; }
+
+    private string CurrentTheme =>
+        ThemeController.NormalizeTheme(this.HttpContext?.Request.Cookies[ThemeController.CookieName]);
+}
 
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" data-theme="@this.CurrentTheme">
 <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
--- a/src/Web/AdminPanel/Components/Layout/ConfigurationSearch.razor.css
+++ b/src/Web/AdminPanel/Components/Layout/ConfigurationSearch.razor.css
@@ -36,7 +36,7 @@
 
 .configuration-search__result-path {
     font-size: 0.78rem;
-    color: #5f6368;
+    color: var(--omu-text-muted);
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
@@ -45,7 +45,7 @@
 
 .configuration-search__result:hover .configuration-search__result-path,
 .configuration-search__result:focus .configuration-search__result-path {
-    color: rgba(0, 0, 0, 0.7);
+    color: var(--omu-text);
 }
 
 @media (max-width: 640.98px) {
--- a/src/Web/AdminPanel/Components/Layout/CreationPanel.razor.css
+++ b/src/Web/AdminPanel/Components/Layout/CreationPanel.razor.css
@@ -6,8 +6,9 @@
     flex-shrink: 0;
     display: flex;
     flex-direction: row;
-    background-color: #fff;
-    border-left: 1px solid #d6d5d5;
+    background-color: var(--omu-surface-2);
+    color: var(--omu-text);
+    border-left: 1px solid var(--omu-border);
     box-shadow: -2px 0 5px rgba(0, 0, 0, 0.08);
     overflow: hidden;
     z-index: 2;
@@ -22,14 +23,14 @@
     width: 1.75rem;
     padding: 0;
     border: none;
-    border-right: 1px solid #d6d5d5;
-    background-color: #f7f7f7;
+    border-right: 1px solid var(--omu-border);
+    background-color: var(--omu-surface-muted);
     cursor: pointer;
-    color: #333;
+    color: var(--omu-text);
 }
 
 .collapse-toggle:hover {
-    background-color: #e9e9e9;
+    background-color: var(--omu-surface-hover);
 }
 
 .creation-panel-body {
--- a/src/Web/AdminPanel/Components/Layout/MainLayout.razor
+++ b/src/Web/AdminPanel/Components/Layout/MainLayout.razor
@@ -1,6 +1,17 @@
 @using MUnique.OpenMU.Web.AdminPanel.Properties
+@using MUnique.OpenMU.Web.Shared.Services
 
 @inherits LayoutComponentBase
+
+@code {
+    [CascadingParameter]
+    public HttpContext? HttpContext { get; set; }
+
+    private bool IsDarkTheme => string.Equals(
+        this.HttpContext?.Request.Cookies[ThemeController.CookieName],
+        "dark",
+        StringComparison.OrdinalIgnoreCase);
+}
 
 <div class="page">
     <div class="sidebar">
@@ -14,14 +25,15 @@
                     <div class="header-search flex-grow-1">
                         <ConfigurationSearch />
                     </div>
-                    <div class="header-actions ml-3">
-                        <a href="https://munique.net" target="_blank" class="text-secondary small">@Resources.About</a>
+                    <div class="header-actions ml-3 d-flex align-items-center">
+                        <ThemeSelector IsDark="@this.IsDarkTheme" />
+                        <a href="https://munique.net" target="_blank" class="text-secondary small ml-3">@Resources.About</a>
                     </div>
                 </div>
             </div>
         </header>
 
-        <div class="breadcrumb-bar bg-white border-bottom px-4 py-3">
+        <div class="breadcrumb-bar border-bottom px-4 py-3">
             <div class="d-flex align-items-center">
                 <BreadcrumbNavigation />
             </div>
--- a/src/Web/AdminPanel/Components/Layout/MainLayout.razor.css
+++ b/src/Web/AdminPanel/Components/Layout/MainLayout.razor.css
@@ -9,7 +9,7 @@ main {
 }
 
 .sidebar {
-    background-image: linear-gradient(180deg, rgb(5, 39, 103) 0%, #3a0647 70%);
+    background-image: var(--omu-sidebar-gradient);
 }
 
 .header-search {
@@ -24,6 +24,7 @@ main {
 
 .breadcrumb-bar {
     min-height: 2.5rem;
+    background-color: var(--omu-surface-2);
 }
 
 .breadcrumb-bar ::deep .breadcrumb-nav {
@@ -58,8 +59,8 @@ main {
 }
 
 #blazor-error-ui {
-    color-scheme: light only;
-    background: lightyellow;
+    background: var(--omu-error-bg);
+    color: var(--omu-error-text);
     bottom: 0;
     box-shadow: 0 -1px 2px rgba(0, 0, 0, 0.2);
     box-sizing: border-box;
--- a/src/Web/AdminPanel/_Imports.razor
+++ b/src/Web/AdminPanel/_Imports.razor
@@ -5,6 +5,7 @@
 @using Microsoft.AspNetCore.Components.Forms
 @using Microsoft.AspNetCore.Components.Routing
 @using Microsoft.AspNetCore.Components.Web
+@using Microsoft.AspNetCore.Http
 @using static Microsoft.AspNetCore.Components.Web.RenderMode
 @using Microsoft.AspNetCore.Components.Web.Virtualization
 @using Microsoft.JSInterop
--- a/src/Web/Shared/Components/Form/AutoForm.razor.css
+++ b/src/Web/Shared/Components/Form/AutoForm.razor.css
@@ -1,6 +1,7 @@
 .auto-form {
-    background: #fff;
-    border: 1px solid #dee2e6;
+    background: var(--omu-surface-2);
+    color: var(--omu-text);
+    border: 1px solid var(--omu-border);
     border-radius: 0.375rem;
     padding: 1.5rem;
     margin-bottom: 1rem;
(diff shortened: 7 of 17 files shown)


nolt:
**1. Constructor injection on ThemeSelector**
Switched to an `[Inject]` property. Worth noting that constructor
injection does in fact work on Blazor components since .NET 5+ —
`CultureSelector.razor.cs` in this repo uses the exact same pattern
and ships in production — but the `[Inject]` property style is by far
the majority pattern in the codebase (MapEditor, ConfigurationSearch,
NavMenu, etc.), so consistency wins. Renamed to `NavigationManager`
(PascalCase) per .NET naming conventions for properties.

**2. Relative `Theme/Set?…` URI**
Added the leading slash. With `<base href="/" />` and
`NavigationManager.NavigateTo(uri, forceLoad: true)` the relative
form does resolve against the app root regardless of current route
(again confirmed by CultureSelector working in production with the
same pattern), but explicit-is-better-than-implicit here, the
absolute form removes a whole class of edge cases for zero cost.

**3. `redirectUri` validation in `ThemeController`**
Good catch — applied exactly as suggested: nullable parameter,
`IsNullOrEmpty || !Url.IsLocalUrl` guard, fallback to
`LocalRedirect("/")`. `LocalRedirect` does throw
`InvalidOperationException` on missing/external URIs, which would
have been a 500 on a malformed request.

**Heads-up on follow-ups**

Comments #2 and #3 also apply verbatim to the existing
`CultureSelector.razor.cs` / `CultureController.cs` — I cloned the
Theme code from them so the same hardenings would be useful there.
Happy to send a separate small PR for that if you'd like, didn't
want to expand the scope of this one.

**Other changes in this push (not from review)**

While testing the final build I noticed three things worth fixing:

- Unchecked checkboxes and radios in dark mode rendered bright
  white. Added `color-scheme: dark` to `[data-theme="dark"]` so
  the browser renders native form controls (and scrollbars) in
  their dark variant.
- Search fields (LookupField, MultiLookupField, Typeahead)
  ignored the theme because `Typeahead.razor.css` hardcoded
  Bootstrap defaults. Converted to the `--omu-*` tokens — fixes
  Item Slot / Skill / Consume Effect on `ItemDefinition` edit and
  every other auto-form lookup in the panel.
- The sidebar's purple/blue gradient cut off at 100vh on FHD with
  Game Configuration expanded — menu items past the cap rendered
  on plain page background, sidebar's light-gray-on-dark text
  effectively invisible. Tracked it down to a layout regression
  introduced by `a3d91226e` (your "Add persistent side panel"
  commit) — full analysis in the commit message body. TL;DR a
  `.sidebar { height: 100vh; sticky }` rule has been sitting
  dormant in `MainLayout.razor.css` since the Blazor Web App
  migration (`f8c8be5ce`, Jan 2), because the AdminPanel CSS
  isolation bundle was never linked from `App.razor`. Your commit
  correctly added that link so the new CreationPanel styles could
  ship, which incidentally activated the dormant rule. Sticky is
  already provided by `.sidebar > nav` in `shared.css`, so
  removing the duplicate restores the pre-regression layout (which
  is what users were seeing from Jan through May 24).

Independent of the dark-mode work — this would have shipped as a
bug whether dark-mode landed or not. Folded it in here so I'm not
opening a second PR for what amounts to a three-line diff, but
happy to split it out if you'd prefer it as a separate change.

nolt:
Thanks for the careful review. Rebased onto current master (which
picked up #788) and pushed a follow-up addressing all three comments
plus a couple of dark-mode coverage gaps I hit while testing.

**1. Constructor injection on ThemeSelector**
Switched to an `[Inject]` property. Worth noting that constructor
injection does in fact work on Blazor components since .NET 5+ —
`CultureSelector.razor.cs` in this repo uses the exact same pattern
and ships in production — but the `[Inject]` property style is by far
the majority pattern in the codebase (MapEditor, ConfigurationSearch,
NavMenu, etc.), so consistency wins. Renamed to `NavigationManager`
(PascalCase) per .NET naming conventions for properties.

**2. Relative `Theme/Set?…` URI**
Added the leading slash. With `<base href="/" />` and
`NavigationManager.NavigateTo(uri, forceLoad: true)` the relative
form does resolve against the app root regardless of current route
(again confirmed by CultureSelector working in production with the
same pattern), but explicit-is-better-than-implicit here, the
absolute form removes a whole class of edge cases for zero cost.

**3. `redirectUri` validation in `ThemeController`**
Good catch — applied exactly as suggested: nullable parameter,
`IsNullOrEmpty || !Url.IsLocalUrl` guard, fallback to
`LocalRedirect("/")`. `LocalRedirect` does throw
`InvalidOperationException` on missing/external URIs, which would
have been a 500 on a malformed request.

**Heads-up on follow-ups**

Comments #2 and #3 also apply verbatim to the existing
`CultureSelector.razor.cs` / `CultureController.cs` — I cloned the
Theme code from them so the same hardenings would be useful there.
Happy to send a separate small PR for that if you'd like, didn't
want to expand the scope of this one.

**Other changes in this push (not from review)**

While testing the final build I noticed three things worth fixing:

- Unchecked checkboxes and radios in dark mode rendered bright
  white. Added `color-scheme: dark` to `[data-theme="dark"]` so
  the browser renders native form controls (and scrollbars) in
  their dark variant.
- Search fields (LookupField, MultiLookupField, Typeahead)
  ignored the theme because `Typeahead.razor.css` hardcoded
  Bootstrap defaults. Converted to the `--omu-*` tokens — fixes
  Item Slot / Skill / Consume Effect on `ItemDefinition` edit and
  every other auto-form lookup in the panel.
- After rebasing onto #788 (admin-layout-fixes), the new sticky
  `.form-actions` footer in `CreationPanel.razor.css` hardcodes
  `background-color: #fff` and `border-top: 1px solid #dee2e6`,
  which stay bright on a dark surface. Added a `[data-theme="dark"]
  .creation-panel-body .form-actions` override in `theme.css` so
  the footer matches the panel in dark mode. (Could alternatively
  be done by switching those two values to `--omu-*` tokens in
  `CreationPanel.razor.css` itself — happy to send that as a
  separate cleanup if you'd prefer the source-of-truth fix.)

nolt:
Pushed one more commit with dark-mode polish I caught while visually
walking the panel. All in `theme.css`, no other
files touched, no history rewrite.

Coverage gaps fixed:

- **QuickGrid**: the `<button class="col-title">` header text inherited
  Bootstrap's global `button { color: #212529 }` so the "Name" column
  header was nearly invisible on dark. Sort indicator and paginator
  arrows (▲▼ ≪ ‹ › ≫) are SVG data-URIs with no `fill` attribute, default
  to black — used `filter: invert(1)` since we can't reach into the
  data-URI to recolor.
- **`.add-new-bar`** (sticky "Add New" footer on `EditConfigGrid` pages
  like Monsters/Items/Skills) is the second sticky-footer #788 added
  with hardcoded `#fff` + `#dee2e6`; same override pattern as
  `.form-actions` from the first commit.
- **Blazored.Modal panel** (`.blazored-modal`) hardcodes
  `background-color: #fff` in its NuGet bundle. The Plugins ⚙ modal and
  any multi-select modal use it.
- **`.multi-lookup-field__*`** had a full set of hardcoded Bootstrap
  defaults — affects the Plugins modal contents and the Excellent /
  Wing option pickers in `ItemEdit`.
- **Bootstrap alert variants** (`.alert-primary` / `-success` / `-info`
  / `-warning` / `-danger` / `-light`) used on `Updates.razor` (and
  `alert-danger` also on `Install.razor`) were unreadable on dark.
  Tinted-bg + bright-accent-text pattern, plus a neutral `<hr>` override
  for the per-variant bright hr lines.

Verified each rule lands in the served `theme.css` via curl and walked
each page (Monsters / Items / Plugins / Updates / Install) by hand in
the browser.

Note: this also indirectly covers a `MultiLookupField` cleanup that
could alternatively live in the source file itself (replacing #fff /
#ced4da with `--omu-*` tokens). Happy to send that as a separate small
"convert hardcoded colors to theme tokens" PR if preferred — kept the
fix here in `theme.css` to avoid spreading changes across more files.


Verification reference (what was checked, for reviewers)
========================================================

Built locally via the project's docker compose, hit the running admin
panel at http://localhost:8082, toggled dark mode, and walked:

  Monsters / Skills / Items / Character classes  — Name header, sort
    arrows, paginator arrows, Add New bar
  Plugins ⚙ Activate                              — Blazored modal,
    MultiLookupField inside
  ItemEdit (any item with excellent/wing options) — MultiLookupField
  Updates (/config-updates)                       — alert-success
    (currently shown), alert-info / -warning / -primary / -danger /
    -light would render correctly with their tinted backgrounds

nolt:
Ready to merge.

sven-n:
Thanks, I didn't look into detail, but I trust you 😅

PR #790, opened 2026-05-31
Feature: item search

eduardosmaniotto:
Adds a filterable search input to ItemTable, which renders every collection
property (e.g. PossibleItems, Skills, BasePowerUpAttributes). The search
filters items by name, matching the same pattern already used in AutoForm.

[admin.webm](https://github.com/user-attachments/assets/ad37d4ae-41bc-471b-91fc-79e5db49995d)


sven-n:
That's nice, thank you :)

PR #792, opened 2026-06-01
Feature: Add auto accept party request from friends and guild members

eduardosmaniotto:
## MU Helper: new flags, party auto-accept, combat update

**New MU Helper flags** — `UseSelfDefense`, `AutoAcceptFriend`, `AutoAcceptGuild`, `FallbackBasicAttack` added to the 257-byte client blob deserialization, `IMuHelperSettings` interface, and model class.

**Party auto-accept** — `PartyRequestHandler.TryAutoAcceptPartyRequestAsync` automatically accepts party requests when the receiver has `AutoAcceptFriend` (checked via `IFriendServer.IsFriendAsync`) or `AutoAcceptGuild` (same guild ID) enabled.

**CombatHandler** — removed buff-based early return. `FallbackBasicAttack` flag controls whether basic attack is performed when no skill is configured.

**Friend server** — added `IsFriendAsync` to `IFriendServer` (in-memory + Dapr). `DeleteFriendAsync` now cleans up both directions (subscribers + persistence).

**Immediate effect** — MU Helper config saves (0xAE packet) re-deserialize `player.MuHelperSettings` so flag changes apply without relog.

**Tests** — 6 party auto-accept tests, CombatHandlerTests updated.

### DIFF
--- a/src/Dapr/FriendServer.Host/FriendServerController.cs
+++ b/src/Dapr/FriendServer.Host/FriendServerController.cs
@@ -78,6 +78,17 @@ public class FriendServerController : ControllerBase
     }
 
     /// <summary>
+    /// Determines whether two players are friends.
+    /// </summary>
+    /// <param name="data">The data.</param>
+    /// <returns>True if the two players are friends; otherwise false.</returns>
+    [HttpPost(nameof(IFriendServer.IsFriendAsync))]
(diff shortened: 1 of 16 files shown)


sven-n:
Nice features, I like them! Is the client getting updated as well?

eduardosmaniotto:
The flags already exist on the client but are currently not being used.

PR #797, opened 2026-06-09
Global message to selected/all servers

nolt:
This allows sending global message (yellow one) to all or specyfic server.

scren1

### DIFF
--- a/src/Web/AdminPanel/Pages/Servers.razor
+++ b/src/Web/AdminPanel/Pages/Servers.razor
@@ -16,6 +16,32 @@
 }
 else
 {
+<div class="card mb-3">
+    <div class="card-body">
+        <div class="row g-2 align-items-end">
+            <div class="col-md-5">
+                <label for="global-message" class="form-label mb-1">@Resources.GlobalMessage</label>
+                <input id="global-message" type="text" class="form-control" placeholder="@Resources.MessagePlaceholder" @bind="_message" @bind:event="oninput" />
+            </div>
+            <div class="col-md-4">
+                <label for="server-select" class="form-label mb-1">@Resources.Target</label>
+                <select id="server-select" class="form-select" @bind="_selectedServerId">
+                    <option value="-1">@Resources.AllGameServers</option>
+                    @foreach (var server in _servers.Where(s => s.Type == ServerType.GameServer && s.ServerState == ServerState.Started))
+                    {
+                        <option value="@server.Id">@server.Description (@(string.Format(Resources.OnlineCount, server.CurrentConnections)))</option>
+                    }
+                </select>
+            </div>
+            <div class="col-md-3">
+                <button class="btn btn-primary w-100" @onclick="SendMessageAsync" disabled="@string.IsNullOrWhiteSpace(_message)">
+                    @Resources.Send
+                </button>
+            </div>
+        </div>
+    </div>
+</div>
+
 <div>
   <table>
     <thead>
@@ -78,6 +104,10 @@ else
 
     private bool _isRestarting;
 
+    private string _message = string.Empty;
+
+    private int _selectedServerId = -1;
+
     /// <summary>
     /// Gets or sets the <see cref="IServerProvider"/>.
     /// </summary>
@@ -89,6 +119,12 @@ else
     /// </summary>
     [Inject]
     public IGameServerInstanceManager? ServerInstanceManager { get; set; }
+
+    /// <summary>
+    /// Gets or sets the game servers.
+    /// </summary>
+    [Inject]
+    public IDictionary<int, IGameServer> GameServers { get; set; } = null!;
 
     /// <inheritdoc />
     public void Dispose()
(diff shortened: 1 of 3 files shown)


sven-n:
## Review

Nice feature — a global-message broadcaster on the Servers page is genuinely useful, and the UI/UX touches (disabling Send on empty input, clearing the field on the sync context to prevent double-submit) are well thought out. The dropdown→send id mapping is also correct: `IDictionary<int, IGameServer>` is keyed by `gameServer.Id` (`GameServerContainer.cs:176`), which is the same `Id` exposed by `IManageableServer`, so the option values line up with the lookup. `MessageType.GoldenCenter` is the right enum for the yellow center message.

I do have one blocking concern and a few minor notes.

### 🔴 Blocking: breaks the distributed (Dapr) AdminPanel host

The page adds a **required** injection:

```csharp
[Inject]
public IDictionary<int, IGameServer> GameServers { get; set; } = null!;
```

That dictionary is only registered in the single-process host (`Startup/Program.cs:257`). The standalone Dapr admin panel (`Dapr/AdminPanel.Host/Program.cs`) only registers `AddManageableServerRegistry()` + a few services and **never registers `IDictionary<int, IGameServer>`**. Blazor `[Inject]` property injection is mandatory, so when the `/servers` page initializes in that deployment it will throw `InvalidOperationException: Cannot provide a value for property 'GameServers'...` and the whole page fails to render — a regression for the distributed setup.

There's a second layer to this: in the Dapr host, `IServerProvider.Servers` returns `ManageableServerClient` instances (`ManagableServerRegistry.cs:55`), which implement `IManageableServer` but **not** `IGameServer`. So even the in-process workaround of casting `_servers` to `IGameServer` wouldn't work there — the remote game-server abstraction is `MUnique.OpenMU.ServerClients.GameServer`, which isn't wired into a dictionary for the panel to consume.

Suggested directions (any one):
- Make the injection optional — `IDictionary<int, IGameServer>? GameServers` — and conditionally render the message card only when it's non-null, so the feature degrades gracefully (works in the single-process host, hidden in Dapr). This is the smallest change.
- Or register `IDictionary<int, IGameServer>` (mapping ids to Dapr `GameServer` clients) in `AdminPanel.Host` so the feature works there too.
- Or introduce a small deployment-agnostic "broadcast" abstraction registered in both hosts.

Given the project supports both deployments, at minimum the page shouldn't crash on the Dapr host.

### Minor notes (non-blocking)

1. **No error handling around the broadcast.** In `SendMessageAsync`, if a single `SendGlobalMessageAsync` throws mid-loop, the remaining servers are skipped and the message — already cleared from `_message` — is lost with no indication to the admin. Consider a try/catch (per server, so one failure doesn't abort the rest) and some success/failure feedback to the user. Compare with `OnReloadAndRestartClickAsync`, which wraps its call in try/finally.

2. **`Resources.Designer.cs` was hand-edited.** The new properties were appended at the end, but that file is auto-generated and regenerates in alphabetical order — so the next regen will produce a noisy diff reordering everything. Not wrong, just worth being aware of.

3. **No success confirmation in the UI.** After sending, the field clears but there's no toast/inline confirmation that the message went out. Optional polish.

Overall: solid feature for the single-process deployment; the Dapr-host injection failure is the one thing I'd want resolved before merge.

---

nolt:
Thanks for the thorough review — all points were spot on. Pushed a follow-up commit that addresses the blocker and the minor notes.

## 🔴 Blocking: Dapr AdminPanel host — fixed

You're right that `[Inject] IDictionary<int, IGameServer>` is only registered in the single-process host (`Startup/Program.cs`) and would throw on the Dapr `AdminPanel.Host`, taking the whole `/servers` page down.

One subtlety worth flagging: just changing the property to a nullable `IDictionary<int, IGameServer>?` is **not** enough on its own — Blazor's `[Inject]` resolves via `GetRequiredService`, so a missing registration still throws regardless of the nullable annotation. (The existing `IGameServerInstanceManager?` injection looks like a precedent for "optional inject", but that service is actually registered in *both* hosts, so it never exercises the missing-service path.)

So I went with explicit optional resolution:

- Removed the mandatory `[Inject] IDictionary<int, IGameServer>`.
- Inject `IServiceProvider` and resolve once in `OnInitialized` via `GetService<IDictionary<int, IGameServer>>()` (returns `null` instead of throwing).
- The global-message card is only rendered when that field is non-null.

Net effect: full feature in the single-process host, gracefully hidden on the Dapr host (no crash). I did not wire up an `IDictionary<int, IGameServer>` of Dapr `GameServer` clients in `AdminPanel.Host` — that's a larger change (the Dapr server set is discovered dynamically via the registry, so it would need a client factory, overlapping with a proper broadcast abstraction). Left as a possible follow-up if broadcasting from the distributed host is wanted.

## Minor notes

**Error handling around the broadcast — done.** `SendMessageAsync` now wraps each server send in its own `try/catch`, so one failing server no longer aborts the broadcast to the rest. Feedback is surfaced to the admin via `IToastService` (the existing convention, e.g. `CreationPanel`): a success toast when all targets received the message, and a per-server error toast on failure. The typed message is restored when nothing could be sent (all failed, or no running target), so it isn't lost silently; on partial success it stays cleared to avoid re-broadcasting to servers that already received it.

**`Resources.Designer.cs` hand-edit — addressed at the root.** Instead of leaving the new properties appended out of order, I regenerated the whole string-property section in alphabetical (OrdinalIgnoreCase) order, matching `StronglyTypedResourceBuilder` output. This makes the file stable against future regenerations. It does make *this* diff larger (the reorder lands here rather than in a later unrelated commit), but the committed file now matches what the generator produces.

**Success confirmation in the UI — done.** Covered by the success toast above.

### New resource strings
`GlobalMessageSent`, `GlobalMessageSendFailed` (`{0}` = server, `{1}` = error), `GlobalMessageNoTarget`.

### Verification
AdminPanel project builds clean with `-p:ci=true` (0 errors).

sven-n (approved):

sven-n:
Thanks, nice feature :)

PR #798, opened 2026-06-09
Added tests to azure pipeline

ze-dom:
Include tests in the pipeline. I wonder if this wasn't done before for any reason.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,3 +27,9 @@ steps:
 - script: dotnet build --configuration $(buildConfiguration) -p:ci=true
   workingDirectory: src
   displayName: 'dotnet build $(buildConfiguration)'
+
+- task: DotNetCoreCLI@2
+  inputs:
+    command: test
+    projects: 'tests/*Tests/*.csproj'
+    arguments: '--configuration $(buildConfiguration)'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,7 +29,8 @@ steps:
   displayName: 'dotnet build $(buildConfiguration)'
 
 - task: DotNetCoreCLI@2
+  displayName: 'dotnet test'
   inputs:
     command: test
     projects: 'tests/*Tests/*.csproj'
-    arguments: '--configuration $(buildConfiguration)'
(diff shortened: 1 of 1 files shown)


sven-n:
## Review

Thanks for this — running the tests in CI is a clear improvement and the change is minimal and well-targeted. I went through the pipeline and the test projects; below is what I found.

### Looks good ✅
- **`--no-build` is valid here.** All nine `*Tests` projects are referenced in `src/MUnique.OpenMU.sln`, and the existing `dotnet build` step builds the whole solution, so the test binaries already exist by the time this step runs. No redundant rebuild.
- **The glob is correct.** `tests/*Tests/*.csproj` matches the nine test projects and correctly **excludes** `tests/MUnique.OpenMU.Network.Benchmarks` (which is not a test project and would otherwise be a problem). Since the task has no `workingDirectory`, the glob resolves from the repo root, which is where `tests/` lives — so this is right.
- **`DotNetCoreCLI@2 test` auto-publishes TRX results**, so test results will show up in the run without extra config.

### Re: "I wonder if this wasn't done before for any reason"
The most common reason to hold off is integration tests needing external resources (e.g. a database). I checked, and that's **not** a blocker here: the EF Core / DB-dependent tests in `TestInitializationWithEfCore` are marked `[Ignore("This is not a real test which should run automatically.")]`, and the active initialization tests use `InMemoryPersistenceContextProvider`. So no database or container is required — these should run cleanly on the hosted agent.

### Minor notes (non-blocking)
1. **`ci=true` is not passed to the test step.** The build runs with `-p:ci=true`, which skips a code-generation `PreBuild` target (`Condition="'$(ci)'!='true'"`) in a couple of projects. Because the test step uses `--no-build`, those build targets won't run, so this is fine in practice — just flagging it so it's a conscious choice.
2. **PR validation depends on Azure branch policies, not this file.** The `trigger:` is `master` only, so this guarantees tests run on `master` CI; getting them to gate PRs requires a build-validation branch policy in Azure DevOps (out of scope for this PR, just worth knowing).

Overall: 👍 The change is correct as written. The notes above are informational, not requests for changes.

---

sven-n:
Thanks 👍

PR #803, opened 2026-06-17
Fix empty merchant NPC - Christine

nolt:
Fix empty merchant NPC - Christine

Problem

Three Season 6 NPCs have a merchant window (NpcWindow.Merchant) but no MerchantStore, so talking to them opens an empty, non-working shop:

- Christine the General Goods Merchant (545) — Loren Market

Change

Adds a configuration update plugin (AddMissingMerchantStoresPlugIn, UpdateVersion = 84) that clones the general-goods store carried by Thompson the Merchant (231) onto each of the three NPCs. Mirrors the existing pattern in AddKalimaPlugIn (cloning a store onto another NPC).

- Idempotent — skips any NPC that already has a store, and bails out if the source store is missing.
- Optional (IsMandatory = false) — server owners opt in via the admin panel's Updates page.

Result

After applying the update, the three NPCs sell general goods (potions, antidotes, bolts/arrows, town portal scroll, armor of guardsman) and show up in the admin panel's merchant editor for further per-NPC tuning. No engine/behavior changes — data only.

Files

- src/Persistence/Initialization/Updates/AddMissingMerchantStoresPlugIn.cs (new)
- src/Persistence/Initialization/Updates/UpdateVersion.cs (+1 enum entry)

### DIFF
--- a/src/Persistence/Initialization/Updates/AddMissingMerchantStoresPlugIn.cs
+++ b/src/Persistence/Initialization/Updates/AddMissingMerchantStoresPlugIn.cs
@@ -1,0 +1,88 @@
+// <copyright file="AddMissingMerchantStoresPlugIn.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.Persistence.Initialization.Updates;
+
+using System.Runtime.InteropServices;
+using MUnique.OpenMU.DataModel.Configuration;
+using MUnique.OpenMU.PlugIns;
+
+/// <summary>
+/// This update gives a merchant store to a merchant NPC which has a merchant window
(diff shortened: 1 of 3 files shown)


sven-n (changes requested):
Adding an update plugin is correct. However, also the `NpcInitialization` class (maybe `MerchantStores.cs`) needs to be updated as well.
The initialization runs for freshly setup servers, the update plugin for existing servers.

And one thing about Moss (the gambler?): Isn't he selling random items in the original game? As far as I remember, I didn't implement this feature yet.

nolt:
Good catch.
I made a deeper reaserch and Julia is comletly not a stora but a warp master, Lorencia ←→ Loren Market, but also not implemented in OpenMU.

I will fix only Christine, Moss I will leave him untouched. Julia will try fix in another branch.

sven-n:
Thanks

PR #809, opened 2026-06-24
Make Market Union Member Julia a working warp NPC for the Loren Market

nolt:
## Summary

Market Union Member Julia (547) had a merchant window but no store, so talking
to her opened an empty, non-working shop. In the original game she is a warp NPC
that moves players between Lorencia and the Loren Market, on both sides.

This wires her up to the warp window the game client already provides.

## Changes

- Change Julia's NPC window to `JuliaWarpMarketServer`, so talking to her opens
  the client's Julia "Warp" window instead of an empty shop.
- Spawn a second instance of her in Lorencia (the entrance), at (139, 138).
- Add a handler for the `EnterMarketPlaceRequest` (sent by the window's "Warp"
  button) which warps the player to the other side: to the Loren Market when
  talking to her in Lorencia, and back to Lorencia when talking to her in the
  Loren Market.
- Add an optional configuration update (`AddLorenMarketJuliaWarp`) so existing
  servers get the same behaviour without re-initializing.

## Notes

- The game client already implements the window, the button and the packet, so
  no client change is required.
- The warp targets each map's existing spawn gate; the di
  the map the player is currently on.
- The handler only acts while Julia's window is open, so the request can't be
  used as a free teleport.

## Testing

- Fresh Season 6 initialization: Julia is spawned in both Lorencia and the Loren
  Market, with the correct NPC window.
- Applied the configuration update on an existing databas
- Verified in-game: talking to Julia warps the player both ways.

julia


src/GameServer/MessageHandler/EnterMarketPlaceHandlerPlugIn.cs
sven-n, line 1: Please add an action class in the GameLogic project. The packet handler should not have that much game logic, as you can see in other packet handlers.

sven-n (changes requested):
OpenMU is lacking a key feature for loren market... Maps which just exist once for all subservers. We need this for other maps as well: Loren Valley, Crywolf and maybe others.
If you can work out a meaningful system for that, that would be great. If not, just tell me, and I'll accept it anyway.

src/GameServer/MessageHandler/EnterMarketPlaceHandlerPlugIn.cs
nolt: Thanks! I've moved the warp logic into a new EnterMarketPlaceAction in the GameLogic project, so the handler just delegates now.
About the shared "single instance for all subservers" maps (Loren Market, Loren Valley, Crywolf, …) — that's a much larger, cross-cutting change than this PR, so I'd rather not bundle it here. I'll take you up on the offer to merge Julia as-is, and we can design the shared-map hosting as a separate feature. Happy to help with that later if you want to sketch the approach.

sven-n (approved):

PR #817, opened 2026-07-07
feature: add duplicate GUIDs validation

eduardosmaniotto:
This will prevent future merges with duplicated GUIDs.


sven-n:
Good check, thank you

PR #819, opened 2026-07-08
bugfix: standardize loading states and add save spinner to Create pages

eduardosmaniotto:
### Problem 1 — Inconsistent loading states
`Merchants`, `EditConfigGrid`, `CreateGameServerConfig`, and `CreateConnectServerConfig` used inline `spinner-border` elements while data loaded, while other pages used the `LoadingOverlayService` modal overlay. The inline spinners also didn't cover the full page.

**Fix:** Replaced inline spinners with `LoadingOverlayService.ShowLoadingIndicator()` modal overlay in all four pages.

### Problem 2 — `_initState` progress messages never visible
The Create pages had a `string? _initState` that displayed step messages ("Creating configuration...", "Saving...") via `InvokeAsync(StateHasChanged)`. Due to Blazor Server's render batching, intermediate states never reached the client — only the final `null` rendered, so the div never appeared. The progress messages and their `InvokeAsync` wrappers were dead code.

**Fix:** Replaced `_initState` with `bool _isProcessing`. Set to `true` at the top of `OnSaveButtonClickAsync`, reset in a `finally` block (so it always clears). Flows into `AutoForm`'s `IsProcessing` parameter, which disables the Save button and shows a `spinner-border-sm`. Gives immediate visual feedback without relying on intermediate string updates.

### Problem 3 — EF tracking conflict on Merchants discard
`OnCancelButtonClickAsync` called `DiscardChangesAsync` then `LoadDataAsync`, but the old `_selectedMerchant` was still tracked by EF, causing a tracking conflict.

**Fix:** Fix: Capture the merchant's Id before discard. After reload, re-select the merchant by matching the ID against the fresh _viewModels.

### DIFF
--- a/src/Web/AdminPanel/Pages/CreateConnectServerConfig.razor
+++ b/src/Web/AdminPanel/Pages/CreateConnectServerConfig.razor
@@ -11,18 +11,8 @@
 
 @if (this._viewModel is null)
 {
-    <span class="spinner-border" role="status" aria-hidden="true"></span>
-    <span class="visually-hidden">@Resources.Loading</span>
     return;
 }
 
-@if (this._initState is { })
-{
-  <span class="spinner-border" role="status" aria-hidden="true"></span>
-  <span class="visually-hidden">@this._initState</span>
-  return;
-}
-
-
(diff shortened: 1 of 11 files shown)


sven-n:
Thank you

PR #821, opened 2026-07-09
bugfix: prevent quest item pickup by players without active quest

### LINKED ISSUE #761: Quest item pick up by party member without a promotion quest
**Describe the bug**
Quest items can be picked up by party members without a promotion quest.

**To Reproduce**
1. Start a party with one character doing the first promotion quest and another character with the promotion already completed.
2. Hunt monsters that drop quest items and pick up the item with the character that already completed the quest.

**Expected behavior**
Only the quest item owner should be able to pick it up.

**Screenshots**
Image

eduardosmaniotto:
Fixes https://github.com/MUnique/OpenMU/issues/761

## Bug: Quest items can be picked up by party members without active quest

**Problem:** When a monster drops a quest item (e.g., promotion quest items like Scroll of Emperor), ALL party members are set as owners of the dropped item. The only guard — `IsBoundToCharacter` — checks whether the picker is an owner, which passes for all party members. This means characters without the active quest can pick up and hoard quest items.

**Root cause:** `AttackableNpcBase.cs:444` assigns `killer.Party?.PartyList.AsEnumerable()` as owners, and `DroppedItem.TryPickUpByAsync()` only checks ownership, not quest eligibility.

### Changes

**`PickupItemAction.cs`** (+22 lines):
- New `PlayerHasActiveQuestForItem` helper — scans `Character.QuestStates` for any active quest whose `RequiredItems` match the item definition and level
- Quest eligibility check: if item is `IsBoundToCharacter` and player has no active quest requiring it → shows "This item doesn't belong to you." and rejects
- Ownership check: if player isn't an owner and the 10s owner-priority window is active → same message

**`DroppedItem.cs`** (+11 lines):
- `IsPlayerAnOwner` changed from `private` to `public` so `PickupItemAction` can query it
- Added `IsOwnerPickupPriorityActive` property exposing the 10-second owner-priority window

**`PlayerMessage.resx` / `PlayerMessage.Designer.cs`** (+12 lines):
- Added `ItemDoesNotBelongToYou` = "This item doesn't belong to you."

### Behavior

| Scenario | Result |
|----------|--------|
| Player already has 1 copy of quest item (limit=1) | "Limit reached" (existing check, still fires first) |
| Player without active quest picks up quest item | "This item doesn't belong to you." |
| Both party members have active quest, same item drops | First-come-first-served (existing race behavior) |
| Non-party player tries to steal regular item within 10s window | "This item doesn't belong to you." |
| After 10s, regular item is free for anyone | Existing behavior preserved |

dont-belong
dont-belong-pt
limit-reach


ze-dom:
> After 10s, item is free for anyone  | Existing behavior preserved
 
Why not keep it locked to the quest player until it disappears? Otherwise the item will end up blocked in another player's inventory (can't sell, store or trade).

eduardosmaniotto:
> > After 10s, item is free for anyone  | Existing behavior preserved
> 
> Why not keep it locked to the quest player until it disappears? Otherwise the item will end up blocked in another player's inventory (can't sell, store or trade).

That's only for regular items. Quest items will not be able to be picked up after the 10s window. I updated the description to make it clearer; currently if you try to steal a regular item, there is no message to the player. This PR fixes that too.

sven-n:
Thanks, I'll merge it, but I have one hint: The code now implicates that every dropped item which is bound to a character is a quest item, too 😄

eduardosmaniotto:
> Thanks, I'll merge it, but I have one hint: The code now implicates that every dropped item which is bound to a character is a quest item, too 😄

I could create a IsQuestItem in ItemDefinition. It would be a simple check during runtime. What do you think?

sven-n:
Makes sense, it simplifies the code 😄

PR #823, opened 2026-07-10
Feature/speedhack anti cheat

Rhefew:
This PR implements configurable walk and attack speedhack anti-cheat checks for OpenMU. Rather than hardcoding these restrictions, the anti-cheat verification has been designed as a customizable and toggleable feature plugin.
### Key Features
1. **Core Protections**:
   - **Walk Speed Checks**: Compares coordinate changes and time differences to identify walk speed violations.
   - **Attack Speed Checks**: Verifies attack execution intervals using player attack speed parameters to detect rapid attack speedhacks.
2. **Dynamic Configurability**:
   - Integrated via the `IFeaturePlugIn` and `ISupportCustomConfiguration<SpeedHackDetectConfiguration>` interfaces.
   - Admin-configurable settings include:
     - Warning thresholds before enforcement actions occur.
     - Move and attack speed tolerances.
     - Toggleable execution options (auto-ban, player disconnect, or log-only warnings).
3. **Unit Tests**:
   - Full test coverage in `SpeedHackAntiCheatTests` validating detection mechanics, bypasses when the plugin is deactivated, and custom violation response behaviors.

### DIFF
--- a/src/GameLogic/Player.cs
+++ b/src/GameLogic/Player.cs
@@ -83,6 +83,157 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
 
     private Lazy<ComboStateMachine>? _comboStateLazy;
 
+    private const int NormalStepDelayMs = 300;
+    private const int WalkSpeedToleranceMs = 900;
+    private const int MaxAllowedWalkStartOffset = 5;
+    private const int MaxWarningsLimit = 3;
+    private const int AlertDebounceSeconds = 5;
+    private const int WarningHistoryHours = 1;
+    private const double MaxAttackTokens = 5.0;
+    private const double AttackSpeedBaseDelayMs = 450.0;
+    private const double AttackSpeedScalingFactor = 1.2;
+    private const double AttackSpeedMinIntervalMs = 60.0;
+
+    private readonly object _speedHackLock = new();
+
+    private double _attackTokens = MaxAttackTokens;
+    private DateTime _lastAttackTokenUpdateTime = DateTime.MinValue;
+
+    internal DateTime _lastAlertTime = DateTime.MinValue;
+
+    internal readonly System.Collections.Generic.Queue<DateTime> _alertTimes = new();
+
+    private struct WalkHistoryEntry : IEquatable<WalkHistoryEntry>
+    {
+        public DateTime Time;
+        public Point StartPoint;
+
+        public readonly bool Equals(WalkHistoryEntry other)
+        {
+            return this.Time == other.Time && this.StartPoint == other.StartPoint;
+        }
+
+        public override readonly bool Equals(object? obj)
+        {
+            return obj is WalkHistoryEntry other && this.Equals(other);
+        }
+
+        public override readonly int GetHashCode()
+        {
+            return HashCode.Combine(this.Time, this.StartPoint);
+        }
+
+        public static bool operator ==(WalkHistoryEntry left, WalkHistoryEntry right)
+        {
+            return left.Equals(right);
+        }
+
+        public static bool operator !=(WalkHistoryEntry left, WalkHistoryEntry right)
+        {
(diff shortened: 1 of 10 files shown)


src/GameLogic/Player.cs
```diff
@@ -83,6 +83,174 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
 
     private Lazy<ComboStateMachine>? _comboStateLazy;
 
+    private const int NormalStepDelayMs = 300;
+
+    private readonly object _speedHackLock = new();
+
+    private double _attackTokens = 5.0;
+    private DateTime _lastAttackTokenUpdateTime = DateTime.MinValue;
```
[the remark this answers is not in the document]
sven-n: I'd even go further a start a System.Diagnostic.Stopwatch when a player connects and then you can access the elapsed timespan. The TickCount might have a bad resolution (16 ms)!

src/GameLogic/Player.cs
sven-n, line 1: I think this are too many additions to the Player class for a "pluggable" feature 😅
Could you introduce new plugin point interfaces and put all this new logic to one or more plugins? I mean everywhere where you placed code for this feature, add a call to a plugin point instead. If I remember right, the state could also live in a plugin.
If you need more info about how to do this, please let me know 😄

sven-n (changes requested):

src/GameLogic/Player.cs
Rhefew: Yeah, this makes a lot of sense actually, I'll update with your suggestions :)

Rhefew:
I have refactored the speedhack anti-cheat checks to decouple all state and check logic from the core `Player` class, keeping it as an isolated plugin, as per your suggestions, @sven-n 

1. **Decoupled Player Class:** Removed all speedhack-specific fields (warnings, history queues, timestamps, lock objects, and helper methods) from `Player.cs`.
2. **Introduced a New Plugin Point:** Added the `ISpeedHackCheatCheckPlugIn` interface to define walk/attack cheat checks and movement resets.
3. **Encapsulated State inside the Plugin:** Implemented the interface in `SpeedHackDetectPlugIn` and encapsulated player speedcheck state dynamically using a thread-safe nested class mapped via `ConditionalWeakTable`.
4. **Action Handlers:** Refactored action handlers (`HitAction`, `AreaSkillAttackAction`, and `TargetedSkillDefaultPlugin`) and movement routines to use the new plugin point interface.
5. **Unit Tests:** Updated unit tests to register `SpeedHackDetectPlugIn` under the new interface and verify warnings via the plugin instance.

thanks for your input!

src/GameLogic/Player.cs
```diff
@@ -1329,14 +1339,32 @@ public async ValueTask WalkToAsync(Point target, Memory<WalkingStep> steps)
...
         var startPoint = steps.Span[0].From;
+
+        var config = this.GameContext.FeaturePlugIns.GetPlugIn<SpeedHackDetectPlugIn>()?.Configuration;
+        int maxAllowedWalkStartOffset = config?.MaxAllowedWalkStartOffset ?? 5;
+
+        var speedCheck = this.GameContext.PlugInManager.GetPlugInPoint<ISpeedHackCheatCheckPlugIn>();
+        if (speedCheck is { })
+        {
+            var eventArgs = new SpeedHackCheckEventArgs();
+            await speedCheck.WalkCheatCheckAsync(this, steps.ToArray(), eventArgs).ConfigureAwait(false);
```
sven-n: Is `steps.ToArray()` really required? Can't we just pass `steps` as-is?

sven-n (changes requested):

src/GameLogic/Player.cs
Rhefew: Good catch, on it

sven-n:
Thanks 🙂

PR #824, opened 2026-07-10
bugfix: fix invisible item in player storage

eduardosmaniotto:
### Summary
Fixed a bug where items placed in a player's personal store would become invisible after logging out and back in, while still occupying their slot (blocking new items from being placed there). Also fixed an inconsistency where, on login, the store's "open" state could appear open to the owner while still being closed to everyone else (with an empty shop name).

#### Root cause

Items in store slots are backed by the same underlying item storage as the regular inventory, but were never included in the login inventory sync (UpdateInventoryListPlugIn), which only covered equip/inventory/extension slots.
The store's item list was only ever sent to the client via IShowShopItemListPlugIn, which per protocol is meant to open the "browse a shop" dialog — not to passively sync the owner's own items. Sending it to the owner regardless of the store's open state caused a shop dialog to pop up client-side with an empty name (since StoreName is only set once a store has actually been opened), while other players correctly saw nothing.


sven-n (approved):

sven-n:
Thanks :)

PR #827, opened 2026-07-12
feat(plugins): implement customizable maximum concurrent connections

Rhefew:
# Limit Customizable Concurrent Accounts per IP
Exposes a new customizable plug-in that limits the number of simultaneous active player sessions (accounts) that can connect from a single IP address.
## Why this is valuable
Many server administrators want to regulate the number of clients a single player can run in parallel (multiclienting/dual-clienting) to maintain economy balance and server stability. 
## Key Features & Robustness
1. **Customizable Limit**: Defaults to `3` concurrent connections, but can be configured in the admin panel or DB.
2. **State-Aware Checks**: Only sessions that are fully authenticated (`Authenticated` state or higher, including character selection or active gameplay) are counted towards the limit. This prevents users from locking others out simply by opening multiple game instances at the `LoginScreen` or `Initial` states.
3. **Graceful Rejection**: When a login attempt is rejected, it cancels the state transition, triggering a clean response which displays as `Account already connected` in the game client.
4. **Platform & Network Independent**: Operates natively on top of standard `IPEndPoint` properties exposed by the server connection logic.
### Core IP Address Tracking
- **`src/GameLogic/Player.cs`**: Added a public `string? IpAddress` property.
- **`src/GameServer/RemoteView/RemotePlayer.cs`**: Resolved and assigned the remote IP address to the `IpAddress` property during player instantiation.
### Plugin & Configuration
- **`src/GameLogic/PlugIns/MaximumConnectionsPerIpPlugInConfiguration.cs`**: Created a configuration class specifying the `MaximumConnectionsPerIp` limit.
- **`src/GameLogic/PlugIns/MaximumConnectionsPerIpPlugIn.cs`**: Implemented `IPlayerStateChangingPlugIn` to intercept state changes transitioning to `PlayerState.Authenticated` and cancel them if the limit is exceeded.
### Tests
- **`tests/MUnique.OpenMU.Tests/MaximumConnectionsPerIpPlugInTests.cs`**: Added a complete suite of unit tests verifying:
  - Parallel connections below the limit succeed.
  - Parallel connections exceeding the limit get cancelled gracefully.
  - Client connections from different IP addresses are isolated and don't count towards each other.
  - Connections sitting at the `LoginScreen` or `Initial` states are not counted.
## Verification
Executed NUnit tests locally with successful results:
```bash
dotnet test --filter MaximumConnectionsPerIpPlugInTests

### DIFF
--- a/src/GameLogic/IHasIpAddress.cs
+++ b/src/GameLogic/IHasIpAddress.cs
@@ -1,0 +1,16 @@
+// <copyright file="IHasIpAddress.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic;
+
+/// <summary>
+/// Interface for objects that expose a remote IP address.
+/// </summary>
+public interface IHasIpAddress
+{
+    /// <summary>
+    /// Gets the IP address of the remote connection.
+    /// </summary>
+    string? IpAddress { get; }
+}
--- a/src/GameLogic/Player.cs
+++ b/src/GameLogic/Player.cs
@@ -208,6 +208,11 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
     /// </summary>
     public string? IpAddress { get; set; }
 
+    /// <summary>
+    /// Gets or sets a custom login result to override the default when login fails.
+    /// </summary>
+    public Views.Login.LoginResult? LoginResultOverride { get; set; }
+
     /// <inheritdoc cref="IPartyMember" />
     public string Name => this.SelectedCharacter?.Name ?? string.Empty;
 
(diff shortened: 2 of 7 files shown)


sven-n:
Some notes:

* These limits are not ideal for some players. As you might know, IPv4 addresses are limited, and are shared (NAT) at some providers, especially for mobile devices. So, deactivate it by default.
* Please move the IP address property to RemotePlayer, since the Player class should not be aware of network connections.

Rhefew:
> Some notes:
> 
> * These limits are not ideal for some players. As you might know, IPv4 addresses are limited, and are shared (NAT) at some providers, especially for mobile devices. So, deactivate it by default.
> * Please move the IP address property to RemotePlayer, since the Player class should not be aware of network connections.

Yeah, this makes sense. I addressed your comments

PR #828, opened 2026-07-13
feat(admin): add account & character search to accounts page

Rhefew:
# Description
This PR introduces an Account and Character Search functionality to the Web Admin Panel. Currently, the admin panel only offers simple paginated navigation (20 accounts per page), making it extremely tedious to find a specific account on servers with hundreds or thousands of players.
With this change, administrators can search for accounts directly using either the **Account Login Name** or the **Character Name**.
### 1. Web Shared Service
* **`AccountService.cs`**:
  * Added a `SearchFilter` property that triggers a `DataChanged` notification when changed.
  * Enhanced `GetAsync` to execute targeted lookups when a search filter is provided. It queries the repository via `GetAccountByLoginNameAsync` and `GetAccountByCharacterNameAsync` (existing interfaces) and returns the aggregated results.
### 2. Web Admin Panel Page
* **`Accounts.razor`**:
  * Added a `<FilterHeader>` inside the `<DataTable>` component.
  * Embedded a search input matching the design language of the Plugins page, bound to `_accountService.SearchFilter` with `@bind:event="oninput"` for instant, live updates as the admin types.
## Visual reference
The search bar integrates cleanly below the header row, spanning the columns and keeping the "Create" buttons and table structure intact.
## Verification
- Verified clean compilation with `dotnet build`.
- Search operates case-insensitively and updates the list dynamically as you type.

### DIFF
--- a/src/Web/AdminPanel/Pages/Accounts.razor
+++ b/src/Web/AdminPanel/Pages/Accounts.razor
@@ -1,4 +1,4 @@
-﻿@page "/accounts"
+@page "/accounts"
 @using MUnique.OpenMU.DataModel
 @using MUnique.OpenMU.DataModel.Entities;
 @using MUnique.OpenMU.Persistence;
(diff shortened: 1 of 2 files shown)


sven-n:
Thanks

PR #830, opened 2026-07-15
Fix modals making app crash

valentinoConti:
### Description

In many places where a Modal should be shown in the AdminPanel (example: Uploading a file into the terrain data of a map) the app crashes with:

`Error: System.InvalidOperationException: Object of type 'MUnique.OpenMU.Web.Shared.Components.Modal.ModalMessage' does not have a property matching the name 'Modal'.`

This PR fixes that.


sven-n:
Thanks :)

PR #839, opened 2026-07-22
Support externally-provisioned databases (opt-in, no drop/create)

vanvonlj:
### Summary

Adds an opt-in setting that lets the server run against a database it is **not** allowed to create or drop. When enabled and the database has no schema yet, the server keeps the existing (empty) database and builds its schema via migrations, instead of dropping and recreating it.

- Setting: `Database:AssumeExternallyProvisioned` in `appsettings.json`, or the `Database__AssumeExternallyProvisioned` environment variable.
- Default `false` → **unchanged** drop-and-recreate behaviour.
- `-reinit` → always drops and recreates, regardless of the setting.

### Motivation

Today the initial `ReCreateDatabaseAsync()` runs `EnsureDeletedAsync()` + migrate, which requires the connecting role to be able to `CREATE`/`DROP` databases. This works when connecting as `postgres` (or a `DB_ADMIN_USER` with equivalent rights).

It does **not** work in increasingly common managed setups — a Kubernetes operator (e.g. CloudNativePG), infrastructure-as-code, or a managed cloud database — where the database is provisioned ahead of time and the application role is intentionally least-privilege: it owns only its own database and cannot create or drop databases.

In that environment the current behaviour is a crash loop: `EnsureDeleted` drops the database (the role owns it), the following `CREATE DATABASE` fails (no `CREATEDB`), and startup dies with `3D000: database "..." does not exist`.

### Behaviour

`DatabaseExistsAsync()` reports "does the database exist" as "have any migrations been applied", so an empty-but-present database is treated as absent and sent down the recreate path. With `AssumeExternallyProvisioned=true`, that path skips the drop/create and only applies migrations to the existing database — cleanly separating database ownership (the operator/infra) from schema ownership (the server).

### Changes

- `ReCreateDatabaseAsync` gains a backward-compatible `dropExistingDatabase = true` parameter; when `false` it skips `EnsureDeletedAsync`.
- `Program` reads the setting (adds `AddEnvironmentVariables()` to the config builder) and passes `dropExistingDatabase: !assumeExternallyProvisioned`.
- Documented in `src/Startup/Readme.md`; registered in `appsettings.json`.

Verified with `dotnet build` of `Startup` and `Persistence.EntityFramework`.

### DIFF
--- a/src/Persistence/EntityFramework/PersistenceContextProvider.cs
+++ b/src/Persistence/EntityFramework/PersistenceContextProvider.cs
@@ -152,21 +152,30 @@ public class PersistenceContextProvider : IMigratableDatabaseContextProvider
     /// <summary>
     /// Recreates the database by deleting and creating it again.
     /// </summary>
+    /// <param name="dropExistingDatabase">
+    /// If <see langword="true"/> (the default), the database is dropped and created again from scratch.
+    /// If <see langword="false"/>, the existing database is kept and only its schema is built via
+    /// migrations — required when the database is provisioned externally and the connecting role is
+    /// not permitted to create or drop databases.
+    /// </param>
     /// <returns>The disposable that should be disposed of when the data creation process is finished.</returns>
-    public async Task<IDisposable> ReCreateDatabaseAsync()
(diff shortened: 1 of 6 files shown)


sven-n:
Makes sense, thank you

PR #841, opened 2026-07-22
Make the server-side bots hold up as a population

nolt:
## Make the server-side bots hold up as a population

The bot feature works, but a bot left running for a day stops behaving like the
player it is meant to imitate: its backpack silts up, its money stops moving, it
fights with a skill its class should never cast, and the whole population ends up
standing on one map. This branch fixes that, one observed failure at a time.

Everything here came out of running a real population - 400 to 500 bots on a
restored production database - and watching what they actually did. None of it
was visible in review.

---

## What this adds

**Loot and money**

- A bot only picks up the jewels it can use, up to a configured stock, instead of
  hoarding twelve kinds it can never spend.
- Each jewel kind gets its own budget per trip, so a full Bless queue no longer
  starves the Souls and Lifes behind it.
- The merchant visit does its job: it repairs before it sells, sells before it
  destroys, and reports honestly what happened.
- A sale refused by the money limit is reported to the client instead of failing
  silently - a plain player-facing bug, not a bot one.
- The emergency potion supply stays an emergency, so the shopping trip has a
  reason to exist.
- A merchant trip needs the means to pay for it, or something to sell once there.

**What a bot fights with**

- A bot fights with the skills its class actually fights with: castle-siege
  skills are refused, multi-hit skills without a flat bonus are recognized as the
  weapons they are, and reach decides between skills worth about the same.
- A bot learns the skills it already qualifies for when it enters the world, not
  only on its next level-up.
- A pet's skill is only used with the pet: Plasma Storm needs a Fenrir.
- A character no longer buffs itself into paralysis with a skill it kept across a
  reset but can no longer cast - which stopped it attacking, and picking up loot,
  entirely.

**Where a bot stands**

- Bots spread over the hunting grounds of a map instead of piling onto one tile.
- Bots spread over the maps of their level band instead of all taking the single
  strongest one.
- A bot which lands no hit for minutes treats its ground as barren and steps down
  to easier maps until it finds something it can farm.
- A bot pays the warp fare and passes every check the game makes of a player
  warping there; the walk home stays free, so it can never be stranded.
- Shopping trips are spread over time, so fifty bots no longer besiege the same
  Potion Girl at once.

**Running the feature**

- Unchecking `Enabled` stops the bots within a tick instead of doing nothing
  until the next restart. Nothing is deleted.
- `Reset bots` works whatever the number of accounts is, including zero.
- A new `Purge bots` deletes every bot account and switches the feature off - the
  single switch for "no bots on this server".
- Deleting a bot account no longer leaves its item storages and items behind as
  unreachable rows.
- Two settings for the jewel and potion stock, in the admin panel.

---

### The economic loop

Bots stopped progressing after a while: the backpack filled up, the loot pickup
stopped, and the trade log claimed everything was fine. Four causes, all meeting
at the merchant visit.

**The money ceiling is `int.MaxValue`, and `TryAddMoney` refuses at it.**
`SellItemToNpcAction` then did nothing at all - no log, no response to the
client - while the calling handler counted the sale anyway, so the log lied.
`ff16d0f24` makes the action report its outcome and answer the client. A human
player at maximum Zen clicks "sell" and gets no response whatsoever; every other
refusal in that method already reports back.

**Bots picked up all twelve jewel kinds** but can only use Bless, Soul and Life.
`086d59e6c` limits the pickup to those, by an allow-list of (group, number)
rather than a group check - the first attempt collected the whole group of 14.

**A shared budget of two jewels per trip** let the Bless queue, which always had
a target, starve the rules behind it. `ab24f0f86` gives each kind its own budget.

**`IsSellableJunk` cut off at group 12**, which made group 13 - jewellery and
pets - unsellable. Fixed in `42c8d16ce`, together with the ordering of the visit:
repairing first is not cosmetic, since for a bot at the money limit the Zen it
spends on repairs is the only headroom its sales will have.

`027c1aaeb` caps the emergency potion refill. It handed out full stacks of 255,
four times what a bot stocks at a merchant, which made the fallback the bot's
normal supply and left the purchase path dead: the fleet averaged 220 charges and
never visited a merchant. `3b1ef2bdf` and `f831e191f` add the trip's
survivability rules and the two admin-panel settings (`d324a143b`) with their
documentation.

### What a bot fights with

`663e41999` replaces "the skill with the highest `AttackDamage`" with a real
valuation, because that rule picked three kinds of wrong skill:

- **Castle-siege skills** carry the highest numbers in every class, so a
  "strongest first" rule walks straight into them - and the game activates them
  nowhere but the siege map. A Dark Knight is handed Crescent Moon Slash at
  character creation, so every one of them used it. Nothing in the server data
  marks them; the list comes from the client
  (`SkillTooltipModel.cpp: IsCastleSiegeOnlySkill`).
- **Skills with no flat bonus but several hits per attack** - Killing Blow, Chain
  Drive, Dragon Roar, Phoenix Shot, which is the Rage Fighter's entire arsenal -
  scored zero and were not even learned.
- **Reach.** `AttackDamage` is added to the character's own damage, so at 8450
  energy the difference between Hellfire (+120) and Evil Spirit (+45) is about
  5%, paid for with three tiles of range and 70 more mana.

`2758028b0` teaches a bot the skills it already qualifies for when it enters the
world; without it a character which reached the requirement between sessions
would never learn them. `f475e8fcf` keeps the siege roles' support skills out
(Stun is an area skill with no damage and one hit - exactly what Twisting Slash
looks like, so nothing but its number tells them apart).

`793ccc424` is the one a live population surfaced last: **every class which knows
Plasma Storm fought with it**, while riding an Imp or nothing. It is the Fenrir's
skill, and nothing in its own numbers gives the missing pet away - the attribute
behind its damage, `FenrirBaseDmg`, is derived from the character's own strength,
agility, vitality and energy by every class, so it is large for any high-level
character with or without the pet. That put it within the score share where reach
decides, and with range 6 it out-reached the real skills of most classes (Strike
of Destruction has 5). The pet slot is checked now, before the skill is
considered at all.

`ce8924701` is the one which was hiding behind all the others. A character keeps
every skill it learned across a reset, but not the level which unlocked them - a
Blade Knight back at level 12 still owns Swell Life, which asks for level 120.
Buffs which target the party go through the skill plugin, and it refuses an unmet
requirement deep inside, silently. The buff handler reported the cast as applied
regardless, so the effect never became active, the buff was due again next tick,
and the tick ended there - the helper runs Buff, Recover, Obtain Item, Attack, and
returns as soon as a buff was applied. The character stood in the world buffing
something it could not cast, twice a second, for good.

It showed only on melee classes, because only their buffs target the party; the
casters' buffs take a path which applies the effect directly and never consults
the requirements. And only on reset servers, because only a reset leaves a
character holding a skill its level no longer permits - of the bots with no
resets, not one was ever affected. The buff rotation now skips what the character
cannot cast, and the same check closes the equivalent hole in the attack skill
selection, which weighed the consumption costs but not the requirements.

### Where a bot stands

`a34f4322b` stops bots from piling onto the same tile. Two causes at once: **44
of 55 maps have single-tile spawn areas**, so picking a random point "inside" the
area always returned the same tile; and the ground rating fell off linearly from
the bot's position, which flattened everything beyond half a map to the same
weight while every bot entered through the same gate.

`f81616132` spreads the shopping trips: every bot's first check was due
immediately and the cooldown was a constant, so a herd synchronized at startup
stayed in phase forever - fifty bots at one Potion Girl.

`165abf308` fixes two ways a bot could be busy all day and earn nothing:

- **The merchant loop.** A bot low on potions went shopping whether or not it
  could pay. Broke, it bought nothing, so the trigger still stood when it got
  back and it set off again. The emergency refill keeps such a bot alive at a
  stock deliberately below the shopping target, which is what turned an
  occasional wasted trip into a permanent loop.
- **Barren ground.** A map can pass every safety check and still pay nothing: the
  monsters the bot may fight are a rare kind among ones it must refuse, or other
  hunters empty the grounds first. The bot now notices the way a player would -
  no hit landed in minutes - and steps down to the strongest map easier than the
  one it stands on, one notch at a time.

`9f88e34eb` draws the map instead of maximizing it. A bot warped to the map with
the strongest monsters it could survive, which is one map per level band, so the
band's other maps stood empty - 208 bots on Vulcanus at 2.2 monsters each, and
nobody at all on Tarkan, Karutan or Kanturu, which their level opens just as
much. The best map now wins about a third of the picks and the runners-up split
the rest. The candidates are collected exactly as before, so each is legally
warpable, meets the map's requirements, and is better than where the bot stands;
the draw only decides between improvements.

`1264c17bd` makes a bot pay its way. It warped by being placed on the target gate
directly, which skipped the fare the game charges - 2.000 to 18.000 Zen depending
on the map. The navigator had grown its own copy of two of the three checks the
warp makes and simply did not have the third. It goes through `WarpAction` now,
so it pays and inherits every check, including any added later. The retreat to
the class home town stays free and now also catches a bot whose ground has gone
barren: without that, charging for the trip would have rebuilt the poverty trap
this branch already fixes for repairs - no money, no way out, no way to earn.

### Running the feature

Three settings in the admin panel read as if they controlled the population, and
none did what they say. `1126656d9`:

- Unchecking `Enabled` did nothing until the next restart; it now stops the bots
  within a tick, without deleting anything, so checking it again brings the same
  characters back.
- `Reset bots` was skipped whenever the number of accounts was zero, because the
  server which carries out a reset was picked as the owner of the *first* bot
  account - so the one setting which says "I want no bots here" was also the one
  which disabled the code that removes them. The role now follows from the set of
  game servers alone.
- Deleting them for good was not possible at all. The new `Purge bots` deletes
  every bot account and switches the feature off - which is what makes it a
  purge, since the same pass would otherwise generate the population again right
  after deleting it.
- A flag which is set but cannot be acted upon is logged instead of ignored. That
  silence is why a purge which never ran looked exactly like one with nothing to
  delete.

`0e9c8477a` fixes what those deletions left behind. A character's inventory is
referenced *by* the character, so no delete cascade reaches it, and the paged
query which finds the bot accounts returns them untracked and without their
characters - so the aggregate walk found nothing to remove either. Purging 80
accounts left 480 storages and 11.725 items in the database as rows nothing can
reach. The account is now loaded with its whole graph, and its vault and each
character's inventory are deleted explicitly.

---

## The one deliberate policy decision

**Destroying junk which cannot be sold** is the part worth arguing about.

At the money limit a sale is impossible, and there is no other way out of the
backpack: a bot cannot trade, and dropping is not an option either - the server
does not stop it, but no human player can drop an upgraded or excellent item, so
a bot doing it would both look wrong and put items on the ground which could not
have come from a player. The alternative to destroying is a bot which stops
picking up loot forever.

It is bounded on purpose: only items already classified as junk, only while the
backpack is under slot pressure, and it is logged with its reason. The repair
cost this branch adds is a real Zen sink, so most bots should never reach the
limit in the first place.

## Note on the jewel purchase

No merchant in the seed sells jewels - all fourteen stores in `MerchantStores.cs`
were checked. That branch never fires on a default configuration; it is there for
servers which add jewels to their shops. Happy to drop it if you would rather not
carry code the default setup never reaches.

---

## Testing

`MUnique.OpenMU.Tests`: **610 passed, 0 failed.** Added over the branch: a
regression test for the starved Soul rule, the per-kind jewel budget, the skill
gate (`BotSkillRepertoireTest` - every castle-siege number refused, multi-hit and
area skills without a flat bonus accepted, a pet's skill recognized as one, a
plain single-hit skill without a bonus still refused; verified against the
previous rule, where 9 of its 11 cases fail), the hunting-ground rating
(`BotHuntingGroundTest`), and the server which carries out a reset or purge
(`BotServerPartitionTest`, including the zero-accounts case that used to disable
it).

Beyond the unit tests, everything above was run against a **restored production
database** with 400 to 500 bots. That is where most of these commits come from;
every one of those problems was invisible in review and obvious within minutes of
watching real bots.

| | before | after |
|---|---|---|
| destroyed Chaos / Creation jewels | 511 / 6702 | 27 / 2433 |
| items per bot | 80 | 43 |
| jewels bought | 6 | 215 |
| destroyed-to-sold ratio | 13:1 | 1.5:1 |
| distinct hunting grounds chosen (20 min) | 2650 of 6520 | 4598 of 5902 |
| bots on the single most popular tile | 73 | 7 |
| bots on the most popular map | 208 | 93 |
| maps populated | a handful | 17 |
| population earning experience | 82% | 99.75% |
| below level 150, earning experience | 1 of 71 | all of them |
| orphaned rows left by deleting 80 bot accounts | 480 storages, 11.725 items | none |

The purge and reset semantics were exercised against the same database, one
server restart per case, counting unreachable rows after each: a purge with the
feature enabled, a fresh generation, a reset onto a smaller population, a reset
with zero accounts configured (which previously did nothing at all), and a purge
with the feature already disabled. Switching `Enabled` off and on again was
tested live from the admin panel: 340 bots logged out within a tick, nothing was
deleted, and switching it back on brought back the same characters - verified by
comparing a hash of the bot character ids before and after.

An 11-minute observation of the running population after the last commit: 442
warps from 220 bots, 96% of them paid, 4% free walks home, no bot exceeding the
warp cooldown, 16 maps populated with the largest share at 24%, one caught engine
exception of a known pre-existing kind, and 1.340 items created with no new
orphaned rows.

## Known limitations

Unchanged from the feature's own documentation. The one this branch opened with -
about a fifth of a population earning nothing at all - is closed: 99.75% of a 400
bot population now gains experience, against 82% before.

### DIFF
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -65,6 +65,14 @@ The *Bots* feature plugin, in the "Feature Plugins" section of the admin panel:
 - **`Bots pay reset costs`** — whether bots pay the configured zen and item
   costs for their resets. Off by default: they take no part in the economy those
   costs are balanced for.
+- **`Jewel stock per kind`** — how many Jewels of Bless, Soul and Life a bot
+  keeps of each kind. Above it, it stops picking them up and sells what it
+  already carries. Depends on the server's drop rates: on high rates a small
+  stock keeps the backpack free, on low rates a larger one is never reached
+  anyway.
+- **`Potion stock charges`** — how many charges of healing and of mana potions a
+  bot restocks to at a merchant. Depends on what the shops sell: a server whose
+  potions come in stacks of 255 fills the target in a single purchase.
 - **`Reset bots`** — purges and regenerates the whole population on the next
   start, then clears itself.
 
@@ -136,13 +144,27 @@ weapon. If the engine refuses the equip after all, the old gear goes straight
 back on. The replaced piece stays in the backpack and is sold on the next trip
 to town, rather than being dropped where the next bot would pick it up again.
 
-When the backpack fills up or the potions run low, the bot walks to a merchant,
-sells its junk, and buys refills from the proceeds while the shop dialog visibly
-occupies it — on a map without a merchant it warps home first, like a player
-would. Looted Jewels of Bless, Soul and Life are spent on its own equipment
-through the regular consume action, with the same success rates and failure
-penalties a player faces, and with the caution a player shows: a Soul is only
-risked where a failure cannot destroy the item's level.
+A merchant trip is the only moment a bot can turn loot into anything, so it goes
+whenever it has something to gain there: the backpack is filling with junk, the
+potions are running low, a jewel is waiting to be used, or a surplus is waiting
+to be sold. It picks the merchant which sells what it needs right now, and on a
+map whose merchants sell no potions while it needs some, it warps home to a real
+town instead. While the shop dialog visibly occupies it, the bot sells its junk,
+repairs its gear — which is what earns the NPC's discount — and buys potions and,
+where a shop offers them, jewels.
+
+Only Jewels of Bless, Soul and Life are collected, and only up to the configured
+stock: a bot cannot trade or craft, so any other kind would be a backpack slot it
+never gets back. They are spent on its own equipment through the regular consume
+action, with the same success rates and failure penalties a player faces, and
+with the caution a player shows: a Soul is only risked where a failure cannot
+destroy the item's level.
+
+A bot which reaches the server's maximum inventory money can no longer sell
+anything — the money simply does not fit. What it cannot sell it keeps, and
+destroys only what has no other way out: jewels beyond its stock, and, while the
+backpack is genuinely full, junk gear. The repair bill is what normally keeps it
+away from that limit in the first place.
 
 Wings do not drop, so bots earn them at the classic milestones instead — the
 first pair at level 180, the second at 280 and the third, master-only pair at
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -96,18 +96,51 @@ chance to be hit — a monster's nominal level says little about its punch on th
 high-end maps. An agility build's dodge therefore counts as the defense it
 really is, and better gear opens tougher maps, exactly like for a player.
 
-Map access follows the game's warp list: a bot enters a map only if its level
-may legally warp there, and a bot which finds itself on a map it may not be on
-(after a reset, for instance) leaves for the best map it may use. The map it
-reached is persisted, so a restarted bot wakes up where it stopped.
+Map access follows the game's warp list, and a bot travels on a player's terms:
+it enters a map only if its level may legally warp there, it meets the map's own
+requirements, and it pays the warp's fare out of its own Zen. A bot which finds
+itself on a map it may not be on (after a reset, for instance) leaves for the
+best map it may use. The map it reached is persisted, so a restarted bot wakes up
+where it stopped.
+
+Which map it goes to is drawn rather than maximized. The maps of a level band
+differ by a few monster levels, so always taking the strongest one made it every
+bot's answer and left the rest of the band deserted. The best map still wins
+about a third of the picks and the runners-up split the rest, so the population
+spreads over the maps a player of that band would choose between. Every map in
+the draw is one the bot may legally reach, can afford, and which is better than
+where it stands - the draw only decides between improvements.
+
+A map can pass every check and still pay nothing: the monsters the bot may fight
+are a rare kind among ones it must refuse, or other hunters empty the grounds
+first. The bot notices the way a player would, by having landed no hit in
+minutes, and steps down to easier ground one notch at a time until it finds
+something it can farm; the regular map choice carries it back up as its level and
+gear recover. A bot which cannot afford any trip at all walks home to its class
+town for free, so it is never stranded on ground it cannot earn on.
 
 ### Fighting and progressing
 
 A bot fights with the strongest skill of its class it has learned and can pay
-for; casters keep their distance and drink mana. Skills are learned against the
+for; casters keep their distance and drink mana. Between skills worth about the
+same it takes the one with the longer reach - the flat bonus of a spell is a
+rounding error next to a high-level character's own damage, while three tiles of
+range are three tiles at any level. Skills the game only activates during a
+castle siege are left out, and so are a pet's skills unless the pet is actually
+equipped: Plasma Storm draws its damage from the Fenrir, but the attribute behind
+it is derived from the character's own stats, so nothing but the pet slot tells a
+mounted character from one riding nothing. Skills are learned against the
 game's own requirements — total energy, leadership, character level — at
 generation and again on every level-up, and the class buffs are kept up on their
 own.
+
+A skill the character cannot currently cast is passed over, in the attack
+rotation and in the buffs alike. That is not the same as not having learned it: a
+reset keeps every skill but takes back the level which unlocked it, so a veteran
+back at level 12 still owns Swell Life, which asks for level 120. The game
+refuses such a cast silently, so a character which kept trying would simply stand
+there — buffing something that never takes effect, and never getting as far as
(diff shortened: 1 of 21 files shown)


sven-n:
That's crazy 😂

PR #842, opened 2026-07-23
Cascade delete mini game ranking entries

### LINKED ISSUE #796: [BUG] Minigame ranking entry FK blocks character deletion
# Characters that have a `MiniGameRankingEntry` cannot be deleted (FK has no cascade)

## Description

Deleting a character fails when that character has any row in `MiniGameRankingEntry`.
It happens both via the Admin Panel and in‑game (the save on
`Player.RemoveFromGameAsync` → `SaveProgressAsync` throws), with:

```
Npgsql.PostgresException 23503: update or delete on table "Character"
violates foreign key constraint "FK_MiniGameRankingEntry_Character_CharacterId"
on table "MiniGameRankingEntry"
```

## Root cause

`MiniGameRankingEntry.Character` is an *optional* navigation:

```csharp
// src/DataModel/Statistics/MiniGameRankingEntry.cs
public virtual Character? Character { get; set; }
```

and the entity is registered without any relationship/delete configuration:

```csharp
// src/Persistence/EntityFramework/EntityDataContext.cs
modelBuilder.Entity<MiniGameRankingEntry>();
```

For an optional relationship EF Core defaults the FK to `NO ACTION`, so the database
rejects the character delete. Every other character‑owned relationship is explicitly
cascaded (see `CharacterExtensions.cs`: letters, skills, stat attributes, quest states,
drop‑item‑groups, guild membership all use `OnDelete(DeleteBehavior.Cascade)`).
`MiniGameRankingEntry` was simply missed — so it is the only one of the seven
`Character` foreign keys that blocks deletion.

## Impact

Any character that ever placed in a mini‑game ranking (Blood Castle / Devil Square /
Chaos Castle, etc.) can no longer be deleted, by admins or players.

## Reproduce

1. Play a mini‑game with a character so it gets a `MiniGameRankingEntry`.
2. Try to delete that character (Admin Panel or in‑game) → FK violation `23503`.

## Two possible fixes (design decision)

- **A — Cascade (simplest, consistent).** Configure the relationship with
  `OnDelete(DeleteBehavior.Cascade)` so ranking entries are removed together with the
  character. Matches how all other character‑owned data behaves. Ranking entries are
  scoped per game instance (`GameInstanceId`), i.e. transient, so dropping them on
  character deletion is reasonable.

- **B — Keep history (`SET NULL` + snapshot).** If ranking entries should survive
  character deletion (a persistent leaderboard), change the FK to `SetNull` (the
  navigation is already nullable) and store a denormalized character‑name snapshot on
  the entry, updating the ranking view to use it. More involved; only worth it if
  persistent leaderboards are intended. The current `NO ACTION` implements neither
  intent — it just blocks deletion.

## Note on existing databases

A model/config change only fixes *new* schemas. Already‑created databases also need a
migration (or a manual `ALTER TABLE data."MiniGameRankingEntry" ... ON DELETE
CASCADE/SET NULL`), since EF won't retroactively alter an existing FK.

## Immediate workaround

Delete the character's ranking rows first, then delete the character:

```sql
DELETE FROM data."MiniGameRankingEntry" WHERE "CharacterId" = '<character-id>';
```

or apply the cascade once at the DB level:

```sql
ALTER TABLE data."MiniGameRankingEntry" DROP CONSTRAINT "FK_MiniGameRankingEntry_Character_CharacterId";
ALTER TABLE data."MiniGameRankingEntry" ADD CONSTRAINT "FK_MiniGameRankingEntry_Character_CharacterId"
  FOREIGN KEY ("CharacterId") REFERENCES data."Character"("Id") ON DELETE CASCADE;
```

nolt:
Fixes #796.

## What this changes

- A character which has ever placed in a mini game ranking (Blood Castle, Devil
  Square, Chaos Castle, …) can be deleted again — both by the player and by an
  administrator in the admin panel.
- A mini game definition can be deleted again in the configuration editor. It
  hit the very same restriction, so it is fixed in the same pass.
- Existing databases are migrated: the two foreign keys of
  `MiniGameRankingEntry` are recreated with `ON DELETE CASCADE`.
- No behaviour changes for anything else — this only defines what happens to a
  ranking entry when the row it points at goes away.

## The defect

Deleting such a character failed with

```
Npgsql.PostgresException 23503: update or delete on table "Character"
violates foreign key constraint "FK_MiniGameRankingEntry_Character_CharacterId"
on table "MiniGameRankingEntry"
```

`MiniGameRankingEntry` is a member of neither the character's nor the mini game
definition's aggregate. The model generator emits a cascading delete only for
navigations marked with `MemberOfAggregateAttribute`
(`Persistence/SourceGenerator/EfCoreModelGenerator.cs`), and the entity was
registered in `EntityDataContext` without any further configuration:

```csharp
modelBuilder.Entity<MiniGameRankingEntry>();
```

So Entity Framework fell back to `NO ACTION` for both optional relationships,
which PostgreSQL implements as a restricting foreign key.

This looks like an oversight rather than a decision: the blanket
`20221121161552_DeleteCascade` migration gave every other character-owned
relationship a cascade — `CharacterQuestState`, `SkillEntry`, `StatAttribute`,
`LetterHeader` — and skipped this one. It was the only one of the seven foreign
keys pointing at `Character` that blocked deletion.

## The fix

Configure both references explicitly:

```csharp
public static void Apply(this EntityTypeBuilder<MiniGameRankingEntry> builder)
{
    builder.HasOne(entry => entry.RawCharacter).WithMany().OnDelete(DeleteBehavior.Cascade);
    builder.HasOne(entry => entry.RawMiniGame).WithMany().OnDelete(DeleteBehavior.Cascade);
}
```

…plus the matching migration, since a model change alone would only fix newly
created databases.

Ranking entries are scoped per game instance (`GameInstanceId`), so dropping
them together with the character is consistent with how the rest of the
character's data behaves. Option B from the issue — `SET NULL` plus a
denormalized name snapshot — would only be worth it if persistent leaderboards
were intended; the current `NO ACTION` implements neither.

### On soft deletion

You mentioned in the issue that for the normal use case you are thinking about
soft deletion (an `IsDeleted` flag) of character and account data. That is a
larger and separate discussion, and this change neither implements nor blocks
it — but it also does not depend on it, for three reasons:

1. **Hard deletion has to keep working either way.** Even with soft deletion in
   place, rows eventually get removed for real: administrative cleanup, data
   retention / erasure requests, purging generated test or bot populations. All
   of those run into this exact foreign key. Leaving it as `NO ACTION` just
   moves the failure to a rarer, less convenient moment.
2. **It does not help the databases that are broken today.** A soft-delete flag
   is a model plus a migration plus query filters — and it would have to be
   mirrored in the InMemory provider, plus partial unique indexes so that the
   name of a deleted character can be reused. That is a cross-cutting change;
   this one is two lines and a migration, and it unblocks affected servers now.
3. **The accident case is already covered on both deletion paths.** Deleting a
   character as a player is not a stray click: `DeleteCharacterAction` verifies
   the account's security code, or the account password when no security code is
   set, and refuses outright while the character is in a guild. In the admin
   panel it takes two deliberate steps — removing the character from the account
   and then saving. So the value soft deletion adds here is not protection from
   mis-clicks; it is the ability to restore on request and to keep an audit
   trail, which is a feature decision rather than a fix for this defect.

Happy to be wrong on the priority — but the cascade seems worth having whether
or not soft deletion lands later.

### Note on the second cascade

The cascade on `MiniGameDefinition` goes one step beyond the reported symptom,
so it should be a conscious decision rather than a side effect: it means that
deleting a mini game definition in the admin panel now also drops that game's
ranking history, where it previously failed with the same FK error. Happy to
split it out if you would rather keep this PR to the character path alone.

## Verification

- `dotnet build` of `MUnique.OpenMU.Persistence.EntityFramework`: succeeds, no
  new warnings (the StyleCop warnings under `Migrations/` are pre-existing, from
  `AddIsQuestItemFlag` and `AddBuff`).
- The model snapshot keeps `HasForeignKey("CharacterId")` / `("MiniGameId")` and
  only gains `OnDelete(DeleteBehavior.Cascade)` — no shadow foreign key was
  introduced by configuring a previously convention-discovered relationship.
- All seven relationships pointing at `Character` now cascade.
- `Down()` restores the previous `NO ACTION` foreign keys.

## Commits

- `Cascade delete mini game ranking entries` — model configuration, migration
  `20260723073950_CascadeDeleteMiniGameRankingEntries`, snapshot update.

### DIFF
--- a/src/Persistence/EntityFramework/EntityDataContext.cs
+++ b/src/Persistence/EntityFramework/EntityDataContext.cs
@@ -44,7 +44,7 @@ public class EntityDataContext : ExtendedTypeContext
         modelBuilder.Entity<Model.AttributeDefinition>();
         modelBuilder.Entity<ConnectServerDefinition>();
         modelBuilder.Entity<ChatServerDefinition>();
-        modelBuilder.Entity<MiniGameRankingEntry>();
+        modelBuilder.Entity<MiniGameRankingEntry>().Apply();
         modelBuilder.Entity<GameServerDefinition>(entity =>
         {
             entity.Property(e => e.PvpEnabled).HasDefaultValue(true);
--- a/src/Persistence/EntityFramework/Extensions/ModelBuilder/MiniGameExtensions.cs
+++ b/src/Persistence/EntityFramework/Extensions/ModelBuilder/MiniGameExtensions.cs
@@ -4,6 +4,7 @@
 
 namespace MUnique.OpenMU.Persistence.EntityFramework.Extensions.ModelBuilder;
 
+using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using MUnique.OpenMU.Persistence.EntityFramework.Model;
 
@@ -40,5 +41,22 @@ internal static class MiniGameExtensions
     {
         builder.Property(p => p.Name).HasConversion(LocalizedStringConverter.Instance);
         builder.Property(p => p.Description).HasConversion(LocalizedStringConverter.Instance);
+    }
+
+    /// <summary>
+    /// Applies the settings for the <see cref="MiniGameRankingEntry"/> entity.
+    /// </summary>
+    /// <param name="builder">The builder.</param>
+    /// <remarks>
+    /// A ranking entry is a member of neither the character's nor the mini game definition's
+    /// aggregate. The model generator only emits a cascading delete for navigations marked with
+    /// <see cref="MUnique.OpenMU.DataModel.Composition.MemberOfAggregateAttribute"/>, so these two
+    /// references are left without a delete behavior; Entity Framework then defaults to no action
+    /// for them, and the database refuses to delete a character which has played a mini game.
+    /// </remarks>
+    public static void Apply(this EntityTypeBuilder<MiniGameRankingEntry> builder)
+    {
+        builder.HasOne(entry => entry.RawCharacter).WithMany().OnDelete(DeleteBehavior.Cascade);
+        builder.HasOne(entry => entry.RawMiniGame).WithMany().OnDelete(DeleteBehavior.Cascade);
     }
 }
--- a/src/Persistence/EntityFramework/Migrations/20260723073950_CascadeDeleteMiniGameRankingEntries.Designer.cs
+++ b/src/Persistence/EntityFramework/Migrations/20260723073950_CascadeDeleteMiniGameRankingEntries.Designer.cs
@@ -3,6 +3,7 @@ using System;
 using MUnique.OpenMU.Persistence.EntityFramework;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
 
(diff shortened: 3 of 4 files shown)


sven-n:
Yea, I menitioned soft-delete because it's strange to have inconsistent rankings. Imagine, someone won mini games, some are second place... now the winner guy deletes his character... who won now? 😂
Anyway, it's probably not a big issue for now, so I'll accept this PR.
Thank you!

nolt:
Yea I thought about this but again someone could ask or event insist that he wants to not be printed on that list cuz he delete character. I know it's difficult to resolve atm, anyways thanks for approve.

PR #846, opened 2026-07-25
Fix zen from own experience

nolt:
## Zen from own experience

| Commit | Subject |
|---|---|
| `53f25aca0` | Derive each player's zen from the experience they actually gained |
| `7cdb34b7c` | Keep a money basis for party members at max level without master quest |
| `468ada58c` | Let strangers pick up party money and split it equally |

No new configuration property, no EF migration, no schema change. Only `GameLogic`
and tests are touched.

## Background

The money amount of a monster drop is computed in one place,
`src/GameLogic/DefaultDropGenerator.cs:517`:

```csharp
droppedMoney = (uint)(gainedExperience + BaseMoneyDrop);   // BaseMoneyDrop = 7
```

So the zen of a kill is the experience gained from it, plus a flat 7. The drop item
group (`SpecialItemType.Money`, 50 % chance) only decides *whether* this runs, never
*how much*. This formula is unchanged by this branch.

### 1. Two different quantities landed in the same variable

`src/GameLogic/NPC/AttackableNpcBase.cs:310`:

```csharp
int exp = await (player.Party?.DistributeExperienceAfterKillAsync(this, player)
                 ?? player.AddExpAfterKillAsync(this));
```

- solo — `AddExpAfterKillAsync` returns the value from `CalculateExpAfterKill`
  (`Player.cs:1209` → `:1230`), **including** the game rate and the personal
  experience rates (`Player.cs:1257-1259`)
- party — `DistributeExperienceAfterKillAsync` returned the total from
  `CalculatePartyExperience` (`Party.cs:489` → `:496`), **excluding** the game rate
  and the personal rates; those were applied later, per member, in
  `AwardExperienceAsync`

On a server with `ExperienceRate = 500` the party number was therefore roughly 500x
too small, and since the money is derived from it, party zen was correspondingly tiny.

### 2. The workaround for it pinned the party pool to a solo-sized amount

`AttackableNpcBase.cs:411-422` (introduced by PR 778) discarded the party
value and recalculated the killer's solo experience for money purposes. That cured the
symptom, but the party money pool no longer grew with the party size, while the
experience pool does (`Party.cs:350-352`: `base x memberCount x 1.05^(n-1)`).

For a party of three level 400 characters on the same monster:

| | experience | zen |
|---|---|---|
| party pool | `3.31x` solo | `1x` solo |
| per member | `~1.10x` solo | `~0.33x` solo |

### 3. Zen was split evenly, experience by level

Zen: `DroppedMoney.cs:147` and `Party.cs:241` divided by the member count.
Experience: `Party.cs:346`, `:358`, `:389`/`:399` weight it by each member's level.

A level 400 and a level 10 character in the same party received identical zen while
their experience differed by a factor of ~40.

### 4. `MoneyAmountRate` applied zero, one or two times

The multiplication in `AttackableNpcBase.cs:407` predates PR 778 (added in
`d2536e9eb`). PR 778 added a second one in the new party branch
(`DroppedMoney.cs:151`), so the bag - already scaled by the killer's rate - was scaled
again by the receiver's.

| Path | Where | Times applied |
|---|---|---|
| `ShouldDropMoney=false`, solo | `AttackableNpcBase.cs:397` | **0** |
| `ShouldDropMoney=false`, party | `Party.cs:244` | 1 |
| `ShouldDropMoney=true`, solo | `AttackableNpcBase.cs:407` | 1 |
| `ShouldDropMoney=true`, party | `AttackableNpcBase.cs:407` **and** `DroppedMoney.cs:151` | **2** |

Invisible at the default `MoneyAmountRate = 1.0`; at 3.0 a party was paid 9x.

### 5. `ClampMoneyOnPickup` was ignored in a party

The clamp lived only in the solo branch (`DroppedMoney.cs:163-173`). The party branch
called `TryAddMoney` directly, so a member at `MaximumInventoryMoney` silently lost their
share while the drop was consumed by the others.

## What was changed

**Each player's zen is derived from the experience that player actually gained.**

`AwardExperienceAsync` (`Party.cs:380`) already computed the per member experience with
all rates applied - it just discarded it. It now returns that value,
`InternalDistributeExperienceAfterKillAsync` returns the per member breakdown, and the
money is split proportionally to it. The workaround from point 2 is gone; the pool follows
the experience, and each member's zen matches their own level, their own rates and their own
master/normal branch.

New types, all internal to the game logic:

| File | Purpose |
|---|---|
| `src/GameLogic/ExperienceShare.cs` | the experience a single player gained from a kill |
| `src/GameLogic/MoneyShare.cs` | the part of a money drop reserved for a single player |
| `src/GameLogic/MoneyDistribution.cs` | all share arithmetic, and the **single** place where `MoneyAmountRate` and the pickup clamp are applied |

Consequences:

- `MoneyAmountRate` is applied exactly once, always for the receiver, on their own share
- the clamp runs on the shared payout path, after the rate, so it clamps the amount
  actually credited
- the units lost to the integer division go to the shares which were cut the most, one
  each (largest remainder method), so no member is systematically favoured
- shares of players who are no longer eligible are redistributed among the rest instead
  of being lost
- money without a per player breakdown - the fixed amount of an item box
  (`ItemBoxDroppedPlugIn.cs:42`) - is still split equally

### Follow-up commit `7cdb34b7c`

Making the money follow the *awarded* experience introduced a regression: a character at
the maximum level which has not completed the master quest gains neither normal nor
master experience, and `AwardExperienceAsync` returned 0 for it - so it received no zen
at all in a party, while the same character still earns zen solo (`AddExpAfterKillAsync`
returns the calculated amount regardless of whether it could be applied).

The normal experience is now always calculated and returned, and only awarded when the
character can still gain it.

### Follow-up commit `468ada58c` - money still has no owner

Recording the shares on the drop introduced a second, subtler regression. Unlike a
dropped item, money has never had an owner: it can be picked up by anyone standing on it
(`PickupItemAction.CanPickup` only checks that the picker is alive and within three tiles -
there is no ownership timer). The first version paid strictly the recorded shares, so a
stranger who picked the money up got nothing, and the drop was left lying on the ground -
which, for a party of strangers, silently destroyed money that on `master` they would
have taken.

The recorded shares now only apply when the party which picks the money up is the one
which earned it. For anyone else there is no experience of theirs to follow, so the money
is split equally between the picking party, exactly as `master` did and as money without
any shares (e.g. from an item box) already is:

```csharp
var earnedByThisParty = this._shares.Any(share => party.IsEligibleForMoney(share.Player, player));
var shares = earnedByThisParty
    ? this._shares
    : MoneyDistribution.CreateEqualShares(this.Amount, party.PartyList.OfType<Player>().ToList());
```

So the earner's party gets the money proportional to the experience it gained, a solo
picker takes it all, and a stranger's party splits it equally - none of which can strand
the money. Covered by `DroppedMoneyTest.StrangerPartySplitsPickedUpMoneyEquallyAsync`
and `EarningPartyReceivesMoneyByExperienceShareAsync`.

## What was deliberately not changed

- **`zen = experience + 7`** stays. Raising `ExperienceRate` still raises zen exactly as
  before.
- **The master experience rate still applies to zen.** Once a character reaches the
  maximum level as a master class, the formula switches from `ExperienceRate` to
  `MasterExperienceRate` (`Player.cs:1253-1254`). With `500` and `200` configured, the
  same monster pays 40 % of what it paid before the character maxed out. This is
  pre-existing behaviour and was left as is by decision.
- **No new configuration field.** The knob for scaling the zen amount remains the
  existing `Stats.MoneyAmountRate` global base attribute (`Stats.cs:153`, default 1.0,
  set up in `GameConfigurationInitializerBase.cs:275`). It is reachable in the admin
  panel under *Game Configuration -> Full configuration -> Global Base Attribute Values
  -> Money Drop Amount Rate* - not under *General*, which is rendered with
  `hide-collections` and therefore omits it. The value is read when a character's
  attribute system is built, i.e. at login.
- **Item drops** are unrelated to experience and were not touched.
- **The fixed money amount of an item box** (e.g. a Box of Kundun) is unchanged in how it
  is divided - still split equally between the picking party, since a box has no per player
  experience to follow. It does now go through the single `MoneyAmountRate` application in
  `MoneyDistribution.TryPay`, which means a **solo** pick up is scaled by the rate too;
  `master` scaled item-box money only on the party path, so this makes the rate consistent
  rather than depending on party state. At the default rate of `1.0` this is invisible.

### Unit tests

`dotnet test src/MUnique.OpenMU.sln -p:ci=true` - **1252 tests, 0 failures**, no new
StyleCop warnings. New coverage in
`tests/MUnique.OpenMU.Tests/MoneyDistributionTest.cs` (14 tests), one added case in
`ExperienceRateSplitTest.cs`, and two in `DroppedMoneyTest.cs`, covering: shares
proportional to experience, shares adding up to the dropped amount, a player with no
experience getting no share, fair remainder distribution, `MoneyAmountRate` applied
exactly once solo and in a party, the clamp honoured per member, redistribution of
ineligible shares, item-box money still split equally, the max-level-without-master-quest
money basis, a stranger's party splitting picked-up money equally, and the earning party
receiving it by experience share.

Everything is built and tested inside a container
(`mcr.microsoft.com/dotnet/sdk:10.0-alpine`, `-p:ci=true` to skip the code generation
pre-build steps).

### Live test with bots

An isolated stack (own volume, network and ports, separate from the `openmu-db` dev
volume) was run with 40 bot accounts x 5 characters = 200 bot characters, on an image
built from this branch plus two temporary log lines - **the instrumentation is not part
of either commit**.

| Test | Rates | Kill amount before the multiplier | credited / raw |
|---|---|---|---|
| A | money 1, exp 1 | ~350 | **1.0000** (5695 / 5695 payouts) |
| B | money 3, exp 1 | ~350, unchanged | **3.0000** (4547 / 4547 payouts) |
| C | money 1, exp 5 | ~6100, scales with experience | **1.0000** (4221 / 4221 payouts) |

Test B is the decisive one: the old code would have paid a party 9x. In every window
100 % of kills satisfied `amount == sum(experience) + 7` and the shares summed exactly
to the dropped amount.

Independent check over 964 party kills with unequal member experience: mean deviation
from the proportional-to-experience model **0.651 zen**, from the equal-split model
**8.706 zen**. A member with `exp=0` received `share=0`, where an equal split would have
paid it half.

The live run is also what surfaced the remainder bias fixed above: `total=52`,
`exp=[9,9,9,9,9]` produced `share=[12,10,10,10,10]` before the largest-remainder change.

### Measurement note

Tracking a bot's zen balance in the database does **not** measure income - bots spend
almost everything at merchants (one was observed dropping to 7 zen after a single
trip). The figures above come from per-payout log lines, not from balance deltas.

### DIFF
--- a/src/GameLogic/DroppedMoney.cs
+++ b/src/GameLogic/DroppedMoney.cs
@@ -148,7 +148,12 @@ public sealed class DroppedMoney : AsyncDisposable, ILocateable
             return true;
         }
 
-        var shares = this._shares.Count > 0
+        // Money has no owner - it can always be picked up, by strangers too. The recorded shares
+        // only apply when the party which picks it up is the one which earned it; then the money
+        // follows the experience. For anyone else there is no experience to follow, so it is split
+        // equally between the picking party, just like money without any shares (e.g. an item box).
+        var earnedByThisParty = this._shares.Any(share => party.IsEligibleForMoney(share.Player, player));
+        var shares = earnedByThisParty
             ? this._shares
             : MoneyDistribution.CreateEqualShares(this.Amount, party.PartyList.OfType<Player>().ToList());
 
--- a/src/GameLogic/DroppedMoney.cs
+++ b/src/GameLogic/DroppedMoney.cs
@@ -6,7 +6,6 @@ namespace MUnique.OpenMU.GameLogic;
 
 using System.Diagnostics;
 using System.Threading;
-using MUnique.OpenMU.GameLogic.Attributes;
 using MUnique.OpenMU.Pathfinding;
 using Nito.AsyncEx;
 
@@ -20,6 +19,8 @@ public sealed class DroppedMoney : AsyncDisposable, ILocateable
     /// </summary>
     private readonly AsyncLock _pickupLock;
 
+    private readonly IReadOnlyList<MoneyShare> _shares;
+
     private Timer? _removeTimer;
 
     private bool _availableToPick = true;
@@ -30,9 +31,14 @@ public sealed class DroppedMoney : AsyncDisposable, ILocateable
     /// <param name="amount">The amount.</param>
     /// <param name="position">The position where the item was dropped on the map.</param>
     /// <param name="map">The map.</param>
-    public DroppedMoney(uint amount, Point position, GameMap map)
+    /// <param name="shares">
+    /// The part of the money which is reserved for each player, matching the experience they gained from the kill.
+    /// When it's empty - for example for money from an item box - the money is split equally instead.
+    /// </param>
+    public DroppedMoney(uint amount, Point position, GameMap map, IReadOnlyList<MoneyShare>? shares = null)
     {
         this.Amount = amount;
+        this._shares = shares ?? [];
         this._pickupLock = new();
         this.Position = position;
         this.CurrentMap = map;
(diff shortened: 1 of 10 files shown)


nolt:
Removed — the reflection was unnecessary: the test helper already enters the player into the world, so party members share a live map and there is no longer any private field to set.

PR #854, opened 2026-07-27
Fix rollbacks caused by the periodic save racing player entity mutations

nolt:
## Summary

Players occasionally lost a whole session's progress **and items** on relog: they
would reconnect at an older position, missing the items and progress they had the
day before. The server log showed the periodic save failing with:

```
Error when saving the progress periodically for player 'Account:[...], Character:[...]'
DbUpdateConcurrencyException: expected to affect 1 row(s), but actually affected 0
```

Once a save starts failing, **every subsequent save for that session fails too**,
including the final save on disconnect — so nothing from that session reaches the
database and the player is rolled back to the last good save.

This affects any server with normal solo gameplay (crafting, gem stacking, selling
to NPCs); it does **not** require `/offlevel`, trade, or a personal store.

## Root cause

The EF change tracker is **not thread-safe**, and two independent flows touch the
same player context concurrently:

- **Packet handling** is sequential per connection (the read loop awaits each packet
  through the funnel in `RemotePlayer.PacketReceivedAsync`), and action handlers
  mutate tracked entities with plain field/collection writes — e.g. crafting toggling
  `item.ItemOptions`, or `RemoveItemAsync` + `PersistenceContext.DeleteAsync` when an
  item leaves the game. These raw mutations bypass the persistence context's own lock.
- **The periodic progress save** (`PeriodicSaveProgressPlugIn`) runs on a *separate*
  timer flow (`GameContext` periodic tasks) and calls `SaveChangesAsync`, whose
  `DetectChanges` enumerates the tracked navigation collections.

If the periodic save enumerates the change tracker while a handler mutates it, the
enumeration throws. Depending on exactly where change detection was, it surfaces as
`InvalidOperationException` ("Collection was modified"), a transiently-null internal
key (`ArgumentNullException`/`NullReferenceException`), an out-of-range index, or —
when a deferred `DeleteAsync` on an entity not yet persisted flushes — a
`DbUpdateConcurrencyException` "affected 0 rows". The change tracker is left corrupted,
so the whole session is lost.

A previous fix (#774) added the context `_lock` to `Attach`/`Detach`, which confirmed
the cause is concurrency but only covered those two methods — the raw entity mutations
still race the save. Crafting is the widest window (a multi-step create/delete with no
intermediate `SaveChanges`), which is why it was the most-reported trigger.

## Reproduction

A standalone harness against real EF Core + PostgreSQL, seeding the full
`VersionSeasonSix` game configuration and then racing two flows on **one** player
context:

- **RACING** (mutations ∥ save): `SaveChangesAsync` throws on the first attempt
  (`InvalidOperationException: Collection was modified` from `DetectChanges`
  enumerating `ItemOptions`).
- **SERIALIZED** (same operations, never overlapping): 512k+ saves, 0 errors.

The control proves the fix direction: serializing the mutations against the save
closes the window deterministically.

### Layer 1 — deterministic serialization (per-player re-entrant lock)

A per-player `AsyncLock` (`Player._persistenceLock`) serializes a player's context
mutations against its own progress saves. It is acquired around:

- **the packet handler funnel** (`RemotePlayer.PacketReceivedAsync`) — dispatch only,
  wrapping the existing `HandlePacketAsync` call; the packet format is untouched;
- **`SaveProgressAsync`** — the periodic and disconnect saves;
- **the offline bot tick** (`OfflinePlayerMuHelper.SafeTickAsync`) — bots run on the
  same periodic-task flow and are in the saved player list, so their loot/combat
  mutations were a major source of the race.

The lock is **re-entrant per asynchronous flow** via a per-instance `AsyncLocal<bool>`
flag, so an inline save inside an already-serialized handler (e.g. `DropItemAction`,
`TradeAcceptAction`) does not deadlock. The flag is a per-**instance** field on purpose:
re-entrancy is tracked per player, so a flow already holding player A's lock still
acquires player B's lock (e.g. during a trade) instead of wrongly skipping it.

### Layer 2 — best-effort bounded retry

`EntityFrameworkContextBase.SaveChangesAsync` retries a bounded number of times
(3, with a short backoff) when the save throws a member of the transient
tracker-corruption family (`DbUpdateConcurrencyException`, `InvalidOperationException`,
`ArgumentNullException`, `NullReferenceException`, `IndexOutOfRangeException`,
`KeyNotFoundException`). A racing mutation is a single quick operation, so a retry
lands on a stable moment. The retry is intentionally broad because a concurrent
mutation surfaces as several exception types, not just "collection modified"; a
genuinely persistent error of the same type is **not** masked — it rethrows once the
retries are exhausted.

Layer 1 is the primary guard; Layer 2 only needs to absorb the rare, bursty sources
that are not on a lock-held flow.

## Deliberate limits

- **Cross-player structural mutations** (e.g. an attacker destroying a victim's
  ammunition during combat, or a level-up applied to another party member from the
  killer's flow) are left to Layer 2. A deterministic fix would require a global
  lock order across players, which risks an AB-BA deadlock in PvP — worse than a rare,
  now-retried save. This is a conscious trade-off.
- **Own-player fire-and-forget combat** (`Task.Run` spawns inherit the re-entrancy
  flag through the execution context and therefore skip the lock) is likewise absorbed
  by Layer 2. Behaviour here is unchanged from before the fix — no regression.
- **Scalar writes** (experience, HP, position, money) are not serialized: they are
  torn reads at worst, never a tracker crash, and they are the hottest paths.
- A documented invariant on the guard method warns against acquiring a second player's
  persistence lock from inside a packet handler (the only future path that could open
  an AB-BA cycle).

## Validation

- Full test suite: **639/639** (3 new tests for the lock: serialization of 50
  concurrent callers, re-entrancy without deadlock, an independent flow waiting while
  the lock is held).
- Harness (real EF + PostgreSQL): SERIALIZED (lock) — 525k saves, 0 errors;
  BURSTY (rare race + retry) — 1.05M saves, **0 unrecovered**.

## No client impact

The `RemotePlayer` change wraps the dispatch call only; the client→server packet
contract is untouched, so no game-client change is required.

### DIFF
--- a/src/GameLogic/Offline/OfflinePlayerMuHelper.cs
+++ b/src/GameLogic/Offline/OfflinePlayerMuHelper.cs
@@ -141,7 +141,11 @@ public sealed class OfflinePlayerMuHelper : AsyncDisposable
     {
         try
         {
-            await this.TickAsync(cancellationToken).ConfigureAwait(false);
+            // Run the whole tick under the player's persistence lock so its structural mutations
+            // (loot pickup, combat ammo/pet destruction, queued equip/jewel actions) never overlap
+            // this bot's periodic progress save, which runs on a separate timer. The tick has no
+            // internal delays, so the lock is held only for its brief duration.
+            await this._player.RunPersistenceExclusiveAsync(() => this.TickAsync(cancellationToken)).ConfigureAwait(false);
             this._player.OnAiTickSucceeded();
         }
         catch (OperationCanceledException)
--- a/src/GameLogic/Player.cs
+++ b/src/GameLogic/Player.cs
@@ -54,6 +54,21 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
     private readonly AsyncLock _moveLock = new();
     private readonly AsyncLock _experienceLock = new();
 
+    /// <summary>
+    /// Serializes context mutations done by this player's action handlers against the periodic and
+    /// disconnect progress saves, which run on an independent timer flow. See
+    /// <see cref="RunPersistenceExclusiveAsync{T}"/>.
+    /// </summary>
+    private readonly AsyncLock _persistenceLock = new();
+
+    /// <summary>
+    /// Tracks, per asynchronous flow, whether <see cref="_persistenceLock"/> is already held, so the
+    /// lock can be re-entered (Nito's <see cref="AsyncLock"/> is not reentrant). It is an instance
+    /// field on purpose: reentrancy must be tracked per player, so a flow holding player A's lock
+    /// still acquires player B's lock (e.g. during a trade) instead of wrongly skipping it.
+    /// </summary>
+    private readonly AsyncLocal<bool> _persistenceLockHeld = new();
+
     private readonly Walker _walker;
 
     private readonly AppearanceDataAdapter _appearanceData;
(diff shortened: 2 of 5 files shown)


nolt:
Replaced the empty-condition for (;;) with a while (true) loop that increments the attempt counter in its body — behaviour is unchanged (the loop still exits via the success return or the rethrow once the bounded retries are exhausted).

Issue #721, opened 2026-03-24
Castle Siege Data Model & Configuration


sven-n:
## Summary

Create all data model classes, configuration types, entity types, and enumerations needed by Castle Siege. This phase establishes the persistent foundation that all subsequent phases build upon.

## Prerequisites

- #720 (Guild Alliance System) — for `GuildPosition.AssistantMaster` and alliance concepts.

## Background

OpenMU uses a `[Cloneable]` attribute (processed by a source generator) for configuration types that need deep-copy support, and `[AggregateRoot]` for top-level persistent entities. Configuration is nested under `GameConfiguration`. Entities are persisted via EF Core with auto-generated model classes through `Persistence.SourceGenerator`.

Key reference types:
- `DataModel/Configuration/GameConfiguration.cs` — root config with `[MemberOfAggregate]` collections.
- `DataModel/Configuration/MiniGameDefinition.cs` — example of a `[Cloneable]` event config.
- `DataModel/Entities/Guild.cs` — example of an `[AggregateRoot]` entity.
- `DataModel/Configuration/GameMapDefinition.cs` — map config example.

#### `CastleSiegeState.cs`
```csharp
public enum CastleSiegeState : byte
{
    Idle1 = 0,
    RegisterGuild = 1,
    Idle2 = 2,
    RegisterMark = 3,
    Idle3 = 4,
    Notify = 5,
    Ready = 6,
    Start = 7,
    End = 8,
    EndCycle = 9,
}
```

#### `CastleSiegeJoinSide.cs`
```csharp
public enum CastleSiegeJoinSide : byte
{
    None = 0,
    Defense = 1,
    Attack1 = 2,
    Attack2 = 3,
    Attack3 = 4,
}
```

#### `CastleSiegeUpgradeType.cs`
```csharp
public enum CastleSiegeUpgradeType : byte
{
    Defense = 1,
    Regen = 2,
    Life = 3,
}
```

### 2. Configuration Classes — **New files in `DataModel/Configuration/`**

All should have `[Cloneable]` and be `partial class`.

#### `CastleSiegeConfiguration.cs`

Main configuration aggregated under `GameConfiguration`:

| Property | Type | Description |
|----------|------|-------------|
| `Enabled` | `bool` | Master toggle |
| `CrownHoldTimeSeconds` | `int` | Seconds to hold the crown to capture (default 30) |
| `RegisterMinLevel` | `int` | Minimum guild master combined level to register (default 200) |
[...]

PR #860, opened 2026-08-01
Add Castle Siege persistence and Season 6 initialization

Zylkien:
## Summary

- configure the EF Core mappings for Castle Siege configuration and persistent entities
- add the follow-up migration required after the data model merged in #721
- initialize the Season 6 schedule, NPC definitions, upgrade tables, machine zones, respawn areas, and persistent state
- add an idempotent update plug-in for existing Season 6 databases
- prevent persistent Castle Siege state changes from being published as configuration changes
- add initialization, update, model-mapping, and round-trip regression coverage

## Testing

- `ConfigurationChangePublishingTests`
- `TestInitializationWithEfCore.TestSeason6DataAsync`

Both focused tests pass.

Closes #732
Part of #735

### DIFF
--- a/src/Persistence/EntityFramework/EntityDataContext.cs
+++ b/src/Persistence/EntityFramework/EntityDataContext.cs
@@ -21,6 +21,16 @@ public class EntityDataContext : ExtendedTypeContext
     /// </summary>
     internal GameConfiguration? CurrentGameConfiguration { get; set; }
 
+    /// <summary>
+    /// Gets the persistent Castle Siege state.
+    /// </summary>
+    internal DbSet<CastleSiegeData> CastleSiegeData => this.Set<CastleSiegeData>();
+
+    /// <summary>
+    /// Gets the Castle Siege guild registrations.
+    /// </summary>
+    internal DbSet<CastleSiegeGuildRegistration> CastleSiegeGuildRegistrations => this.Set<CastleSiegeGuildRegistration>();
+
     /// <inheritdoc/>
     protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
     {
@@ -58,6 +68,11 @@ public class EntityDataContext : ExtendedTypeContext
         modelBuilder.Entity<Account>().Apply();
         modelBuilder.Entity<Character>().Apply();
         modelBuilder.Entity<CharacterClass>().Apply();
+        modelBuilder.Entity<CastleSiegeConfiguration>().Apply();
+        modelBuilder.Entity<CastleSiegeData>().Apply();
+        modelBuilder.Entity<CastleSiegeGuildRegistration>().Apply();
+        modelBuilder.Entity<CastleSiegeNpcDefinition>().Apply();
+        modelBuilder.Entity<CastleSiegeNpcState>().Apply();
         modelBuilder.Entity<DropItemGroup>().Apply();
         modelBuilder.Entity<ExitGate>().Apply();
         modelBuilder.Entity<GameConfiguration>().Apply();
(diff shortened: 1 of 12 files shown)


sven-n:
## Code Review

Reviewed the hand-written changes (the ~5,800-line `*.Designer.cs` and model-snapshot files are EF-generated and were skipped). Overall this is a clean, well-structured PR: the model mappings, migration, initializer, idempotent update plug-in, and regression tests all hang together nicely, and the test coverage is genuinely thorough (schedule, NPCs, upgrades, zones, persistent round-trip, and update-plug-in idempotency).

### What the PR does
- Adds EF Core mappings for the Castle Siege config/persistent entities (`CastleSiegeExtensions`), wires them into `EntityDataContext`, and ships the follow-up migration after the #721 data model merge.
- Seeds the Season 6 Castle Siege configuration and persistent state (`CastleSiegeInitializer`) and hooks it into `GameConfigurationInitializer`.
- Adds an idempotent `AddCastleSiegeDataUpdatePlugIn` (v100) for existing DBs.
- Stops persistent (`data`-schema) entity changes from being broadcast as configuration changes via `EntityFrameworkContextBase.PublishesConfigurationChanges`.

### Potential issues

- **Duplicate defense machine zone (likely a bug).** In `CastleSiegeInitializer.InitializeMachineZones`, the last two `DefenseMachineZones` entries are identical:
  ```csharp
  configuration.DefenseMachineZones.Add(this.CreateZone(84, 52, 102, 66));
  configuration.DefenseMachineZones.Add(this.CreateZone(84, 52, 102, 66));
  ```
  The attack side has four distinct zones, so this looks like a copy/paste error where the 4th defense zone should have different coordinates. Note the test (`AssertZones(configuration.DefenseMachineZones, [... (84, 52, 102, 66), (84, 52, 102, 66)])`) currently encodes the duplicate, so it would need updating alongside the fix. Worth confirming against the intended map layout.

- **`config` change filter is a broad behavioral change.** `PublishesConfigurationChanges` now narrows `OnSavedChangesAsync` publishing to entities whose schema is `SchemaNames.Configuration`. This is the right fix for Castle Siege persistent state, but it changes publishing behavior for *every* context, not just this feature. Previously any changed non-`Unchanged` entity was published. Please confirm no other non-config entity type was relying on being published as a configuration change (the new `ConfigurationChangePublishingTests` only covers the two Castle Siege types).

- **Migration makes `MonsterDefinitionId` non-nullable with `defaultValue: Guid.Empty`.** Any pre-existing `CastleSiegeNpcDefinition` row with a NULL `MonsterDefinitionId` would be set to `Guid.Empty`, which then violates the re-added (Restrict) FK to `MonsterDefinition`. This is almost certainly a non-issue in practice since the table was introduced in #721 and is empty on real databases, but flagging in case any dev/test DB already has rows.

### Minor / nits
- `CastleSiegeInitializer.InitializeData` returns `CastleSiegeData`, but the caller in `AddCastleSiegeDataUpdatePlugIn.ApplyAsync` discards the return value. Fine as-is; just noting the return is only used by `Initialize()`.
- The persistent-vs-config split relies on `IsPersistedToDatabase`/`isPersisted` matching only monster numbers 277 (gate) and 283 (statue). The `CurrentHp` assignment (`== GateMonsterNumber ? gateHitPoints : statueHitPoints`) implicitly assumes every persisted NPC that isn't the gate is a statue — correct given current data, but slightly fragile if more persisted NPC types are added later.

### Verdict
Looks solid and mergeable once the duplicate defense zone is confirmed/corrected. The main thing to double-check is the intent behind that duplicate coordinate and the broadened config-change publishing filter.

---

PR #866, opened 2026-08-06
Rage Fighter Master Tree Finishup

ze-dom:
**To-do**
- [x] Comment changes with sources
- [x] UpdatePlugin

**Developments**
- Added master skill extra mana consumption depending on skill level
- Added the magic effects: _IncreaseBlockPowerUp_, _IncreaseBlockMastery_, _IncreaseHealthStrengthener_
- Added effects/target attributes for the RF master skills: _DefSuccessRateIncPowUp_, _EquippedWeaponMastery_, _DefSuccessRateIncMastery_, _StaminaIncreaseStrengthener_, _DecreaseMana_, 

**Bugfixes**
- On equipped weapon skills, no longer add the skill if a replacement master skill exists (existed previously, but only for RF weapons)
- Fixed values/added caps for RF skills _Increase Block_, _Increase Stamina_ and _Ignore Defense_


src/GameLogic/AttackableExtensions.cs
```diff
@@ -576,9 +576,17 @@ public static int GetRequiredValue(this IAttacker attacker, AttributeRequirement
...
 
         var extraCost = 0;
-        if (addExtraManaCost && attributeRequirement.Attribute == Stats.CurrentMana)
+        if (attributeRequirement.Attribute == Stats.CurrentMana)
         {
-            extraCost = (int)attacker.Attributes[Stats.SkillExtraManaCost];
+            if (skillEntry?.Level > 0)
+            {
+                extraCost += (int)(attributeRequirement.MinimumValue * MasterSkillExtensions.CalculateManaRateValue(skillEntry.Level));
+            }
```
ze-dom, line 584: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjUseSkill.cpp#L457), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjUseSkill.cpp#L428)

zTeamS6.3 does not have this, but the S6E3 client values also include this extra cost. Also, the [function](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L4813) for it exists. E.g. (base mana requirement = 55):
Image

src/GameLogic/MasterSkillExtensions.cs
```diff
@@ -12,6 +12,8 @@ namespace MUnique.OpenMU.GameLogic;
 /// </summary>
 public static class MasterSkillExtensions
 {
+    private const string FormulaManaRateIncrease = "(2 + (((((((level - 30) ^ 3) + 25000) / 499) / 6)))) * 1.5";
+
```
ze-dom, line 16: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/MasterSkillEffect.lua#L131)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -117,8 +117,10 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
         attributeRelationships.Add(this.CreateAttributeRelationship(Stats.MaximumPhysBaseDmg, 1, Stats.PhysicalBaseDmg));
         attributeRelationships.Add(this.CreateAttributeRelationship(Stats.MinimumPhysBaseDmg, 1, Stats.PhysicalBaseDmgIncrease, aggregateType: AggregateType.Multiplicate));
         attributeRelationships.Add(this.CreateAttributeRelationship(Stats.MaximumPhysBaseDmg, 1, Stats.PhysicalBaseDmgIncrease, aggregateType: AggregateType.Multiplicate));
+
         attributeRelationships.Add(this.CreateAttributeRelationship(Stats.MaximumHealth, 1, Stats.SwellLifeHealthIncrease, aggregateType: AggregateType.Multiplicate));
         attributeRelationships.Add(this.CreateAttributeRelationship(Stats.MaximumMana, 1, Stats.SwellLifeManaIncrease, aggregateType: AggregateType.Multiplicate));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.IncreaseBlockBonus, 1, Stats.DefenseRatePvm));
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L5887), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L6021)
Though this only comes online for _DefSuccessRateIncPowUp_ and _DefSuccessRateIncMastery_. That's why we start with a 0 value `AggregateType.Multiplicate` on [IncreaseBlockEffectInitializer.cs](https://github.com/MUnique/OpenMU/pull/866/changes#diff-7fed80f6159e570dc5d5498d3d644e9fa7d5489dd68304415e8b9fd461daf33b)

src/Persistence/Initialization/CharacterClasses/ClassFairyElf.cs
```diff
@@ -83,6 +83,7 @@ protected CharacterClass CreateFairyElf(CharacterClassNumber number, string name
         result.AttributeCombinations.Add(this.CreateAttributeRelationship(Stats.ArcheryAttackMode, 1, Stats.IsBowEquipped));
         result.AttributeCombinations.Add(this.CreateAttributeRelationship(Stats.ArcheryAttackMode, 1, Stats.IsCrossBowEquipped));
         result.AttributeCombinations.Add(this.CreateAttributeRelationship(Stats.MeleeAttackMode, 0, Stats.ArcheryAttackMode, InputOperator.ExponentiateByAttribute));
+        result.AttributeCombinations.Add(this.CreateAttributeRelationship(Stats.SkillExtraManaCost, 0, Stats.AmmunitionConsumptionRate, InputOperator.ExponentiateByAttribute, AggregateType.Multiplicate));
```
ze-dom, line 86: If _InfiniteArrow_ effect is active (`Stats.AmmunitionConsumptionRate` = 0), so will `Stats.SkillExtraManaCost`. Otherwise, it's 0.

src/GameLogic/Player.cs
```diff
@@ -1548,28 +1548,31 @@ public ValueTask NewLocateablesInScopeAsync(IEnumerable<ILocateable> newObjects)
...
+            return false;
+        }
+
         if (skill.Requirements.Any(r => r.MinimumValue > this.Attributes![r.Attribute]))
         {
             return false;
         }
 
-        var addExtraManaCost = this.Attributes![Stats.AmmunitionConsumptionRate] == 0
-            && skill.SkillType is SkillType.DirectHit or SkillType.AreaSkillAutomaticHits;
```
ze-dom, line 1574: Replaced this with skill attributes for the elf's shooting skills on [SkillsInitializer.cs](https://github.com/MUnique/OpenMU/pull/866/changes#diff-0d0ee717baf69d9690226f905c34e1dc8244bf695b5523913b41faa4ddf77894).

src/Persistence/Initialization/Skills/IgnoreDefenseEffectInitializer.cs
ze-dom, line 1: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/BuffSkillEffect.xml#L185)

src/Persistence/Initialization/Skills/IncreaseBlockEffectInitializer.cs
ze-dom, line 1: BuffSkillEffect.xml: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/BuffSkillEffect.xml#L196)

ObjUseSkill.cpp: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjUseSkill.cpp#L6420), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjUseSkill.cpp#L6538)
BuffEffect.cpp: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/BuffEffect.cpp#L145), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/BuffEffect.cpp#L142)

The client actually applies it as a [multiplier](https://github.com/sven-n/MuMain/blob/1a678b3089b725f40e5d924d3304a3eba82f4964/src/source/UI/NewUI/Character/NewUICharacterInfoWindow.cpp#L975) (verified on S6E3 official client). My guess is that it is a bug and the server sources are correct. Otherwise at max value (~1k energy) this would double the defense rate of the buff targets, which seems too much to me.

Image

src/Persistence/Initialization/Skills/IncreaseBlockMasteryEffectInitializer.cs
ze-dom, line 1: BuffSkillEffect.xml: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/BuffSkillEffect.xml#L198)

MasterLevelSkillTreeSystem.cpp: [zTeamS6.3,](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L5994), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L6117)



Image

src/Persistence/Initialization/Skills/IncreaseBlockPowerUpEffectInitializer.cs
ze-dom, line 1: BuffSkillEffect.xml: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/BuffSkillEffect.xml#L197)

MasterLevelSkillTreeSystem.cpp: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L5887), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L6021)

Image

src/Persistence/Initialization/Skills/IncreaseHealthEffectInitializer.cs
ze-dom, line 1: BuffSkillEffect.xml: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/BuffSkillEffect.xml#L194)

ObjUseSkill.cpp: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjUseSkill.cpp#L6414), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjUseSkill.cpp#L6532)

src/Persistence/Initialization/Skills/IncreaseHealthStrengthenerEffectInitializer.cs
ze-dom, line 1: BuffSkillEffect.xml: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Skill/BuffSkillEffect.xml#L195)

MasterLevelSkillTreeSystem.cpp: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L5881), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L6015)

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -195,7 +198,7 @@ public override void Initialize()
         this.CreateSkill(SkillNumber.InfinityArrow, "Infinity Arrow", CharacterClasses.MuseElfAndHighElf, distance: 6, abilityConsumption: 10, manaConsumption: 50, levelRequirement: 220, skillType: SkillType.Buff, targetRestriction: SkillTargetRestriction.Self);
         this.CreateSkill(SkillNumber.FireScream, "Fire Scream", CharacterClasses.AllLords, DamageType.Physical, 130, 6, 10, 45, energyRequirement: 70, leadershipRequirement: 150, skillType: SkillType.AreaSkillAutomaticHits);
         this.AddAreaSkillSettings(SkillNumber.FireScream, true, 2f, 3f, 6f);
-        this.CreateSkill(SkillNumber.Explosion79, "Explosion", CharacterClasses.AllLords, DamageType.Physical, distance: 2);
+        this.CreateSkill(SkillNumber.Explosion79, "Explosion", CharacterClasses.None, DamageType.Physical, distance: 2);
```
ze-dom, line 201: This is the _FireScream_'s explosion effect. It's not a castable skill. Everything was working previously, I just found out that `CharacterClasses.None` existed 😁 , so using it to simplify.

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -232,7 +235,7 @@ public override void Initialize()
         this.CreateSkill(SkillNumber.GiganticStorm, "Gigantic Storm", CharacterClasses.AllMGs, DamageType.Wizardry, 110, 6, 10, 120, 220, 118, elementalModifier: ElementalType.Wind, skillType: SkillType.AreaSkillAutomaticHits);
         this.CreateSkill(SkillNumber.ChaoticDiseier, "Chaotic Diseier", CharacterClasses.AllLords, DamageType.Physical, 190, 6, 15, 50, 100, 16, skillType: SkillType.AreaSkillAutomaticHits);
         this.AddAreaSkillSettings(SkillNumber.ChaoticDiseier, true, 1.5f, 1.5f, 6f, minimumHitsPerAttack: 7);
-        this.CreateSkill(SkillNumber.DoppelgangerSelfExplosion, "Doppelganger Self Explosion", CharacterClasses.AllMGs, DamageType.Wizardry, 140, 3, 25, 20, 100, elementalModifier: ElementalType.Fire);
+        this.CreateSkill(SkillNumber.DoppelgangerSelfExplosion, "Doppelganger Self Explosion", CharacterClasses.None, DamageType.Wizardry, 140, 3, 25, 20, 100, elementalModifier: ElementalType.Fire);
```
ze-dom, line 238: This is the [Doppelganger](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/Monster/Monster.xml#L485) mob's attack (`attacktype="239"`).
Not a castable skill:
Image

src/Persistence/Initialization/Skills/MagicEffectNumber.cs
```diff
@@ -342,6 +342,21 @@ internal enum MagicEffectNumber : short
...
+    /// <summary>
+    /// The increase block mastery effect of the rage fighter.
+    /// </summary>
+    IncreaseBlockMastery = 154,
+
+    /// <summary>
+    /// The increase health (stamina) strengthener effect of the rage fighter.
+    /// </summary>
+    IncreaseHealthStrengthener = 155,
+
```
ze-dom, line 359: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/BuffEffect_.h#L156)

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -929,13 +940,11 @@ private void InitializeMasterSkillData()
...
-
-        // todo: Increases the probability of Double Damage while using gloves according to the assigned Skill Level.
-        this.AddMasterSkillDefinition(SkillNumber.EquippedWeaponMastery, SkillNumber.EquippedWeaponStrengthener, SkillNumber.Undefined, 3, 3, SkillNumber.Undefined, 20, Formula120);
-        this.AddMasterSkillDefinition(SkillNumber.DefSuccessRateIncMastery, SkillNumber.DefSuccessRateIncPowUp, SkillNumber.Undefined, 3, 3, SkillNumber.IncreaseBlock, 20, Formula502);
-        this.AddMasterSkillDefinition(SkillNumber.StaminaIncreaseStrengthener, SkillNumber.IncreaseHealth, SkillNumber.Undefined, 3, 4, SkillNumber.IncreaseHealth, 20, Formula1154);
-        this.AddMasterSkillDefinition(SkillNumber.DecreaseMana, SkillNumber.Undefined, SkillNumber.Undefined, 3, 4, SkillNumber.Undefined, 20, Formula722);
+        this.AddMasterSkillDefinition(SkillNumber.DefSuccessRateIncPowUp, SkillNumber.IncreaseBlock, SkillNumber.Undefined, 3, 2, SkillNumber.IncreaseBlock, 20, $"{Formula502} / 100", Formula502, Stats.IncreaseBlockBonus, AggregateType.Multiplicate);
```
ze-dom: Image

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -929,13 +940,11 @@ private void InitializeMasterSkillData()
...
+        this.AddPassiveMasterSkillDefinition(SkillNumber.EquippedWeaponMastery, Stats.GloveWeaponMasteryDoubleDamageChance, AggregateType.AddRaw, Formula120Value, Formula120, 3, 3, SkillNumber.EquippedWeaponStrengthener);
```
ze-dom, line 944: Image

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -929,13 +940,11 @@ private void InitializeMasterSkillData()
...
+        this.AddMasterSkillDefinition(SkillNumber.DefSuccessRateIncMastery, SkillNumber.DefSuccessRateIncPowUp, SkillNumber.Undefined, 3, 3, SkillNumber.DefSuccessRateIncPowUp, 20, Formula502, Formula502, Stats.DefenseFinal, AggregateType.AddFinal);
```
ze-dom, line 945: Tooltip innacurate/misleading - it adds defense.

Image

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -929,13 +940,11 @@ private void InitializeMasterSkillData()
...
+        this.AddMasterSkillDefinition(SkillNumber.StaminaIncreaseStrengthener, SkillNumber.IncreaseHealth, SkillNumber.Undefined, 3, 4, SkillNumber.IncreaseHealth, 20, Formula1154, Formula1154, Stats.TotalVitality, AggregateType.AddFinal);
```
ze-dom, line 946: Image

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -929,13 +940,11 @@ private void InitializeMasterSkillData()
...
+        this.AddPassiveMasterSkillDefinition(SkillNumber.DecreaseMana, Stats.ManaUsageReduction, AggregateType.AddRaw, Formula722Value, Formula722, 4, 3);
```
ze-dom, line 947: Image

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -929,13 +940,11 @@ private void InitializeMasterSkillData()
...
-
-        // todo: Increases the probability of Double Damage while using gloves according to the assigned Skill Level.
-        this.AddMasterSkillDefinition(SkillNumber.EquippedWeaponMastery, SkillNumber.EquippedWeaponStrengthener, SkillNumber.Undefined, 3, 3, SkillNumber.Undefined, 20, Formula120);
-        this.AddMasterSkillDefinition(SkillNumber.DefSuccessRateIncMastery, SkillNumber.DefSuccessRateIncPowUp, SkillNumber.Undefined, 3, 3, SkillNumber.IncreaseBlock, 20, Formula502);
-        this.AddMasterSkillDefinition(SkillNumber.StaminaIncreaseStrengthener, SkillNumber.IncreaseHealth, SkillNumber.Undefined, 3, 4, SkillNumber.IncreaseHealth, 20, Formula1154);
-        this.AddMasterSkillDefinition(SkillNumber.DecreaseMana, SkillNumber.Undefined, SkillNumber.Undefined, 3, 4, SkillNumber.Undefined, 20, Formula722);
+        this.AddMasterSkillDefinition(SkillNumber.DefSuccessRateIncPowUp, SkillNumber.IncreaseBlock, SkillNumber.Undefined, 3, 2, SkillNumber.IncreaseBlock, 20, $"{Formula502} / 100", Formula502, Stats.IncreaseBlockBonus, AggregateType.AddRaw);
```
ze-dom, line 943: Image

sven-n:
## Review

Went through the whole diff against `master` and traced the new attributes/effects through the runtime paths that consume them. Overall this is well put together — the attribute modelling is sound and the update plug-in is unusually thorough. A few things below, roughly in order of importance.

### 1. Accidentally deleted XML doc tag

`src/Persistence/Initialization/Updates/UpdateVersion.cs` — the opening `/// <summary>` of `ConfigureCastleSiegeRegistration` got removed along with the preceding blank line:

```csharp
    FinishSummonerMasterTree = 101,

    /// The version of the <see cref="ConfigureCastleSiegeRegistrationUpdatePlugIn"/>.
    /// </summary>
    ConfigureCastleSiegeRegistration = 102,
```

Looks unintentional — the new `FinishRageFighterMasterTree` entry below it is fine.

### 2. `IncreaseBlockBonus` is missing at two other `DefenseRatePvm` read sites

`GetHitChanceTo` now reads `DefenseRatePvm + IncreaseBlockBonus`, but two other places still read the raw value:

* `AttackableExtensions.Overrates` (`AttackableExtensions.cs:718`), which drives the `dmg * 0.3` PvM reduction at line 205. So while the buff is up a Rage Fighter dodges more but doesn't cross the overrate threshold any earlier — the two halves of the same defense-rate concept disagree.
* `Offline/CombatHandler.GetExpectedHitShare` (`CombatHandler.cs:372`). Its own doc comment says it mirrors the hit chance in `AttackableExtensions`, and this PR already touches that file for the mana consumption, so it's worth keeping in sync.

A tiny extension used by all three would keep them from drifting again:

```csharp
public static float GetDefenseRatePvm(this IAttributeSystem attributes)
    => attributes[Stats.DefenseRatePvm] + attributes[Stats.IncreaseBlockBonus];
```

(Separately: since the bonus never lands on `DefenseRatePvm` itself, the client's displayed defense rate won't move while the buff is active. That's forced by the `Multiplicate`-on-`DefenseRatePvm` design — just flagging it as a known consequence.)

### 3. `NovaSkillStartPlugin` — inlined consumption drops a check

```csharp
if (novaStart.ConsumeRequirements.Any(r => player.GetRequiredValue(r) > player.Attributes[r.Attribute]))
```

`TryConsumeForSkillAsync` also validates `skill.Requirements` before consuming; the inlined version only looks at `ConsumeRequirements`, so the nova start's requirements are no longer enforced. The skill entry is available here (`player.SkillList?.GetSkill(skillId)`), so the original single call site can be kept:

```csharp
var novaStartEntry = player.SkillList?.GetSkill(skillId);
if (novaStartEntry is null || !await player.TryConsumeForSkillAsync(novaStartEntry).ConfigureAwait(false))
{
    return;
}
```

That also avoids duplicating the consume loop and keeps the new master-skill mana rate applying uniformly.

### 4. The magic-effect deep-copy is now duplicated ten times

`IncreaseBlockPowerUpEffectInitializer`, `IncreaseBlockMasteryEffectInitializer` and `IncreaseHealthStrengthenerEffectInitializer` each carry ~45 lines of "copy `MagicEffectDefinition` from another one", and `FinishRageFighterMasterTreePlugIn` carries the same three bodies again with `this.Context` → `context`. Together with the four that already exist (`WizardryEnhanceStrengthener`/`WizardryEnhanceMastery`, `CriticalDamageIncreaseMastery`, `LifeSwellProficiency`) that's ten copies of the same block.

Worth extracting one helper, e.g.

```csharp
internal static MagicEffectDefinition CopyEffect(this IContext context, GameConfiguration config, MagicEffectDefinition source, MagicEffectNumber number, string name)
```

and having both the initializers and the update plug-in call it. Concretely it would already have prevented one divergence: the pre-existing `WizardryEnhanceStrengthenerEffectInitializer` copies `boostRelatedValueCopy.AggregateType = boostRelatedValue.AggregateType;`, and all six new copies omit that line. It's harmless today (the sources use the `AddRaw` default), but it's exactly the kind of drift a shared helper removes. Conversely the new copies *do* handle `Duration.RelatedValues`, which the older ones skip — so a merged helper would be strictly better than either.

### 5. `SkillListViewPlugIn._weaponSkillReplacements` restates configuration data

The 16 new constants and the map encode relationships the configuration already has — I checked all eight entries and each pair is exactly a `MasterDefinition.ReplacedSkill` link (`FallingSlashStreng`→`FallingSlash`, `CycloneStrengthener`/`CycloneStrengthenerDuelMaster`→`Cyclone`, `TripleShotMastery`→`TripleShotStrengthener`→`TripleShot`, …). `BuildSkillList` already derives this generically from `MasterDefinition.ReplacedSkill`; `AddSkillAsync` could do the same:

```csharp
if (this.SkillList.Any(s => s is not null && s.GetBaseSkills().Contains(skill)))
{
    return;
}
```

That covers every class without a hand-maintained list, including future master skills. If you'd rather keep the explicit map, please make it `private static readonly` — right now it's allocated per player view plug-in instance.

### 6. Questions

* **`Explosion79` / `DoppelgangerSelfExplosion` → `CharacterClasses.None`.** The explicit `Explosion79` filter in `BuildSkillList` was dropped in favour of `QualifiedCharacters`. `AddItemSkillAsync` does honour `QualifiedCharacters`, but skills already sitting in a character's `LearnedSkills` are not filtered by it, and the update plug-in only touches the configuration. Do you know whether any existing characters can have these two in `LearnedSkills`? If so they'd start showing up in the skill list after the update.
* **`SkillExtraManaCost` re-modelling.** The old rule was `AmmunitionConsumptionRate == 0 && SkillType is DirectHit or AreaSkillAutomaticHits`. The new one is the `0 ^ AmmunitionConsumptionRate` multiplicate combination on the three elf classes plus mirror relationships on five named skills. Net effect: a non-elf who equips a bow no longer pays the extra mana at all — neither the zeroing combination nor the mirror relationships apply to them. Intended? (Master variants are fine — `EnsureSkillAttributes` walks `GetBaseSkills()`.)
* **`IgnoreDefense` didn't get a `SubType`** while `IncreaseBlock` (74) and `IncreaseHealth` (73) did. Deliberate, or an oversight in the same family of buffs?
* **`IgnoreDefense` constant `-0.0104f`** with the per-energy term means the chance is negative below ~104 energy. Harmless given the 404-energy parchment, but if that buff can ever land on a low-energy character it silently becomes "never" rather than "small". Worth a floor at 0?
* **Class numbers in the update plug-in.** `charClass.Number == 8 || charClass.Number == 10 || ...` — elsewhere in `Updates/` the convention is `(CharacterClassNumber)c.Number is CharacterClassNumber.FairyElf or CharacterClassNumber.MuseElf or ...` (see `FixAttackSpeedCalculationUpdate.cs:177`). Same for the `24`/`25` check.

### What I verified as correct

Noting these so you know they were checked rather than skipped:

* The `CreateConstValueAttribute(0, Stats.IncreaseBlockBonus)` base value is genuinely required, not defensive: `ComposableAttribute` forces `rawValues = 1` when *all* elements are `Multiplicate`, so without it `IncreaseBlockBonus` would equal `DefenseRatePvm` outright. The inline comment is accurate.
* The `0f` placeholder power-ups (`IncreaseBlockBonus`/`AddRaw`, `DefenseFinal`/`AddFinal`) line up with the `TargetAttribute == powerUpDef.TargetAttribute && Aggregation == powerUp.AggregateType` match in `MagicEffectPowerUpExtensions.AppedMasterSkillPowerUp`, so the master values do reach the buff.
* `EquippedWeaponMastery` and `DecreaseMana` are `SkillType.PassiveBoost`, so `SkillList.CreatePowerUpWrappers` picks up their newly assigned `TargetAttribute`s.
* `MapSkillsToEffects()` runs before `InitializeMasterSkillData()`, so the three new `_effectsOfSkills` entries win over `skill.MagicEffectDef ??= replacedSkill.MagicEffectDef`.
* Root/rank survive the `AddMasterSkillDefinition` → `AddPassiveMasterSkillDefinition` switch even though the two overloads take those two bytes in opposite order — `DecreaseMana` stays root 3 / rank 4, `EquippedWeaponMastery` root 3 / rank 3.
* Update plug-in vs. fresh initialization agree on every touched master skill, including the display formulas the plug-in doesn't explicitly set (`Formula120Value` ≡ `Formula120 / 100` and `-1 * Formula120Value` ≡ `Formula120 / -100`, so those two are value-preserving rewrites).
* `ignoreDefenseEffect.Duration?.MaximumValue = 180f;` — null-conditional assignment is valid C# 14, fine on `net10.0`.

---

ze-dom:
Yeah, Claude appears to have caught some discrepancies. Will fix them later.

ze-dom:
All set!

PR #867, opened 2026-08-07
feat: support global coordinates for maps larger than 256x256

mrallanodreman:
## Summary
Adds a backward-compatible global-coordinate path for maps larger than 256x256.

- Introduces ushort-based global Point/terrain/pathfinding and EF persistence migration.
- Adds global movement, map-change, respawn, scope, NPC and character-information packets while retaining legacy packets for coordinates <= 255.
- Adds generated packet bindings, boundary tests, coordinate tests and protocol documentation.

## Verification
- Network packet project: 0 errors, 0 warnings.
- GameServer build: 0 errors; existing analyzer/style warnings remain.
- GlobalPacketHeaderTests: 1/1.
- WorldCoordinatesTest: 4/4.
- Runtime fork verification: World1 512x512 movement/render across 255 -> 256.

The feature is intentionally isolated from the companion implementation in the fork.


sven-n:
Thanks for the effort that clearly went into this — the write-up in `docs/GlobalCoordinates.md` is thorough and the proposed upstream split at the bottom of it is exactly the right instinct. I'm not going to merge this in its current form, though, and I'd like to explain why in enough detail to be useful.

## The main reason: this is a protocol dialect, not a feature

No stock MU client can use any of this. The global packets are only decodable by your own client, which lives outside this repository. Merging the server half would make OpenMU the maintainer of a wire protocol that exactly one out-of-tree client speaks, including permanent opcode allocations (`C1 D5`, `C1 D6`, `C3 1C 10`, `C3 F3 61`) that a real MU version may want later. Meanwhile every existing user pays for it: a mandatory schema migration, a wider `Point`, and ~11k lines of surface area, for a capability none of them can reach.

That's a decision I'd want to make in a design issue before reviewing any code, so please don't read the list below as "fix these and it's in".

## Blocking defects

These would break existing setups today, independent of the design question. CI hasn't run on the branch, which is presumably why they're unnoticed.

1. **Two shipped maps fail to load.** `GameMapTerrain.ReadTerrainData` now derives the map size from `sqrt(payload)` and throws when it isn't a perfect square (`GameMapTerrain.cs:180`). `Resources/Terrain81.att` and `Resources/Terrain82.att` are 65540 bytes, so the payload after `AsSpan(3)` is 65537 — not a perfect square. The other 74 `.att` files are 65539 and are fine. On `master` the stray byte is harmlessly folded into `[0,0]`.

2. **`new GameMapTerrain(null)` throws.** `DefaultTerrain` is still `short.MaxValue` = 32767 bytes (`GameMapTerrain.cs:18`) and 181² = 32761, so the no-terrain-data path now raises `ArgumentException`. That path is reachable from the map editor and `ExitGatePicker`.

3. **The existing test suite fails.** Six helpers set `TerrainData = new byte[ushort.MaxValue + 3]`, giving a 65535-byte payload, which is also not a perfect square: `GameContextTestHelper`, `PlayerTestHelper`, `PartyTest`, `MoveItemActionTests`, `SpeedHackAntiCheatTests`, and the benchmark helper. None of those files are touched by this PR.

The padding workaround at `GameMapTerrain.cs:172` special-cases payloads of exactly `65534` and `262142`, which matches none of the above. More generally, silently treating an unexpected blob length as "pad the tail with walkable terrain" isn't something I want in the terrain loader — an unexpected length should be a loud, specific error, not a guess.

4. **Anti-cheat tolerance is loosened without mention.** `Player.cs:1519` changes the `MaxAllowedWalkStartOffset` fallback from `5` to `12`. That widens the speed-hack tolerance for every server by 140%, and it's unrelated to coordinate width. This needs to be reverted regardless of what happens to the rest.

## Design concerns

5. **The new scope packets are ambiguous on the wire.** `AddCharacterToScopeGlobal` is declared as `C2HeaderWithSubCode` / code `12` / subcode `D6`, but the legacy `AddCharactersToScope` on code `12` has no subcode — byte index 4 is `CharacterCount`. The subcode therefore lands exactly on the count byte, so a standard client decodes the global packet as 214 characters of garbage. `AddNpcsToScopeGlobal` (`13` / `D5`) has the same problem with `NpcCount`. This isn't backward compatible in the protocol sense; it's only non-fatal because the server never emits the variant below 256 coordinates, and a client has no way to tell the two apart. It's also singular where the legacy packet batches, which costs one packet per character in scope.

6. **Unconverted paths now throw instead of degrading.** The `checked((byte)position.X)` casts added across the view plug-ins turn every not-yet-migrated path into an `OverflowException` at runtime on a large map — fireworks, party list, drops, skills, item drop effects. `docs/GlobalCoordinates.md` already acknowledges that drops, party cards, skills and companion packets still need an audit, so the feature is documented as incomplete and the incomplete parts fail hard rather than falling back.

7. **The admin panel wasn't included.** `Web/Shared/GameMapTerrainExtensions.ToImage` still hardcodes `0x100` (`:24-27`), so a 512×512 map renders only its top-left quadrant, and `MapCoordinateService` / `MapDragState` still carry `const int MapSize = 256` with `byte` casts.

## Scope

8. A large refactor of the warp lifecycle rides along and isn't described in the PR: `WarpToAsync` becomes `ValueTask<WarpResult>` with a `costs` parameter, two new locks and a hand-rolled rollback path; `PlaceAtGateAsync` now places the summon on the player's exact tile instead of a random gate point (`Player.cs:2366`); and gate validation now rejects gates that aren't fully in bounds or lack a walkable tile, which will break configurations that work today. Each of those is its own review.

9. `GlobalPoint`, `ChunkId` and `LocalPoint` are added as three new files with no consumers anywhere in the tree.

10. Smaller things: the docs list `WalkRequestGlobal` as `C2:D5` while the XML declares `C1Header` / `D5`; the new packet definitions in the XML use a single-line field style that doesn't match the rest of the file; the branch currently conflicts with `master`.

## What I'd be happy to take

Two pieces stand on their own merit and would be welcome as small, separate PRs:

- The re-entrancy guard in `ClientReadyAfterMapChangeAsync`. On `master` a repeated `F3:12` does look like it can re-add the player and summon to the map — if you can confirm that, it's a genuine bug fix and I'd like it independently of everything else.
- The bounds check before `Terrain.WalkMap[target.X, target.Y]` in `WalkToAsync`.

If you want to pursue large maps upstream, please open an issue describing the goal and the client story first, and let's agree on that before any of the code. If the answer there is yes, your own proposed step 1 (data model plus migration, with robust terrain-size handling and no behaviour change for 256×256 maps) would be the reviewable starting point.

I'm closing this one for now — not because the work isn't substantial, but because it can't be reviewed or maintained as a single unit.

PR #863, opened 2026-08-03
Define and validate Castle Siege network packets

Zylkien:
## Summary

- Add the complete Castle Siege server-to-client packet definitions.
- Validate the existing client-to-server packets against the current MuMain client.
- Correct the Castle Gate operation field semantics from `CloseState` to `IsOpen`.
- Define the Alliance removal guild name as a fixed 8-byte string.
- Regenerate the packet structs, reference structs, connection extensions, tests, indexes, and packet documentation.
- Add protocol conformance tests covering packet headers, codes, lengths, offsets, endianness, fixed strings, and variable-length structures.

## Protocol validation

The packet layouts were checked against the current MuMain client structures, dispatch logic, and generated sending bindings:

- [WSclient.h](https://github.com/sven-n/MuMain/blob/main/src/source/Network/Server/WSclient.h)
- [WSclient.cpp](https://github.com/sven-n/MuMain/blob/main/src/source/Network/Server/WSclient.cpp)

Where the draft issue description and the current client differ, the definitions follow the actual MuMain wire layout. Notable examples include:

- `B2:02` contains the Castle Siege give-up state.
- `B2:07` includes the requested upgrade value.
- `B9:05` sends the hunting-zone entrance money.
- NPC repair responses contain current and maximum HP.
- Minimap player and NPC positions use the `C2-B6` and `C2-BB` packets.
- Life Stone build progress uses `B9:01` with the NPC index and build time.

The Guild Alliance definitions already present on `master` were also validated. The Alliance removal request was corrected so `GuildName` is explicitly defined as the fixed 8-byte field consumed by MuMain.

## Testing

- Packet generator completed successfully.
- Packet test project builds successfully.
- 576 packet tests pass.
- Generated output was reviewed.
- `git diff --check` passes.
- No project-specific code or configuration is included.

Closes #731

PR #868, opened 2026-08-07
Implement Castle Siege guild selection and participant tracking

### LINKED ISSUE #724: Castle Siege Guild Selection & Participant Tracking
## Summary

Implement the guild selection algorithm (run during the `Notify` state) that ranks registered guilds and assigns attacking sides, and the participant tracking system that records battle participation for rewards.

## Prerequisites

- #722 (State Machine) — for `CastleSiegeContext` and `Notify` state enter handler.
- #723 (Registration) — for `CastleSiegeGuildRegistration` data.
- #720 (Guild Alliance) — for `IGuildServer.GetAllianceGuildsAsync`.

### 1. Guild Selection Algorithm — **New file: `GameLogic/CastleSiege/CastleSiegeGuildSelector.cs`**

Called from `CastleSiegePlugIn.OnEnterStateAsync(Notify)`.

**Steps:**
1. Query all registered guilds from `CastleSiegeContext.RegisteredGuilds`.
2. Filter: `Marks > 0 AND memberCount > 0`.
3. For each guild, compute:
   - `memberCount` = total guild members (from `IGuildServer.GetGuildListAsync`).
   - `combinedLevel` = guild master's level + master level (may need to be stored at registration time or queried).
   - `score = (Marks * 5) + memberCount + (combinedLevel / 4)`.
4. Sort descending by `(score, registrationOrder)` — higher score first; earlier registration wins ties.
   - The composite sort uses `totalScore = ((long)(uint)transformedRegisterId) | (((long)(uint)transformedScore) << 32)` where `transformedRegisterId = 0x100000 - registrationOrder` (earlier = higher).
5. Top N guilds (from `Configuration.MaxAttackingGuilds`, default 3) become `Attack1`, `Attack2`, `Attack3`.
6. If the castle is occupied (`SiegeData.IsOccupied`), the owner guild becomes `Defense`.
7. For each selected guild, query all alliance member guilds via `IGuildServer.GetAllianceGuildsAsync` and add them all to `FinalGuildList` with the same side.
8. Save the final guild list to the database (clear and re-insert `castle_siege_guild` equivalent).

### 2. Join Side Assignment — method on `CastleSiegeContext`

`SetPlayerJoinSideAsync(bool killLifeStones)`:

1. Iterate all online players in the Castle Siege map (obtained from the `GameMap`).
2. For each player:
   - If not in a guild → `CastleSiegeJoinSide.None`.
   - Otherwise, look up guild in `FinalGuildList` (by runtime uint guild ID).
   - Assign `CastleSiegeJoinSide` and store on a per-player tracking mechanism.
   - Send `CastleSiegeJoinSideNotification` packet.
   - Apply the corresponding magic effect/buff (`BUFF_CASTLE_SIEGE_DEFENSE`, `BUFF_CASTLE_SIEGE_ATTACK_1`, etc.).
3. If `killLifeStones` is true, destroy all active Life Stones (Phase 9).

**Called from:**
- `Ready` state enter (initial assignment).
- Every 5 seconds during `Start` state (for newly joined players).
- On castle capture / side swap.

### 3. Player Join Side Storage

Need a way to track each player's current join side. Options:
- Add a `CastleSiegeJoinSide` property to `Player` (simplest, though couples the concept).
- Use a `ConcurrentDictionary<Player, CastleSiegeJoinSide>` on `CastleSiegeContext`.

Recommendation: Use the context-based dictionary, queried via `CastleSiegeContext.GetPlayerJoinSide(Player)`.

### 4. Participant Tracking — **New file: `GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs`**

During `Start` state, called every 5 seconds:

1. Iterate all characters in the CS map who belong to a participating guild.
2. For each, add/update in `CastleSiegeContext.ParticipantTracking`:
   - Increment `Seconds` by 5.

### 5. Post-Battle Rewards — method on `CastleSiegeContext` or a helper

Called from `CastleSiegePlugIn.OnEnterStateAsync(End)`:

1. Iterate all participants.
2. If `Seconds >= Configuration.ParticipantRewardMinSeconds`:
   - If player is online: send reward item to their Gremory Case / inventory.
   - If offline: insert a `PendingReward` record for later retrieval.
3. Award guild score: castle owner guild gets `GuildScoreCastleSiege` points via `IGuildServer.IncreaseGuildScoreAsync`.
4. Alliance member guilds get `GuildScoreCastleSiegeMembers` points each.

### 6. Guild List Packet Handlers

For the client to view the final guild list and registered guild list:

- **`CastleSiegeGuildListHandlerPlugIn`** — handles `C1-B5` (CastleOwnerListRequest). Returns the final participating guild list.
- **`CastleSiegeRegisteredGuildListHandlerPlugIn`** — handles `C1-B4`. Returns all registered guilds.

### 7. View Interfaces & Remote Views

- `ICastleSiegeJoinSidePlugIn` — sends join side notification.
- `ICastleSiegeGuildListPlugIn` — sends the final guild list.
- `ICastleSiegeRegisteredGuildListPlugIn` — sends registered guild list.

## Files to Create

| File | Description |
|------|-------------|
| `GameLogic/CastleSiege/CastleSiegeGuildSelector.cs` | Guild selection and scoring algorithm |
| `GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs` | Battle participation tracking |
| `GameLogic/Views/CastleSiege/ICastleSiegeJoinSidePlugIn.cs` | View interface for join side |
| `GameLogic/Views/CastleSiege/ICastleSiegeGuildListPlugIn.cs` | View interface for guild list |
| `GameLogic/Views/CastleSiege/ICastleSiegeRegisteredGuildListPlugIn.cs` | View interface for registered list |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeGuildListHandlerPlugIn.cs` | Packet handler |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeRegisteredGuildListHandlerPlugIn.cs` | Packet handler |
| `GameServer/RemoteView/CastleSiege/CastleSiegeJoinSidePlugIn.cs` | Remote view |
| `GameServer/RemoteView/CastleSiege/CastleSiegeGuildListPlugIn.cs` | Remote view |
| `GameServer/RemoteView/CastleSiege/CastleSiegeRegisteredGuildListPlugIn.cs` | Remote view |

## Acceptance Criteria

- [ ] Guild selection correctly computes scores: `(marks * 5) + members + (level / 4)`.
- [ ] Guilds are sorted by score descending, with earlier registration winning ties.
- [ ] Top N guilds are assigned Attack1/2/3 sides correctly.
- [ ] Castle owner is automatically assigned the Defense side.
- [ ] All alliance members are added to the final list with the same side as their master guild.
- [ ] Join sides are sent to players entering the CS map.
- [ ] Participants are tracked every 5 seconds during `Start`.
- [ ] Post-battle rewards are given to participants who meet the minimum time threshold.
- [ ] Guild score is awarded to the winning guild and alliance members.
- [ ] Final guild list and registered guild list are available via packet requests.

Zylkien:
## Summary

Implements Castle Siege guild selection, join-side synchronization, participant tracking, and post-battle rewards.

- Selects attacking guilds during the `Notify` state using `(marks * 5) + members + (combined level / 4)`, with registration order as the tie-breaker.
- Assigns the current castle owner to `Defense`.
- Expands selected guilds to their alliance members and persists the final guild list.
- Restores the persisted guild list after a server restart.
- Synchronizes player join sides and corresponding Castle Siege effects during `Ready` and `Start`, including players who enter or reconnect after the battle begins.
- Tracks participating characters every five seconds during the battle.
- Delivers configured participant rewards directly or stores them as persistent pending rewards for later delivery.
- Awards configured guild scores to the winning guild and its alliance members.
- Implements the registered-guild and final-guild-list packet handlers and remote views.
- Adds the required persistence entities, EF mappings, migration, initialization update, and guild-server lookup support.
- Castle Siege currently follows the single-game-server architecture specified in #735; exact-once guild-score coordination across multiple game-server processes is deferred to the planned IEventPublisher-based multi-server support.

## Implementation notes

- Participant item rewards use `CastleSiegeConfiguration.RewardItemDefinition`. No arbitrary default reward item is introduced.
- If the reward cannot be delivered because the character is offline or the inventory is full, it is persisted and delivered when the character next enters the game.
- Packet definitions are not duplicated in this PR; it uses the Castle Siege packet definitions already merged through #863.

## Testing

- Castle Siege and related guild-server tests: **50 passed**
- Season 6 EF initialization test: **1 passed**
- Relevant GameServer, persistence, and initialization projects build successfully.
- Warning-level analyzer checks for all changed hand-written files pass.
- `git diff --check` passes.

Closes #724


src/GameServer/RemoteView/CastleSiege/CastleSiegeJoinSidePlugIn.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CastleSiegeJoinSidePlugIn"/> class.
+    /// </summary>
+    /// <param name="player">The player.</param>
+    public CastleSiegeJoinSidePlugIn(RemotePlayer player) => this._player = player;
+
+    /// <inheritdoc />
+    public ValueTask ShowJoinSideAsync(JoinSide side)
+        => this._player.Connection.SendCastleSiegeJoinSideNotificationAsync((CastleSiegeJoinSide)side);
```
sven-n: `this._player.Connection` is dereferenced without a null check. The sibling view plug-ins added in this PR (`CastleSiegeGuildListPlugIn`, `CastleSiegeRegisteredGuildListPlugIn`) both guard with `if (this._player.Connection is not { } connection) return;`. Since `SynchronizePlayerJoinSideAsync` / `ClearPlayerJoinSideAsync` are invoked from the siege tick and from map add/remove events, a player disconnecting concurrently will have a null `Connection` and this throws a NullReferenceException inside the siege tick.

```suggestion
    public ValueTask ShowJoinSideAsync(JoinSide side)
        => this._player.Connection?.SendCastleSiegeJoinSideNotificationAsync((CastleSiegeJoinSide)side) ?? default;
```

---

src/GameLogic/CastleSiege/CastleSiegeContext.cs
```diff
@@ -416,4 +618,83 @@ private async ValueTask LoadRegistrationsAsync()
...
+                     npc => npc.IsAlive
+                            && npc.Definition.MonsterDefinition?.Number == LifeStoneMonsterNumber))
+        {
+            if (lifeStone.SpawnedInstance?.CurrentMap is { } map)
+            {
+                await map.RemoveAsync(lifeStone.SpawnedInstance).ConfigureAwait(false);
+            }
+
+            lifeStone.IsAlive = false;
+            lifeStone.SpawnedInstance = null;
```
sven-n: `KillLifeStonesAsync` is dead code as of this PR: every call site of `SetPlayerJoinSideAsync` (Ready, Start, EndCycle, and the participant tick in `CastleSiegePlugIn`) passes `killLifeStones: false`, so the `killLifeStones` branch never runs. Either wire it up to the state that is supposed to remove the Life Stones, or drop the parameter and the method.

---

src/Dapr/ServerClients/GuildServer.cs
```diff
@@ -72,6 +72,20 @@ public async ValueTask<bool> GuildExistsAsync(string guildName)
...
+    public async ValueTask<uint> GetGuildIdAsync(Guid guildId)
+    {
+        try
+        {
+            return await this._daprClient.InvokeMethodAsync<Guid, uint>(this._targetAppId, nameof(this.GetGuildIdAsync), guildId).ConfigureAwait(false);
+        }
+        catch (Exception ex)
+        {
+            this._logger.LogError(ex, "Unexpected error when retrieving a runtime guild identifier.");
+            return 0;
```
sven-n: Swallowing the exception and returning `0` is indistinguishable from "guild not found" for the new callers. In `CastleSiegeContext.LoadFinalGuildListAsync` and `CastleSiegeGuildSelector.SelectGuildsAsync` a `0` result silently drops the guild from the siege, and in `CastleSiegeParticipantTracker.AwardRewardsAsync` it silently skips the guild-score reward. A transient Dapr error during Notify/Ready would therefore quietly disqualify a registered guild with no visible failure.

---

sven-n (commented):
Low-effort diff review: 3 findings posted inline.

---

src/GameLogic/CastleSiege/CastleSiegePlugIn.cs
```diff
@@ -266,13 +366,26 @@ private async ValueTask OnTickAsync(CastleSiegeContext context, DateTime utcNow)
...
+            && context.NextParticipantUpdateUtc <= utcNow)
+        {
+            await context.SetPlayerJoinSideAsync().ConfigureAwait(false);
+            await CastleSiegeParticipantTracker
+                .TrackAsync(context, (int)ParticipantUpdateInterval.TotalSeconds)
+                .ConfigureAwait(false);
+            context.NextParticipantUpdateUtc = GetNextInterval(
+                context.StateStartTimeUtc,
+                utcNow,
+                ParticipantUpdateInterval);
```
sven-n, line 394: Participation is credited a hardcoded `(int)ParticipantUpdateInterval.TotalSeconds` (5 s) per tick, but the tick only fires when the periodic task runs *and* `NextParticipantUpdateUtc <= utcNow`, and the next due time is snapped forward via `GetNextInterval(StateStartTimeUtc, utcNow, 5s)`, which discards every missed interval.

So the credited amount is fixed while the real elapsed time between ticks is not. If the periodic task's period is longer than 5 s, or a tick is delayed, presence is under-credited — e.g. a player who was on the map for 60 s can be credited 5 s. `ParticipantRewardMinSeconds` may then never be reached and no participant is rewarded.

Suggest crediting the actual elapsed wall time since the previous update for that character instead of the constant.

---

src/GameLogic/CastleSiege/CastleSiegeGuildSelector.cs
sven-n, line 1: **Stale persisted guild list** (around `SelectGuildsAsync`, ~line 600 — commenting at file level since the line is outside the diff hunks):

`SelectGuildsAsync` returns early (after `FinalGuildList.Clear()`) when the context is not an `IGameServerContext`, which skips `SaveFinalGuildListAsync`.

The persisted `CastleSiegeData.Guilds` then still holds the *previous* cycle's guilds while the runtime `FinalGuildList` is empty. On the next restart `LoadFinalGuildListAsync` repopulates that stale selection, so a siege can start with guilds from an earlier cycle.

Either persist the cleared list before returning, or don't clear the runtime list on this path.

---

src/GameServer/MessageHandler/CastleSiege/CastleSiegeGuildListHandlerPlugIn.cs
sven-n, line 1: **Duplicated ordering** (~line 118 — commenting at file level since the line is outside the diff hunks):

The `OrderBy(Side).ThenByDescending(IsAllianceMaster).ThenBy(GuildName, OrdinalIgnoreCase)` chain is duplicated verbatim from `CastleSiegeContext.SaveFinalGuildListAsync` in this same PR.

Two copies of the display ordering will drift — a change in one place silently gives the client a different order than what was persisted. Worth extracting a single helper and calling it from both.

---

sven-n (commented):
Second review pass, on the updated head (`63cff82`). Confirmed the `Connection` null-dereference from the earlier pass is fixed in `CastleSiegeJoinSidePlugIn`.

Three new findings: one correctness issue on participant time crediting, one on the persisted final guild list going stale, and one duplicated-ordering cleanup. Details inline.

---

src/GameLogic/CastleSiege/CastleSiegeContext.cs
```diff
@@ -166,12 +180,73 @@ public CastleSiegeContext(IGameContext gameContext, CastleSiegeConfiguration con
...
-            ? guild.Side
+               && this.FinalGuildList.TryGetValue(guildStatus.GuildId, out var participatingGuild)
+            ? participatingGuild.Side
             : CastleSiegeJoinSide.None;
     }
 
+    /// <summary>
+    /// Assigns the current Castle Siege side to all online players on the Castle Siege map.
+    /// </summary>
+    /// <returns>A task that represents the asynchronous synchronization operation.</returns>
```
sven-n, line 203: Gating on `CurrentState != PlayerState.EnteredWorld` also excludes `Dead`, `ChangingMap`, `TradeOpened` and `NpcDialogOpened` — all normal states for a player who is standing on the siege map.

Consequences: a dead player loses their join side and their participation credit for as long as they are dead (which during a siege is often), and gets a duplicate join-side packet when they respawn back into `EnteredWorld`.

Suggest checking that the player is on the siege map and has a selected character, rather than for one specific state.

---

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Starts tracking a player who entered the Castle Siege map during the battle.
+    /// </summary>
+    /// <param name="context">The Castle Siege context.</param>
+    /// <param name="player">The player.</param>
+    /// <param name="utcNow">The current UTC time.</param>
+    internal static void StartTracking(CastleSiegeContext context, Player player, DateTime utcNow)
```
sven-n, line 106: `StartTracking` resets `LastUpdateUtc` to now without first crediting the time already accrued.

`PlayerStateChangedAsync` reaches this on every transition back into `EnteredWorld` — respawn, closing an NPC dialog, ending a trade. Each of those throws away the interval since the last update, so a player who dies (or opens a dialog) more often than the update interval never accumulates `ParticipationTime`, never reaches `ParticipantRewardMinSeconds`, and never gets the reward. During a siege that describes most participants.

Credit the elapsed span before resetting the timestamp.

---

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
```diff
@@ -0,0 +1,170 @@
...
+                }
+
+                await CastleSiegeRewardDelivery
+                    .QueueAsync(context.GameContext, participant.CharacterId, rewardDefinition)
+                    .ConfigureAwait(false);
+            }
+        }
+
+        if (context.GameContext is not IGameServerContext gameServerContext
+            || context.SiegeData.OwnerGuildId is not { } ownerGuildId
```
sven-n: `AwardRewardsAsync` derives the winning guild from `SiegeData.OwnerGuildId`, but nothing in this PR updates `OwnerGuildId` from the battle outcome — it still holds the pre-battle owner at the point the rewards run.

So the guild score always goes to the guild that owned the castle before the siege, and attackers who actually capture it get nothing. Either set the new owner before awarding, or take the winner from the battle result rather than from the stored owner.

---

src/GameLogic/CastleSiege/CastleSiegeGuildSelector.cs
```diff
@@ -0,0 +1,264 @@
...
+        CastleSiegeContext context,
+        IGameServerContext gameServerContext,
+        uint masterGuildId,
+        Guid persistentMasterGuildId,
+        string masterGuildName,
+        CastleSiegeJoinSide side,
+        int score)
+    {
+        var allianceGuilds = await gameServerContext.GuildServer
+            .GetAllianceGuildsAsync(masterGuildId)
```
sven-n, line 189: `GetAllianceGuildsAsync` only enumerates guild containers that are already loaded in memory, i.e. guilds with at least one online member.

An alliance guild whose members are all offline at selection time is therefore silently dropped from `FinalGuildList`. When those members log in during the siege they resolve to join side `None` — no side, no effect, no participation credit — even though their alliance was selected. Since selection happens in `Notify`, well before the battle, that is a likely rather than exotic case.

The alliance membership should be resolved from persistence, not from the loaded containers.

---

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
```diff
@@ -0,0 +1,170 @@
...
+    /// <param name="context">The Castle Siege context.</param>
+    /// <returns>A task that represents the asynchronous reward operation.</returns>
+    public static async ValueTask AwardRewardsAsync(CastleSiegeContext context)
+    {
+        var eligibleParticipants = context.ParticipantTracking.Values
+            .Where(participant => participant.ParticipationTime.TotalSeconds >= context.Configuration.ParticipantRewardMinSeconds)
+            .ToList();
+        var onlinePlayers = (await context.GameContext.GetPlayersAsync().ConfigureAwait(false))
+            .Where(player => player.SelectedCharacter is not null)
+            .ToDictionary(player => player.SelectedCharacter!.Id);
```
sven-n: `ToDictionary` keyed on the character id throws `ArgumentException` if the same character appears twice — which happens transiently while a player reconnects and the old and new player objects overlap.

Because this runs at the end of the siege, one duplicate aborts the whole reward pass: no participant rewards and no guild scores for that siege, for everyone. Use `GroupBy(...).ToDictionary(g => g.Key, g => g.First())` (or a `TryAdd` loop) so a duplicate can't take down the payout.

---

src/GameLogic/CastleSiege/CastleSiegePendingRewardPlugIn.cs
```diff
@@ -0,0 +1,62 @@
...
+                .FirstOrDefault(item => item.GetId() == pendingReward.ItemDefinitionId);
+            if (rewardDefinition is null
+                || !await CastleSiegeRewardDelivery
+                    .TryAddToInventoryAsync(player, rewardDefinition)
+                    .ConfigureAwait(false))
+            {
+                continue;
+            }
+
+            await persistenceContext.DeleteAsync(pendingReward).ConfigureAwait(false);
```
sven-n: The pending-reward row is deleted and that deletion committed while the granted item still lives unsaved in the player's own context.

If the process crashes (or the player's save fails) between the two, the row is gone and the item was never persisted — the reward is lost permanently, with no way to reconstruct it. The whole point of the pending-reward table is to survive exactly that.

Delete the row only after the item is durably saved, so the worst case is a duplicate delivery attempt rather than a silent loss.

---

src/GameLogic/CastleSiege/CastleSiegePlugIn.cs
```diff
@@ -57,20 +58,49 @@ public async ValueTask ObjectAddedToMapAsync(GameMap map, ILocateable addedObjec
...
+        try
+        {
+            CastleSiegeParticipantTracker.StopTracking(
+                context,
+                player,
+                this._timeProvider.GetUtcNow().UtcDateTime);
+            await context.ClearPlayerJoinSideAsync(player).ConfigureAwait(false);
+        }
+        finally
         {
```
sven-n: The map add/remove and player-state hooks take the unbounded `ExecutionLock` before checking whether the event is even relevant to the siege, and the periodic task holds that same lock across database loads, per-guild remote guild-server calls, and view sends.

The result is server-wide contention on a hot path: every login and every map move anywhere on the server can block behind a siege tick doing I/O. `GameMap` raises these events outside its own lock, so this is contention rather than deadlock — but it is still a stall proportional to how slow the guild server and DB are.

Two things would help: cheap relevance checks (siege active, correct map) before acquiring the lock, and moving the I/O out of the locked region.

---

src/GameLogic/CastleSiege/CastleSiegeContext.cs
sven-n, line 1: **Unsynchronized read of `ActiveEffects`** (in `SetJoinSideMagicEffectAsync`, ~line 567 — file-level comment because the line falls outside the diff hunks):

The method enumerates `MagicEffectList.ActiveEffects` directly. That is a `SortedList` guarded inside `MagicEffectList` by `_addLock` — reading it from outside without the lock races with concurrent effect add/remove and can throw or observe a torn view.

This runs every 5 seconds for every online player on the siege map, so the window is not small. Go through the `MagicEffectList` API that takes the lock instead of touching the collection directly.

---

src/GameLogic/CastleSiege/CastleSiegeGuildSelector.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            return onlineGuildMaster.Level
+                   + (int)(onlineGuildMaster.Attributes?[Stats.MasterLevel] ?? 0);
+        }
+
+        using var persistenceContext = context.GameContext.PersistenceContextProvider.CreateNewTypedContext(
+            typeof(Character),
+            false,
+            context.GameContext.Configuration);
+        var character = (await persistenceContext.GetAsync<Character>().ConfigureAwait(false))
```
sven-n: When the guild master is offline, `GetCombinedLevelAsync` falls back to materializing the entire `Character` table and filtering in memory — once per registered guild in that situation, and all of it while the execution lock is held.

On a populated server that is a full table scan per guild during selection. Query the characters by guild instead of loading them all.

---

src/GameLogic/CastleSiege/CastleSiegePendingRewardPlugIn.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            return;
+        }
+
+        using var persistenceContext = player.GameContext.PersistenceContextProvider.CreateNewTypedContext(
+            typeof(CastleSiegePendingReward),
+            false,
+            player.GameContext.Configuration);
+        var pendingRewards = (await persistenceContext
+                .GetAsync<CastleSiegePendingReward>()
```
sven-n: This loads the entire `CastleSiegePendingReward` table and filters it in memory, on every character login — including the overwhelming majority of logins that have no pending reward at all.

Filter by character id in the query so a login costs an indexed lookup instead of a full table load.

---

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
```diff
@@ -0,0 +1,170 @@
...
+                .GetGuildIdAsync(guild.PersistentGuildId)
+                .ConfigureAwait(false);
+            if (runtimeGuildId == 0)
+            {
+                continue;
+            }
+
+            for (var point = 0; point < score; point++)
+            {
+                await gameServerContext.GuildServer.IncreaseGuildScoreAsync(runtimeGuildId).ConfigureAwait(false);
```
sven-n: `IncreaseGuildScoreAsync` is invoked once per score point in a loop, so awarding a configured score of a few hundred means a few hundred round trips to the guild server — and each iteration re-resolves a guild id that is already available here.

Pass the amount once (or at minimum hoist the id resolution out of the loop).

---

sven-n (commented):
Third review pass, on head `819bfc51` — a deeper read of the hand-written sources (`src/GameLogic/CastleSiege/*`, views, handlers, guild server, data model, EF/initialization) with the surrounding types opened for context. 11 findings inline: six correctness, two concurrency, three efficiency.

The ones I'd consider blocking are the reward path — participation time being discarded on every respawn, the winner being read from the pre-battle `OwnerGuildId`, and the pending-reward row being deleted before the item is durably saved — since together they mean rewards can silently go to nobody or to the wrong guild.

Note the earlier finding about the hardcoded 5-second credit in `CastleSiegePlugIn` is superseded: the tracking code was reworked in this commit, and the same underlying bug now lives in `CastleSiegeParticipantTracker.StartTracking`.

Checked and found fine: the `CastleSiegeJoinSide` DataModel↔network enum values match; `CastleSiegeMagicEffectNumber` 14–17 don't collide with any effect used in initialization, and the update plug-in does create the definitions; the durability convention in `CastleSiegeRewardDelivery` matches existing code; `RegisteredGuilds`, `FinalGuildList` and `ParticipantTracking` are concurrent collections, so the lock-free packet handlers are safe; and `GameMap` raises its add/remove events outside any lock, so the lock usage above is contention rather than deadlock.

Merge conflicts against `master` were out of scope for this pass.

---

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
```diff
@@ -0,0 +1,185 @@
...
+        Player player,
+        DateTime utcNow,
+        bool isTracking,
+        bool allowPlayerOutsideSiegeMap)
+    {
+        if ((!allowPlayerOutsideSiegeMap
+             && player.CurrentMap?.Definition.Number != context.Configuration.CastleSiegeMapDefinition?.Number)
+            || player.SelectedCharacter is not { } character
+            || player.GuildStatus is not { } guildStatus
+            || context.GetPlayerJoinSide(player) == CastleSiegeJoinSide.None)
```
sven-n, line 139: `StopTracking` (called from `ObjectRemovedFromMapAsync` with `allowPlayerOutsideSiegeMap: true`) will always bail out here: `CastleSiegeContext.GetPlayerJoinSide` now returns `None` as soon as `player.CurrentMap` is no longer the siege map, which is exactly the case when the player is being removed from the map (or logs out, where `CurrentMap` is null). So the final tracking interval is never flushed and the participant stays with `IsTracking = true` and a stale `LastUpdateUtc` — a player who leaves right before the End state loses all participation time accumulated since the last 5s tick, and a disconnecting player keeps an interval that is never closed. The `allowPlayerOutsideSiegeMap` flag needs to bypass the join-side lookup too (e.g. fall back to the already-cached `context.PlayerJoinSides[character.Id]` / `FinalGuildList`).

---

src/Dapr/ServerClients/GuildServer.cs
```diff
@@ -196,11 +210,15 @@ public async ValueTask<GuildPosition> GetGuildPositionAsync(Guid characterId)
...
-    public async ValueTask IncreaseGuildScoreAsync(uint guildId)
+    public async ValueTask IncreaseGuildScoreAsync(uint guildId, int amount)
     {
         try
         {
-            await this._daprClient.InvokeMethodAsync(this._targetAppId, nameof(this.IncreaseGuildScoreAsync), guildId).ConfigureAwait(false);
+            await this._daprClient.InvokeMethodAsync(
+                    this._targetAppId,
+                    nameof(this.IncreaseGuildScoreAsync),
+                    (GuildId: guildId, Amount: amount))
```
sven-n, line 220: Passing a `ValueTuple` as the Dapr request body silently breaks in the hosted setup: `System.Text.Json` only serializes public *properties*, and `ValueTuple`'s `Item1`/`Item2` are fields, so this serializes to `{}` and the controller's `[FromBody] (uint GuildId, int Amount)` deserializes to `(0, 0)`. Guild scores would then never be increased (and `IncreaseGuildScoreAsync` would be called for guild id 0) in the Dapr deployment. Use a small DTO/record with properties for the body on both sides.

---

sven-n (commented):
Low-effort review pass: 2 correctness findings posted inline.

---

src/Dapr/ServerClients/GuildServer.cs
sven-n, line 220: **The score increase is silently lost in the Dapr deployment.**

`ValueTuple` exposes `Item1`/`Item2` as public *fields*, and `System.Text.Json` (which `DaprClient` uses, with no custom `JsonSerializerOptions` configured anywhere in `src/Dapr` — `DaprService.cs` just calls `services.AddDaprClient()`) does not serialize fields: `JsonSerializerOptions.IncludeFields` defaults to `false`. The request body therefore goes out as `{}`, and `[FromBody] (uint GuildId, int Amount)` on `GuildServerController.IncreaseGuildScoreAsync` deserializes it back to `(0, 0)`. `GuildServer.IncreaseGuildScoreAsync` then hits `amount > 0 == false` and returns without doing anything — no exception, nothing logged.

Concrete failure: on a Dapr deployment, the winning alliance never receives `GuildScoreCastleSiege`/`GuildScoreCastleSiegeMembers` after a siege. It is also a **regression for guild war**: `GuildWarAnswerAction` previously sent a bare `uint` (which serializes fine) and now goes through this same path, so guild-war winners stop getting their point too.

Every other multi-argument method in this very file uses a dedicated record for exactly this reason (`GuildCreationArguments`, `GuildMemberCreationArguments`, `GuildMemberRoleChangeArguments`, `PlayerOnlineStateArguments`). The tuples at lines 234/290/304 are not a counter-example — those methods have no corresponding controller endpoint at all, so they were never exercised.

Suggest adding e.g. `public record GuildScoreIncreaseArguments(uint GuildId, int Amount);` next to the other argument records and using it on both sides.

---

src/GameLogic/CastleSiege/CastleSiegeRewardDelivery.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        if (player.Inventory is null)
+        {
+            return false;
+        }
+
+        var item = player.PersistenceContext.CreateNew<Item>();
+        item.Definition = rewardDefinition;
+        item.Durability = item.IsStackable() ? 1 : rewardDefinition.Durability;
+        if (await player.Inventory.AddItemAsync(item).ConfigureAwait(false))
```
sven-n, line 32: **The delivered reward item is never sent to the client.**

`InventoryStorage.AddItemAsync` only mutates server-side state; it does not notify the client. Every other place in the codebase that grants an item into an online player's inventory follows up with `IItemAppearPlugIn` (see `QuestCompletionAction.cs:124`, `ItemStackAction.cs:64`, `PickupItemAction.cs:48`, `SymbolOfKundunStackedPlugIn.cs:41`).

Both call sites are affected:

1. `CastleSiegeParticipantTracker.AwardRewardsAsync` — a participant who is still online when the siege enters `End` gets the item added server-side, but their client keeps showing that slot as empty until they relog.
2. `CastleSiegePendingRewardPlugIn` — worse, because it runs on the `CharacterSelection → EnteredWorld` transition, which `Player.OnPlayerEnteredWorldAsync` raises from `ClientReadyAfterMapChangeAsync()` at line 2731, i.e. **after** `IUpdateInventoryListPlugIn.UpdateInventoryListAsync()` was already sent at line 2718. So the pending reward is added, committed via `SaveProgressAsync`, and the client never learns about it in that session either — the "delivered when the character next enters the game" path only becomes visible on the login *after* the one that delivered it.

Adding `await player.InvokeViewPlugInAsync<IItemAppearPlugIn>(p => p.ItemAppearAsync(item))` on the success path fixes both.

---

src/GuildServer/GuildServer.cs
```diff
@@ -437,10 +446,24 @@ public async ValueTask<IImmutableList<AllianceGuildEntry>> GetAllianceGuildsAsyn
             return ImmutableList<AllianceGuildEntry>.Empty;
         }
 
-        return this._guildDictionary.Values
-            .Where(g => GetAllianceMasterGuid(g.Guild) == masterGuid)
-            .Select(g => new AllianceGuildEntry(g.Id, g.Guild.Name ?? string.Empty, g.Guild.Members.Count, g.Guild.Logo))
-            .ToImmutableList();
+        using var context = this._persistenceContextProvider.CreateNewGuildContext();
+        var persistentGuilds = await context.GetAlliancesAsync(masterGuid).ConfigureAwait(false);
```
sven-n, line 450: **This turns alliance chat into a per-message database round-trip.**

Replacing the in-memory `_guildDictionary` scan with `CreateNewGuildContext()` + `GetAlliancesAsync()` correctly fixes the "fully offline alliance guilds are dropped" problem for Castle Siege, but `GetAllianceGuildsAsync` is not a Castle-Siege-only API. Its hot caller is `GameServerContext.ForEachAlliancePlayerAsync` (`src/GameServer/GameServerContext.cs:149`), which `GameServer.cs:223` invokes for **every alliance chat message**. It is also called from the client-triggered `AllianceListRequestHandlerPlugIn`, which a player can spam.

So each alliance chat line now allocates a `DbContext`, issues a `SELECT ... WHERE AllianceGuildId = @p` with `Include(RawMembers)`, and (in the Dapr split-server setup) does that behind a remote call. Previously it was a dictionary scan.

Since `CreateGuildContainerAsync` already eagerly materialises the whole alliance whenever any of its members enters the game, and `GuildMemberLeftGameAsync` deliberately keeps alliance containers resident ("Keep alliances in memory for simplicity"), the DB round-trip is only needed for the case Castle Siege cares about: an alliance where *no* member is online. Consider keeping the in-memory path as the fast path and only falling back to the DB query when it yields nothing, or moving the DB-backed lookup behind a separate method used by the siege selector.

---

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
```diff
@@ -0,0 +1,185 @@
...
+                    ParticipationTime = participationTime,
+                    LastUpdateUtc = utcNow,
+                    IsTracking = isTracking,
+                };
+            });
+    }
+
+    private static CastleSiegeGuildParticipant? GetWinningGuild(CastleSiegeContext context)
+    {
+        if (context.MiddleOwnerGuildId is { } runtimeGuildId)
```
sven-n, line 176: **`MiddleOwnerGuildId` is never assigned by production code, so the fallback always wins — and it names the *previous* owner.**

Grepping the whole tree, the only writes to `MiddleOwnerGuildId` are `CastleSiegePlugIn.cs:354` (`= null` on `EndCycle`) and two assignments in `CastleSiegeGuildSelectionTests`. Nothing in the Crown/switch flow sets it yet ("Start-state Crown, switch and mini-map ticks are implemented by their dedicated phases"). And `SiegeData.OwnerGuildId` is not updated from the battle outcome anywhere either — `OnEnterStateAsync(End)` awards rewards but never writes a new owner.

Net effect at runtime: `GetWinningGuild` always takes the `SiegeData.OwnerGuildId` branch, which is the guild that owned the castle *before* this siege. Concrete scenario: guild A defends and loses the castle to attacker B. On `End`, A is still `OwnerGuildId`, so A's `Defense` side receives `GuildScoreCastleSiege` / `GuildScoreCastleSiegeMembers` and B receives nothing — the exact inversion of the intended reward. When the castle is unowned (`OwnerGuildId == null`, the initialiser's default) no side is scored at all.

The plumbing added here is right; it just needs the winner to actually be recorded. Until the Crown phase lands, it would be safer to skip the guild-score award entirely when `MiddleOwnerGuildId` is null rather than falling back to the stale owner.

---

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
```diff
@@ -0,0 +1,185 @@
...
+
+            var runtimeGuildId = await gameServerContext.GuildServer
+                .GetGuildIdAsync(guild.PersistentGuildId)
+                .ConfigureAwait(false);
+            if (runtimeGuildId == 0)
+            {
+                continue;
+            }
+
+            await gameServerContext.GuildServer.IncreaseGuildScoreAsync(runtimeGuildId, score).ConfigureAwait(false);
```
sven-n, line 96: **Guild score is multiplied by the number of game-server processes.**

`CastleSiegePlugIn` holds one `CastleSiegeContext` per `IGameContext` (`ConditionalWeakTable<IGameContext, CastleSiegeContext>`), and `ExecuteTaskAsync` is an `IPeriodicTaskPlugIn` that runs independently in every game server. Each of them therefore enters the `End` state on its own schedule and calls `AwardRewardsAsync`.

Participant item rewards are naturally partitioned (each server only tracks its own players), but `IncreaseGuildScoreAsync` targets the *shared* guild server. With N game servers the winning guild gets `N × GuildScoreCastleSiege` and each alliance member `N × GuildScoreCastleSiegeMembers`. `ClearRegistrationsAsync` and `SaveFinalGuildListAsync` are likewise re-executed N times.

The class remark on `CastleSiegeContext` only warns about concurrent *registration* changes; this is a new globally-visible side effect that needs either an owning-server election or an idempotency guard (e.g. persisting a "scores awarded for cycle X" marker on `CastleSiegeData` and checking it before awarding).

---

src/GameLogic/CastleSiege/CastleSiegeContext.cs
```diff
@@ -166,12 +180,66 @@ public CastleSiegeContext(IGameContext gameContext, CastleSiegeConfiguration con
...
+    {
+        var mapNumber = this.Configuration.CastleSiegeMapDefinition?.Number;
+        if (mapNumber is null)
+        {
+            return;
+        }
+
+        var activeCharacterIds = new HashSet<Guid>();
+        var activePlayers = new HashSet<Player>();
+        foreach (var player in await this._gameContext.GetPlayersAsync().ConfigureAwait(false))
```
sven-n, line 214: **Every 5 seconds during the battle this walks *all* online players and takes each one's magic-effect lock four times.**

`OnTickAsync` calls `SetPlayerJoinSideAsync()` on every `NextParticipantUpdateUtc` tick (`ParticipantUpdateInterval` = 5 s) while the state is `Start`. This loop iterates `GetPlayersAsync()` — the whole server, not just the siege map — and for every player *not* on the siege map calls `ClearPlayerJoinSideAsync` → `SetJoinSideMagicEffectAsync(player, None)`, which awaits `TryGetActiveEffectAsync` four times (once per `CastleSiegeMagicEffectNumber`), each acquiring that player's `MagicEffectsList._addLock`.

On a 1000-player server that is ~4000 sequentially awaited lock acquisitions every 5 seconds for the entire battle, all inside `ExecutionLock`, and contending with those players' own packet handlers. Almost all of that work is a no-op: a player who was never on the siege map can never have one of these effects.

Cheap fixes: iterate the already-maintained `_siegeMapPlayers` set for the assign path, and only run the clear path for character ids actually present in `PlayerJoinSides` / players present in `_notifiedPlayerJoinSides`, instead of for the entire server population.

---

src/GameLogic/CastleSiege/CastleSiegePendingRewardPlugIn.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        var pendingRewards = await player.PersistenceContext
+            .GetPendingCastleSiegeRewardsAsync(character.Id)
+            .ConfigureAwait(false);
+        var deliveredAny = false;
+        foreach (var pendingReward in pendingRewards)
+        {
+            var rewardDefinition = player.GameContext.Configuration.Items
+                .FirstOrDefault(item => item.GetId() == pendingReward.ItemDefinitionId);
+            if (rewardDefinition is null
```
sven-n, line 38: **An unresolvable `ItemDefinitionId` leaves the row in the table forever.**

The `continue` covers two very different cases. "Inventory full" is a legitimate retry-next-login. "`rewardDefinition is null`" is not: the item definition is gone from the game configuration (config re-initialised, item removed, or the row was written by a server with a different configuration), so the lookup will fail identically on every future login. The row is then re-read and re-skipped on every `CharacterSelection → EnteredWorld` transition for the lifetime of the character.

Worth splitting the two branches and deleting the row (with a warning log) when the definition cannot be resolved, so the queue can't accumulate permanently-undeliverable entries.

---

src/GameLogic/CastleSiege/CastleSiegeRewardDelivery.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <param name="gameContext">The game context.</param>
+    /// <param name="characterId">The persistent character identifier.</param>
+    /// <param name="rewardDefinition">The reward item definition.</param>
+    /// <returns>A task that represents the asynchronous queue operation.</returns>
+    internal static async ValueTask QueueAsync(
+        IGameContext gameContext,
+        Guid characterId,
+        ItemDefinition rewardDefinition)
+    {
+        using var persistenceContext = gameContext.PersistenceContextProvider.CreateNewTypedContext(
```
sven-n, line 53: One persistence context and one `SaveChangesAsync` round-trip per queued reward.

`AwardRewardsAsync` calls this in a loop over `eligibleParticipants`, so a siege where the reward could not be handed over directly (offline characters, full inventories) produces N `CreateNewTypedContext` + N `SaveChangesAsync` calls back to back, synchronously blocking the state transition into `End` while `ExecutionLock` is held. Passing the whole batch and writing it through a single context would keep this to one round-trip.

---

src/GameLogic/CastleSiege/CastleSiegeGuildSelector.cs
```diff
@@ -0,0 +1,264 @@
...
+            await AddGuildAndAllianceAsync(
+                    context,
+                    gameServerContext,
+                    candidate.RuntimeGuildId,
+                    candidate.Registration.GuildId,
+                    candidate.Registration.GuildName,
+                    side,
+                    candidate.Score)
+                .ConfigureAwait(false);
+            attackSideIndex++;
```
sven-n, line 129: `attackSideIndex` is incremented unconditionally, even when `AddGuildAndAllianceAsync` added nothing.

If the candidate's alliance is non-empty but every `GetPersistentGuildIdAsync` returns `null` (the `continue` inside the alliance loop), the attack slot is consumed and `Attack1` ends up with zero guilds, while the next-ranked candidate is pushed to `Attack2` — or dropped entirely once `maximumAttackers` is reached. Having `AddGuildAndAllianceAsync` report whether it actually added a participant, and only advancing the index on success, would keep the configured number of attacking sides populated.

---

sven-n (commented):
Reviewed at `97c7c8b3` (max effort, single careful diff pass).

**Status of the 11 findings from the earlier deep pass**

| # | Finding | Status at `97c7c8b3` |
|---|---|---|
| 1 | `PlayerState != EnteredWorld` gate excluded Dead/ChangingMap/TradeOpened/NpcDialogOpened | **Fixed** — both gates removed from `SetPlayerJoinSideAsync` and `SynchronizePlayerJoinSideAsync`; side/effect now survive transient states. Test added. |
| 2 | `StartTracking` reset `LastUpdateUtc` without crediting | **Fixed** — `CastleSiegeParticipant.IsTracking` + functional `AddOrUpdate`; credit is driven by the *previous* interval's flag. `StopTracking` runs from `ObjectRemovedFromMapAsync` while `CurrentMap` is still the siege map, so the interval does close. Verified by the new 23 s / 30 s assertions. |
| 3 | Winner derived from `SiegeData.OwnerGuildId`, never updated from the battle | **Remains** — `GetWinningGuild` adds the `MiddleOwnerGuildId` path, but nothing in production ever sets it (only tests, plus `= null` on `EndCycle`), and `OwnerGuildId` is still never written at `End`. See inline comment. |
| 4 | `GetAllianceGuildsAsync` only enumerated loaded containers | **Fixed** — now DB-backed via `GetAlliancesAsync`, offline alliance guilds are materialised. But it regressed the alliance-chat hot path; see inline comment. |
| 5 | `ToDictionary` on character id threw on reconnect overlap | **Fixed** — `GroupBy(...).ToDictionary(g => g.Key, g => g.First())`, with a dedicated regression test. |
| 6 | Pending-reward row deleted+committed before the item was durably saved | **Fixed** — the row is now loaded/deleted through `player.PersistenceContext` and committed together with the item by `SaveProgressAsync`. (`CastleSiegePendingReward` is not a Configuration type, so `AccountContext` does not ignore it — the `Set<>` call resolves.) |
| 7 | `ExecutionLock` taken before relevance checks and held across I/O | **Partially / superseded** — player map and state-change events no longer take the lock at all, which was the harmful part. The state machine still holds it across DB/remote/view I/O, but it is acquired with `WaitAsync(0)`, so a slow tick is skipped rather than queued. Acceptable. |
| 8 | `SetJoinSideMagicEffectAsync` read `ActiveEffects` without `_addLock` | **Fixed** — new `MagicEffectsList.TryGetActiveEffectAsync` takes `_addLock`, and the lock is released before `DisposeAsync`, so no re-entrancy against `OnEffectTimeOutAsync`. |
| 9 | `GetCombinedLevelAsync` materialised the whole `Character` table | **Fixed** — replaced by `GetAccountByCharacterNameAsync`, which is an indexed lookup. It still pulls the full account aggregate per offline guild master, but that is bounded by the number of registered guilds. |
| 10 | `CastleSiegePendingRewardPlugIn` loaded the whole pending-reward table per login | **Fixed** — new `IPlayerContext.GetPendingCastleSiegeRewardsAsync` filters server-side, backed by the `IX_CastleSiegePendingReward_CharacterId` index added in the migration. |
| 11 | `IncreaseGuildScoreAsync` called once per score point | **Fixed in shape, broken in the Dapr transport** — the loop is gone and the interface takes an amount, but the new remote payload is a `ValueTuple`, which `System.Text.Json` cannot serialise. See the inline comment on `src/Dapr/ServerClients/GuildServer.cs`; this also regresses guild-war scoring. |

**New findings this pass** — 9 inline comments, ordered roughly by severity: the Dapr `ValueTuple` payload (11), the missing `IItemAppearPlugIn` notification on reward delivery, the alliance-chat DB round-trip introduced by (4), the dead winner resolution (3), per-game-server score multiplication, the 5-second all-players magic-effect sweep, undeletable pending rows, per-reward persistence contexts, and the consumed-but-empty attack slot.

Everything else I checked came back clean: packet handler keys `0xB4`/`0xB5` do not collide with existing handlers, the `CastleSiegeJoinSide` data-model and network enums have identical values, magic-effect numbers 14–17 are unused elsewhere, both `PlugInResources` designer/resx pairs are complete, the migration matches the model builder configuration, and `CastleSiegeData.Guilds` is eagerly loaded by the repository (no lazy-load-on-disposed-context hazard).

---

sven-n (changes requested):

src/GameLogic/CastleSiege/CastleSiegeParticipantTracker.cs
sven-n, line 1: **`GetWinningGuild` always returns `null`, so no guild score is ever awarded** (commenting at file level — the method sits at the end of the file, outside the diff hunks).

```csharp
private static CastleSiegeGuildParticipant? GetWinningGuild(CastleSiegeContext context)
{
    return context.MiddleOwnerGuildId is { } runtimeGuildId
        ? context.FinalGuildList.GetValueOrDefault(runtimeGuildId)
        : null;
}
```

`context.MiddleOwnerGuildId` is never *assigned* anywhere in production code — the only production write is in `CastleSiegePlugIn`, where `EndCycle` sets it back to `null`. The assignments I can find live in tests.

So at `End` state it is `null`, `GetWinningGuild` returns `null`, and `AwardRewardsAsync` hits

```csharp
if (context.GameContext is not IGameServerContext gameServerContext
    || GetWinningGuild(context) is not { } ownerGuild)
{
    return;
}
```

and returns before the score loop. The winning alliance receives neither `GuildScoreCastleSiege` nor `GuildScoreCastleSiegeMembers`, every siege.

This is the same root cause I raised on the previous head, where it credited the pre-siege owner via `OwnerGuildId`. The `MiddleOwnerGuildId` indirection was added since, but nothing feeds it, so the outcome moved from *wrong guild* to *no guild*. Whatever determines the castle owner at the end of the battle needs to write `MiddleOwnerGuildId` (or `GetWinningGuild` needs to derive the winner from the battle result directly).

Note the participant *item* rewards above this point are unaffected — they run before the early return.

---

src/GuildServer/GuildServer.cs
sven-n, line 1: **The DB fallback in `GetAllianceGuildsAsync` is unreachable** (file-level comment — the method is around line 449, outside the diff hunks).

Moving the runtime cache in front of the query fixed the hot-path problem I raised last round (`ForEachAlliancePlayerAsync` on every alliance chat message, and the client-spammable `AllianceListRequestHandlerPlugIn`) — good. But it also made the `GetAlliancesAsync` fallback dead:

`masterGuid` can only be resolved from a container that is already in `_guildDictionary`, and that container itself satisfies `GetAllianceMasterGuid(...) == masterGuid`. So on the reachable path `runtimeGuilds` always contains at least that one guild and is never empty — the `if (runtimeGuilds.Count == 0)` branch never runs.

The consequence is that the case the DB fallback was added for is still not covered: an alliance guild with no online member is absent from `_guildDictionary`, so it is still dropped from the alliance enumeration, and its members still resolve to join side `None` during a siege.

Resolving the alliance membership from persistence when the *requested* alliance is not fully represented in the cache — rather than only when the runtime result is entirely empty — would cover it without putting the query back on the chat path.

---

sven-n (commented):
Full-depth pass on head `46af7113`. Most of the previous round is addressed — two findings left, both inline.

**New / remaining**

1. `CastleSiegeParticipantTracker.GetWinningGuild` — `MiddleOwnerGuildId` is never assigned in production (only cleared, in `EndCycle`), so `AwardRewardsAsync` returns before the score loop and the winning alliance gets no guild score at all. This is the previous "winner taken from the pre-battle owner" finding in a new shape: the indirection was added, but nothing feeds it, so the outcome moved from *wrong guild* to *no guild*.
2. `GuildServer.GetAllianceGuildsAsync` — putting the runtime cache first fixed the hot-path query, but it made the `GetAlliancesAsync` DB fallback unreachable, so fully-offline alliance guilds are still dropped and their members still resolve to join side `None`.

**Fixed since the last pass**

- The Dapr `ValueTuple` body — now a `GuildScoreIncreaseArguments` record wired through interface, client and controller, with `GuildWarAnswerAction` passing `1`. That also un-breaks guild war.
- `IItemAppearPlugIn` is invoked after `AddItemAsync`, and the pending path reuses `TryAddToInventoryAsync`, so a delivered reward is visible immediately instead of on the next login.
- Pending rows with an unresolvable `ItemDefinitionId` are logged and deleted instead of retrying forever.
- `QueueAsync` takes a collection and writes all rows in one context/save.
- `AddGuildAndAllianceAsync` returns `bool` and the attack slot is only consumed on success, plus an explicit skip for already-selected alliance members.

**Judged acceptable**

- N× `AwardRewardsAsync` across game-server processes: the code now carries an explicit remark pointing at the future multi-server coordinator and the PR description defers this to #735 — fine as documented scope.
- The 5-second join-side sweep: now iterates only `GetSiegePlayers()` and takes a single `GetActiveEffectsSnapshotAsync` lock instead of four. Still runs under `ExecutionLock` every 5 s, but that is a perf note rather than a correctness problem.

This environment has no .NET SDK, so the review is static analysis — nothing was compiled or run.

---

Issue #726, opened 2026-03-24
Castle Siege Crown & Switch Win-Condition Mechanics


sven-n:
## Summary

Implement the core Castle Siege win condition: the crown capture system. Three players from the same attacking side must simultaneously occupy the crown and both switches for a configured duration to capture the castle.

## Prerequisites

- #725 (NPCs) — for `CastleSiegeCrown`, `CastleSiegeSwitch` NPC classes and their intelligence.
- #724 (Guild Selection) — for join side assignments and `SetPlayerJoinSideAsync`.
- #722 (State Machine) — for `CastleSiegeContext` and `Start` state tick handlers.

### 1. `CheckMiddleWinner()` — **In `GameLogic/CastleSiege/CastleSiegeCrownMechanics.cs`** (new file)

Called every tick (every second) during `Start` state.

**Logic:**
1. Get `CrownUser`, `SwitchUsers[0]`, `SwitchUsers[1]` from `CastleSiegeContext`.
2. All three must be:
   - Non-null.
   - Alive (`IsAlive`).
   - In a guild (`GuildStatus != null`).
   - On an **attacking** side (not `None`, not `Defense`).
   - On the **same** attacking side.
3. If conditions are not met:
   - If the crown user changed or left, send `CrownAccessState.Fail` to the previous user.
   - Do not reset accumulated time (capped at `CrownHoldTimeSeconds - 1` on fail).
   - Return.
4. Increment `CrownAccumulatedTime` by the tick interval (1 second).
5. Send `CrownAccessState.Attempt` to the crown user with the current accumulated time.
6. If `CrownAccumulatedTime >= Configuration.CrownHoldTimeSeconds`:
   - Send `CrownAccessState.Success` to the crown user.
   - Call `ChangeWinnerGuild()`.

### 2. `ChangeWinnerGuild()` — In the same file

Called when a capture succeeds.

**Logic:**
1. Set `MiddleOwnerGuildId` to the crown user's guild ID.
2. Swap sides:
   - The capturing guild's side becomes `Defense`.
   - The previous `Defense` side becomes the capturing guild's old attack side.
3. Call `SetPlayerJoinSideAsync(killLifeStones: true)` to:
   - Re-assign all players to their new sides.
   - Kill all Life Stones.
4. Respawn all non-defense players to the attack respawn area.
5. Lock the crown (`IsCrownAvailable = false`).
6. Reset `CrownAccumulatedTime` to zero.
[...]

PR #869, opened 2026-08-07
Implement Castle Siege NPC lifecycle and management

### LINKED ISSUE #725: Castle Siege NPCs (Gates, Statues, Crown, Switches, Levers)
## Summary

Implement the Castle Siege NPC types: Castle Gates, Guardian Statues, Crown, Crown Switches, Gate Levers, and Siege Machines. This includes NPC classes, AI intelligence plug-ins, the upgrade/buy/repair system, and gate terrain blocking.

## Prerequisites

- #721 (Data Model) — for NPC definitions, upgrade definitions, `CastleSiegeNpcState`.
- #722 (State Machine) — for `CastleSiegeContext` and event state awareness.

## Background

OpenMU NPC hierarchy:
- `NonPlayerCharacter` — base NPC (not attackable).
- `AttackableNpcBase : NonPlayerCharacter` — attackable, has `IAttributeSystem`, `MagicEffectsList`.
- `Monster : AttackableNpcBase` — full AI with movement, attack, drops.
- `Destructible : AttackableNpcBase` — attackable but no AI/movement/attack.

NPC AI is provided via `INpcIntelligence` implementations wired through `PlugInManager`.

The Valley of Loren map (`Persistence/Initialization/VersionSeasonSix/Maps/ValleyOfLoren.cs`) already defines spawn positions for Crown (NPC 216), Crown Switch 1 (217), Crown Switch 2 (218), Guard (220), Sinior (223), Guardsman (224).

#### `CastleSiegeGate : AttackableNpcBase`
- Monster ID 277.
- Attackable, no movement, no attack ability.
- Tracks upgrade levels: `DefenseLevel`, `LifeLevel`.
- Can block terrain (6×2 tile area) when closed.
- State: open or closed.
- Override `OnDiedAsync` to remove terrain blocking.

#### `CastleSiegeStatue : AttackableNpcBase`
- Monster ID 283.
- Attackable, no movement, no attack ability.
- Tracks upgrade levels: `DefenseLevel`, `LifeLevel`, `RegenLevel`.
- Regenerates HP based on `RegenLevel` (on AI tick).

#### `CastleSiegeCrown : NonPlayerCharacter`
- Monster ID 219.
- Non-attackable. Interactive.
- Tracks: current crown user, accumulated time, availability state.
- States: idle (0), locked (1), captured (2).

#### `CastleSiegeSwitch : NonPlayerCharacter`
- Two instances (ID 0 and ID 1).
- Non-attackable. Interactive.
- Tracks: current player standing on the switch.

#### `CastleSiegeLever : NonPlayerCharacter`
- Interactive NPC.
- Associates with a specific gate.
- On interaction: toggles the associated gate open/closed.

#### `CastleSiegeMachine : NonPlayerCharacter`
- Monster IDs 221 (attack) and 222 (defense).
- Interactive. Non-attackable.
- Tracks: current operator (Player), active state.

### 2. NPC Intelligence Plug-ins — **New files in `GameLogic/CastleSiege/Intelligence/`**

Each implements `INpcIntelligence`:

#### `CastleSiegeGateIntelligence`
- No autonomous behavior. Gate just exists and takes damage.
- On death: remove terrain blocking, update context.

#### `CastleSiegeStatueIntelligence`
- Regeneration tick: if `RegenLevel > 0` and `IsAlive`, restore HP periodically.
- Rate depends on regen level (from `CastleSiegeConfiguration.StatueRegenUpgrades`).

#### `CastleSiegeCrownIntelligence`
- Track player proximity / interaction.
- If a player "stands on" the crown → set as crown user in `CastleSiegeContext`.
- If the player leaves/dies → clear crown user, send `CrownAccessState.Fail`.
- Accumulate time while player is on crown.

#### `CastleSiegeSwitchIntelligence`
- Track player proximity / interaction.
- If a player stands on the switch → set as switch user in `CastleSiegeContext`.
- If the player leaves/dies → clear switch user.
- Report state changes to context for `SendSwitchInfo()`.

#### `CastleSiegeLeverIntelligence`
- No autonomous behavior.
- Responds to player interaction (talk/use) by toggling the associated gate.

#### `CastleSiegeMachineIntelligence`
- No autonomous behavior.
- Responds to player commands (Phase 8).

#### `CastleSiegeNpcUpgradeAction` — **New file: `GameLogic/CastleSiege/Actions/CastleSiegeNpcUpgradeAction.cs`**

**Validation:**
- Player is castle owner guild's master or `AssistantMaster`.
- Current state is NOT `Start` (can't upgrade during battle).
- Target NPC is alive and active.
- Current level < max level (3) for the requested upgrade type.
- Player has enough Zen and Jewels of Guardian (consume from inventory).

**Effect:**
- Increment the upgrade level on the NPC runtime and persisted state.
- Apply the stat change:
  - **Defense**: Set monster's defense attribute to the value from `GateDefenseUpgrades` / `StatueDefenseUpgrades`.
  - **Life**: Set monster's max HP. Increase current HP by the difference.
  - **Regen**: Increment regen level (used by statue AI).
- Persist `CastleSiegeNpcState`.
- Send result to player.

#### `CastleSiegeNpcBuyAction` — **New file: `GameLogic/CastleSiege/Actions/CastleSiegeNpcBuyAction.cs`**

Re-purchase a destroyed NPC:
- Only castle owner guild master or assistant.
- NPC must not currently be alive.
- Player must have enough Zen (`GateBuyPrice` or `StatueBuyPrice`).
- Reset upgrade levels to 0. Set initial HP.
- Spawn the NPC on the map.

#### `CastleSiegeNpcRepairAction` — **New file: `GameLogic/CastleSiege/Actions/CastleSiegeNpcRepairAction.cs`**

Restore NPC to full HP:
- Cost formula for Gate (277): `(MaxHP - CurrentHP) * 5 + defenseLevel * 1,000,000`.
- Cost formula for Statue (283): `(MaxHP - CurrentHP) * 3 + (defenseLevel + regenLevel) * 1,000,000`.
- Cannot repair during active battle.

#### `CastleSiegeGateOperateAction` — **New file: `GameLogic/CastleSiege/Actions/CastleSiegeGateOperateAction.cs`**

When a gate is closed (`CloseCastleGate`):
- Apply a terrain lock attribute to a 6×2 tile area around the gate position in `GameMapTerrain`.
- This blocks player movement through the gate.
- Method: use `GameMapTerrain`'s attribute bytes (similar to safezone flags).

When a gate is opened (via lever interaction):
- Remove the terrain lock attribute from the same area.

**Validation for gate operation:**
- During `Start` state: only `Defense` side can operate.
- Outside `Start` state: only castle owner or alliance member.

### 5. NPC Lifecycle & Spawning

In `CastleSiegePlugIn`:

**On `Ready` enter:**
1. Load `CastleSiegeNpcState` records from DB.
2. For each DB-persisted NPC definition in `CastleSiegeConfiguration.NpcDefinitions`:
   - If a saved state exists: apply upgrade levels and HP.
   - Spawn the NPC on the Castle Siege map.
   - Store reference in `CastleSiegeContext.ActiveNpcs`.
3. Close all gates and apply terrain blocking.

**On `Start` enter:**
- Spawn non-DB NPCs (machines) at their configured positions.

**On `End` enter:**
- Save all DB-persisted NPC states.
- Despawn all non-DB NPCs (machines).

**Every 120 seconds (outside `Start`):**
- Call `VerifyNpcExist()` to sync runtime state with actual monster state.
- Call `SaveNpcStatesAsync()` to persist.

### 6. NPC List Packet Handlers

- **`CastleSiegeGateListHandlerPlugIn`** — handles `C1-B3-01`. Returns list of gate NPCs with upgrade levels and HP.
- **`CastleSiegeStatueListHandlerPlugIn`** — handles `C1-B3-02`. Returns list of statue NPCs.

### 7. Message Handlers for Upgrade/Buy/Repair

- `CastleSiegeDefenseBuyHandlerPlugIn` — `C1-B2-05`
- `CastleSiegeDefenseRepairHandlerPlugIn` — `C1-B2-06`
- `CastleSiegeDefenseUpgradeHandlerPlugIn` — `C1-B2-07`
- `CastleSiegeGateOperateHandlerPlugIn` — `C1-B2-12`

## Files to Create

| File | Description |
|------|-------------|
| `GameLogic/CastleSiege/NPC/CastleSiegeGate.cs` | Gate NPC class |
| `GameLogic/CastleSiege/NPC/CastleSiegeStatue.cs` | Statue NPC class |
| `GameLogic/CastleSiege/NPC/CastleSiegeCrown.cs` | Crown NPC class |
| `GameLogic/CastleSiege/NPC/CastleSiegeSwitch.cs` | Switch NPC class |
| `GameLogic/CastleSiege/NPC/CastleSiegeLever.cs` | Lever NPC class |
| `GameLogic/CastleSiege/NPC/CastleSiegeMachine.cs` | Machine NPC class |
| `GameLogic/CastleSiege/Intelligence/CastleSiegeGateIntelligence.cs` | Gate AI |
| `GameLogic/CastleSiege/Intelligence/CastleSiegeStatueIntelligence.cs` | Statue AI |
| `GameLogic/CastleSiege/Intelligence/CastleSiegeCrownIntelligence.cs` | Crown AI |
| `GameLogic/CastleSiege/Intelligence/CastleSiegeSwitchIntelligence.cs` | Switch AI |
| `GameLogic/CastleSiege/Intelligence/CastleSiegeLeverIntelligence.cs` | Lever AI |
| `GameLogic/CastleSiege/Intelligence/CastleSiegeMachineIntelligence.cs` | Machine AI |
| `GameLogic/CastleSiege/Actions/CastleSiegeNpcUpgradeAction.cs` | Upgrade logic |
| `GameLogic/CastleSiege/Actions/CastleSiegeNpcBuyAction.cs` | Buy logic |
| `GameLogic/CastleSiege/Actions/CastleSiegeNpcRepairAction.cs` | Repair logic |
| `GameLogic/CastleSiege/Actions/CastleSiegeGateOperateAction.cs` | Gate open/close |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeDefenseBuyHandlerPlugIn.cs` | Packet handler |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeDefenseRepairHandlerPlugIn.cs` | Packet handler |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeDefenseUpgradeHandlerPlugIn.cs` | Packet handler |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeGateOperateHandlerPlugIn.cs` | Packet handler |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeGateListHandlerPlugIn.cs` | Packet handler |
| `GameServer/MessageHandler/CastleSiege/CastleSiegeStatueListHandlerPlugIn.cs` | Packet handler |

## Acceptance Criteria

- [ ] Gates can be spawned, take damage, die, and be re-purchased.
- [ ] Statues regenerate HP based on regen upgrade level.
- [ ] NPCs can be upgraded (defense, life, regen) with correct cost validation.
- [ ] Gate terrain blocking prevents player movement through closed gates.
- [ ] Gate levers correctly toggle gates open/closed with proper side validation.
- [ ] Crown tracks the current user and accumulated time.
- [ ] Switches track occupant players.
- [ ] NPC state persists between server restarts (upgrade levels, HP).
- [ ] Non-DB NPCs (machines) only exist during the `Start` state.

Zylkien:
## Summary

Implements the Castle Siege NPC lifecycle and management system described in #725.

- Adds Castle Gates, Guardian Statues, Crown, Crown Switches, Gate Levers, and battle-only Siege Machines.
- Adds the corresponding runtime intelligence and event lifecycle handling.
- Implements defense-structure purchase, repair, and upgrade operations.
- Restores persisted NPC health and upgrade levels after server restarts.
- Synchronizes NPC state periodically outside the active battle.
- Adds the existing B2/B3 packet handlers and RemoteView serializers required by the NPC interfaces.
- Adds localized plug-in metadata for the new handlers and views.

## Gate behavior

- Closed gates apply server-side terrain blocking across their configured area.
- Opening or destroying a gate removes the dynamic terrain restriction.
- Originally blocked terrain remains protected when the dynamic gate restriction is removed.
- Players entering the map receive the current gate state.
- Movement validation checks the complete walking path, preventing players from skipping across a blocked gate area.
- Gate lever operations validate the player’s Castle Siege side or castle ownership before changing the gate state.

## Defense structures

- Gates and Guardian Statues can be purchased, repaired, and upgraded outside the active battle.
- Authorization is limited to the castle owner guild master or assistant master.
- Upgrade requirements, Zen costs, and Jewel of Guardian consumption are validated before applying changes.
- Failed operations do not consume Zen or items.
- Guardian Statues regenerate health according to their configured regeneration level.
- Attack permissions are restricted by the current Castle Siege state and participant side.

## Lifecycle

- Persistent Castle Siege NPCs are restored during the `Ready` state.
- Gates are closed when the event is prepared.
- Siege Machines are spawned only during the `Start` state.
- Machine NPCs are removed and persistent NPC state is saved when the battle ends.
- Destroyed and currently unspawned defense structures remain represented in persistence.

## Crown mechanics boundary

This change tracks Crown and Crown Switch occupants and exposes the runtime state needed by the capture system. Capture-time accumulation, access result packets, ownership changes, and the final win condition remain part of #726.

The Crown uses monster number 216, matching the existing Valley of Loren configuration. Monster number 219 is the Gate Lever; the Crown number in the issue requirements appears to be a typo.

## Testing

- Added lifecycle, authorization, terrain, persistence, purchase, repair, upgrade, regeneration, Crown, switch, lever, and failure-side-effect regression tests.
- Added serialization coverage for NPC lists, operation results, gate interfaces, gate operations, and gate state packets.
- Added startup scheduling and destroyed-state persistence tests.
- CI-style build: 0 warnings and 0 errors.
- Castle Siege focused tests: 42 passed.
- Full `MUnique.OpenMU.Tests` project: 701 passed.
- StyleCop and analyzer checks pass for the changed scope.

Closes #725

### DIFF
--- a/src/DataModel/Configuration/CastleSiegeConfiguration.cs
+++ b/src/DataModel/Configuration/CastleSiegeConfiguration.cs
@@ -74,6 +74,21 @@ public partial class CastleSiegeConfiguration
     public int StatueBuyPrice { get; set; }
 
     /// <summary>
+    /// Gets or sets the gate repair cost in Zen per restored health point.
+    /// </summary>
+    public int GateRepairCostPerHealthPoint { get; set; } = 5;
+
+    /// <summary>
+    /// Gets or sets the Guardian Statue repair cost in Zen per restored health point.
+    /// </summary>
+    public int StatueRepairCostPerHealthPoint { get; set; } = 3;
+
+    /// <summary>
+    /// Gets or sets the additional repair cost in Zen per applied defense or regeneration level.
+    /// </summary>
+    public int RepairCostPerUpgradeLevel { get; set; } = 1_000_000;
+
+    /// <summary>
     /// Gets or sets the map definition for the Valley of Loren (map 30), where the siege takes place.
     /// </summary>
     public virtual GameMapDefinition? CastleSiegeMapDefinition { get; set; }
--- a/src/GameLogic/CastleSiege/Actions/CastleSiegeGateOperateAction.cs
+++ b/src/GameLogic/CastleSiege/Actions/CastleSiegeGateOperateAction.cs
@@ -1,0 +1,129 @@
+// <copyright file="CastleSiegeGateOperateAction.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic.CastleSiege.Actions;
+
+using MUnique.OpenMU.DataModel.Configuration;
+using MUnique.OpenMU.GameLogic.Views.CastleSiege;
+
+/// <summary>
+/// Opens and closes Castle Siege gates after validating the requesting player's side.
+/// </summary>
+public static class CastleSiegeGateOperateAction
+{
+    /// <summary>
+    /// Tries to open the Castle Siege gate-operation interface.
+    /// </summary>
+    /// <param name="player">The requesting player.</param>
+    /// <param name="context">The Castle Siege context.</param>
+    /// <param name="gateId">The target gate identifier.</param>
+    /// <returns>A task that represents the asynchronous operation.</returns>
+    public static async ValueTask ShowInterfaceAsync(
+        Player player,
+        CastleSiegeContext? context,
+        ushort gateId)
+    {
+        var result = await ValidateAsync(player, context, gateId).ConfigureAwait(false);
+        await player.InvokeViewPlugInAsync<ICastleSiegeNpcOperationResultPlugIn>(
+                view => view.ShowGateInterfaceAsync(result, gateId))
+            .ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Tries to open or close a Castle Siege gate.
+    /// </summary>
+    /// <param name="player">The requesting player.</param>
+    /// <param name="context">The Castle Siege context.</param>
+    /// <param name="gateId">The target gate identifier.</param>
+    /// <param name="open">Whether the gate should be opened.</param>
+    /// <returns>The operation result.</returns>
+    public static async ValueTask<CastleSiegeNpcOperationResult> OperateAsync(
+        Player player,
+        CastleSiegeContext? context,
+        ushort gateId,
+        bool open)
+    {
+        var (result, isOpen) = await OperateCoreAsync(player, context, gateId, open).ConfigureAwait(false);
+        await player.InvokeViewPlugInAsync<ICastleSiegeNpcOperationResultPlugIn>(
+                view => view.ShowGateOperationResultAsync(result, isOpen, gateId))
+            .ConfigureAwait(false);
+        return result;
+    }
+
+    private static async ValueTask<(CastleSiegeNpcOperationResult Result, bool IsOpen)> OperateCoreAsync(
+        Player player,
+        CastleSiegeContext? context,
+        ushort gateId,
+        bool open)
+    {
+        if (context is not { Configuration.Enabled: true })
+        {
+            return (CastleSiegeNpcOperationResult.Failed, open);
+        }
+
+        await context.ExecutionLock.WaitAsync().ConfigureAwait(false);
+        try
+        {
+            if (!await IsAuthorizedAsync(player, context).ConfigureAwait(false))
+            {
+                return (CastleSiegeNpcOperationResult.NotAuthorized, open);
+            }
+
+            if (context.NpcController.FindGate(gateId) is not { IsAlive: true } gate)
+            {
+                return (CastleSiegeNpcOperationResult.Failed, open);
+            }
+
+            if (open)
+            {
+                await gate.OpenAsync().ConfigureAwait(false);
+            }
+            else
+            {
+                await gate.CloseAsync().ConfigureAwait(false);
+            }
+
+            return (CastleSiegeNpcOperationResult.Success, !gate.IsClosed);
+        }
+        finally
+        {
+            context.ExecutionLock.Release();
+        }
+    }
+
+    private static async ValueTask<CastleSiegeNpcOperationResult> ValidateAsync(
+        Player player,
+        CastleSiegeContext? context,
+        ushort gateId)
+    {
+        if (context is not { Configuration.Enabled: true })
+        {
+            return CastleSiegeNpcOperationResult.Failed;
+        }
+
+        await context.ExecutionLock.WaitAsync().ConfigureAwait(false);
+        try
+        {
+            if (!await IsAuthorizedAsync(player, context).ConfigureAwait(false))
+            {
+                return CastleSiegeNpcOperationResult.NotAuthorized;
+            }
+
+            return context.NpcController.FindGate(gateId) is { IsAlive: true }
+                ? CastleSiegeNpcOperationResult.Success
+                : CastleSiegeNpcOperationResult.Failed;
+        }
+        finally
+        {
+            context.ExecutionLock.Release();
+        }
+    }
+
+    private static ValueTask<bool> IsAuthorizedAsync(Player player, CastleSiegeContext context)
+    {
+        return context.CurrentState == CastleSiegeState.Start
+            ? ValueTask.FromResult(context.GetPlayerJoinSide(player) == CastleSiegeJoinSide.Defense)
+            : CastleSiegeNpcAuthorization.IsOwnerAllianceMemberAsync(player, context);
+    }
+}
--- a/src/GameLogic/CastleSiege/Actions/CastleSiegeNpcAuthorization.cs
+++ b/src/GameLogic/CastleSiege/Actions/CastleSiegeNpcAuthorization.cs
@@ -1,0 +1,46 @@
+// <copyright file="CastleSiegeNpcAuthorization.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic.CastleSiege.Actions;
+
+using MUnique.OpenMU.Interfaces;
+
+/// <summary>
+/// Authorization checks shared by Castle Siege NPC management actions.
+/// </summary>
+internal static class CastleSiegeNpcAuthorization
+{
+    /// <summary>
+    /// Checks whether a player is a guild master or assistant master of the castle-owner guild.
+    /// </summary>
+    /// <param name="player">The player to check.</param>
+    /// <param name="context">The Castle Siege context.</param>
+    /// <returns><see langword="true"/> when the player is an authorized owner-guild officer.</returns>
+    public static async ValueTask<bool> IsOwnerOfficerAsync(Player player, CastleSiegeContext context)
+    {
+        if (player.GuildStatus is not { } guildStatus
+            || guildStatus.Position is not (GuildPosition.GuildMaster or GuildPosition.AssistantMaster)
+            || player.GameContext is not IGameServerContext gameServerContext
+            || await gameServerContext.GuildServer
+                .GetPersistentGuildIdAsync(guildStatus.GuildId)
+                .ConfigureAwait(false) is not { } persistentGuildId)
+        {
+            return false;
+        }
+
+        return context.SiegeData.OwnerGuildId == persistentGuildId;
+    }
+
+    /// <summary>
+    /// Checks whether a player belongs to the castle-owner guild or one of its alliance guilds.
+    /// </summary>
+    /// <param name="player">The player to check.</param>
+    /// <param name="context">The Castle Siege context.</param>
+    /// <returns><see langword="true"/> when the player belongs to the owner alliance.</returns>
+    public static async ValueTask<bool> IsOwnerAllianceMemberAsync(Player player, CastleSiegeContext context)
+    {
+        var guildId = await CastleSiegeGuildResolver.ResolveRegistrationGuildIdAsync(player).ConfigureAwait(false);
+        return guildId is not null && context.SiegeData.OwnerGuildId == guildId;
+    }
+}
--- a/src/GameLogic/CastleSiege/Actions/CastleSiegeNpcBuyAction.cs
+++ b/src/GameLogic/CastleSiege/Actions/CastleSiegeNpcBuyAction.cs
@@ -70,10 +70,10 @@ public static class CastleSiegeNpcBuyAction
 
             var (price, initialHealth) = runtime.Definition.MonsterDefinition?.Number switch
             {
-                CastleSiegeGate.MonsterNumber => (
+                var number when number == CastleSiegeGate.MonsterNumber => (
                     context.Configuration.GateBuyPrice,
                     GetInitialValue(context.Configuration.GateLifeUpgrades)),
-                CastleSiegeStatue.MonsterNumber => (
+                var number when number == CastleSiegeStatue.MonsterNumber => (
                     context.Configuration.StatueBuyPrice,
                     GetInitialValue(context.Configuration.StatueLifeUpgrades)),
                 _ => (0, 0),
--- a/src/GameLogic/CastleSiege/Actions/CastleSiegeNpcBuyAction.cs
+++ b/src/GameLogic/CastleSiege/Actions/CastleSiegeNpcBuyAction.cs
@@ -1,0 +1,114 @@
+// <copyright file="CastleSiegeNpcBuyAction.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.GameLogic.CastleSiege.Actions;
+
+using MUnique.OpenMU.GameLogic.CastleSiege.NPC;
+using MUnique.OpenMU.GameLogic.Views.CastleSiege;
+
+/// <summary>
+/// Re-purchases destroyed Castle Siege gates and Guardian Statues.
+/// </summary>
+public static class CastleSiegeNpcBuyAction
+{
+    /// <summary>
+    /// Tries to re-purchase a destroyed defense structure.
+    /// </summary>
+    /// <param name="player">The requesting player.</param>
+    /// <param name="context">The Castle Siege context.</param>
+    /// <param name="npcNumber">The target NPC number.</param>
+    /// <param name="npcIndex">The target NPC instance identifier.</param>
+    /// <returns>The operation result.</returns>
+    public static async ValueTask<CastleSiegeNpcOperationResult> BuyAsync(
+        Player player,
+        CastleSiegeContext? context,
+        uint npcNumber,
+        uint npcIndex)
+    {
+        var result = await BuyCoreAsync(player, context, npcNumber, npcIndex).ConfigureAwait(false);
+        await player.InvokeViewPlugInAsync<ICastleSiegeNpcOperationResultPlugIn>(
+                view => view.ShowBuyResultAsync(result, npcNumber, npcIndex))
+            .ConfigureAwait(false);
+        return result;
+    }
+
+    private static async ValueTask<CastleSiegeNpcOperationResult> BuyCoreAsync(
+        Player player,
+        CastleSiegeContext? context,
+        uint npcNumber,
+        uint npcIndex)
+    {
+        if (context is not { Configuration.Enabled: true })
+        {
+            return CastleSiegeNpcOperationResult.Failed;
+        }
+
+        await context.ExecutionLock.WaitAsync().ConfigureAwait(false);
+        try
+        {
+            if (context.CurrentState == CastleSiegeState.Start)
+            {
+                return CastleSiegeNpcOperationResult.Failed;
+            }
+
+            if (!await CastleSiegeNpcAuthorization.IsOwnerOfficerAsync(player, context).ConfigureAwait(false))
+            {
+                return CastleSiegeNpcOperationResult.NotAuthorized;
+            }
+
+            var runtime = context.NpcController.FindDefenseStructure(npcNumber, npcIndex);
+            if (runtime is { IsAlive: true })
+            {
+                return CastleSiegeNpcOperationResult.RequirementNotMet;
+            }
+
+            if (runtime is not { PersistedState: { } state })
+            {
+                return CastleSiegeNpcOperationResult.Failed;
+            }
+
+            var (price, initialHealth) = runtime.Definition.MonsterDefinition?.Number switch
+            {
+                CastleSiegeGate.MonsterNumber => (
+                    context.Configuration.GateBuyPrice,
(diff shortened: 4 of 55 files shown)


sven-n:
## Review

Reviewed the 49 changed files (`a9732ec..f908adf`, +4449/-16). Overall this is well-structured work: the `CanBeAttackedBy` hook on `AttackableNpcBase` is a clean extension point, the poison/bleed/reflect no-op overrides correctly close the alternate damage paths, `ExecutionLock` discipline is consistent (`WaitAsync(0)` in the periodic task, blocking waits in the actions), and `DespawnAsync` opening the gate before removal correctly avoids leaking a blocked walkmap. Test coverage of authorization, failure side-effects, restart persistence and terrain restore is good, and CI is green.

Findings below, roughly by severity.

---

### 1. Guardian Statue regeneration races with combat damage (lost damage)

`src/GameLogic/CastleSiege/NPC/CastleSiegeStatue.cs:85`

```csharp
this.Health = Math.Min(this.MaximumHealth, this.Health + restored);
```

This is a plain read-modify-write on `Health`, while `AttackableNpcBase.TryHit` applies damage with `Interlocked.Add(ref this._health, -(int)damage)`. `CastleSiegeStatueIntelligence` fires this from a `Timer` thread every 5 s and `Regenerate()` is not gated on siege state, so it runs during `Start` while attackers are hitting the statue.

Any damage that lands between the regen tick's read of `this.Health` and its write is silently discarded. Concretely: statue at 1,000,000 HP, tick reads 1,000,000 and computes 1,050,000; a party lands 300,000 damage (`_health` → 700,000); the tick then writes 1,050,000. The 300,000 damage is gone. With many attackers on a statue this is repeatable every 5 s, so a statue can absorb substantially more than its configured HP.

Suggest applying the regeneration through `Interlocked` (CAS loop) or routing it through the same path combat uses. `SetMaximumHealth` / `RestoreFullHealth` have the same shape but run under `ExecutionLock` outside battle, so they're much lower risk.

### 2. Gates stay open if the server starts up during the `Start` state

`src/GameLogic/CastleSiege/CastleSiegePlugIn.cs:258-266`

`CloseGatesAsync()` is only called from the `Ready` branch. On startup the flow is `PeriodicTaskExecute` → `InitializeAsync` → `OnEnterStateAsync(context, isStartup: true)`, so if the server is restarted mid-battle and the current period is `Start`, only `PrepareAsync()` + `SpawnMachinesAsync()` run. Every gate then spawns open, no terrain blocking is applied, and attackers walk straight into the castle for the remainder of the siege.

Gate open/closed state also isn't persisted, so there's nothing to restore from either. Closing gates inside `PrepareAsync` (or in the `Start` branch, at least when `isStartup`) would cover this.

### 3. `Player.WalkToAsync` path validation is a server-wide behavior change

`src/GameLogic/Player.cs:1391` and `CanWalkPath` at `:2047`

Previously only the target tile was validated; now a single non-walkable intermediate step rejects the whole walk, rubber-bands the player, and emits a `LogWarning`. This applies to every player on every map, not just the Castle Siege map — any pre-existing walkmap/client `.att` discrepancy, or a client path that clips the corner of a blocked tile, now becomes a hard rubber-band plus a warning-level log on a hot path. The existing log text ("but it's not an allowed target") also becomes inaccurate when the *target* was fine and a mid-path step wasn't.

Suggest truncating the path at the first blocked step (walk to the last reachable tile) rather than rejecting it wholesale. That still prevents skipping across a closed gate, but leaves paths that merely graze a blocked tile working as before. At minimum this is worth calling out explicitly — it's the one change here with blast radius beyond Castle Siege — and the log message/level adjusted.

### 4. 1 Hz polling over the full server player list

`CastleSiegeCrownIntelligence.cs:65` and `CastleSiegeSwitchIntelligence.cs:59`

Both tick every second, per NPC (crown + 2 switches = 3×/s), and each tick calls `GameContext.GetPlayersAsync()`, which takes a reader lock and copies the entire server player list — only to filter down to players within 1 tile. `this.Npc.CurrentMap.GetAttackablesInRange(this.Npc.Position, 1).OfType<Player>()` answers the same question directly from the AoI buckets.

Related: nothing in production reads the results yet. `IsCrownAvailable` is only ever assigned `false` (at `EndCycle`), so `CastleSiegeCrown.State` is permanently `Locked`; `SwitchUsers`, `CastleSiegeSwitch.Occupant`, `CastleSiegeMachine.Operator`/`IsActive` are written but never read outside tests. I understand these are the #726 boundary — but if so, it may be simpler to start the polling timers in #726 alongside their consumer rather than run them now.

### 5. Gate broadcasts copy the player list twice per gate, while holding `ExecutionLock`

`CastleSiegeGate.cs:188` and `:197`

`NotifyTerrainChangeAsync` and `NotifyGateStateAsync` each call `GetPlayersAsync()` and then `await` one view invocation per player sequentially. `CloseGatesAsync` walks all six configured gates, so a single state transition does 12 full player-list copies and 12×N sequential sends — all under `ExecutionLock`, which simultaneously blocks the state machine and every `ObjectAddedToMapAsync` map entry (which also takes the lock). `MiniGameContext.ForEachPlayerAsync` shows the fan-out-then-`WhenAll` pattern; since the plugin already hooks `IObjectAddedToMapPlugIn`, it could maintain its own set of players on the siege map and broadcast once per transition instead of once per gate.

### 6. Repair cost is hardcoded while everything else is configuration-driven

`CastleSiegeNpcRepairAction.cs:91,93`

```csharp
(missingHealth * 5) + ((long)state.DefenseLevel * 1_000_000),
(missingHealth * 3) + ((long)(state.DefenseLevel + state.RegenLevel) * 1_000_000),
```

`GateBuyPrice`, `StatueBuyPrice` and every upgrade's `RequiredZen` live in `CastleSiegeConfiguration`, so server operators can retune buying and upgrading but not repairing. Worth moving these multipliers into the configuration and seeding them in `CastleSiegeInitializer`.

### 7. Gate terrain area is a hardcoded 6×2 block

`CastleSiegeGate.GetTerrainArea()` (`:175`) hardcodes width 6 / height 2, centres on X but not on Y (`startY = Position.Y`), and ignores `CastleSiegeNpcDefinition.Direction`. It fits the six seeded Valley of Loren gates, which are all horizontal, but a rotated or differently-sized gate can't be expressed. The X/Y asymmetry in particular deserves a comment, or ideally per-definition configuration.

Also, if two closed gates' areas ever overlap, the second gate's `_originalTerrain` records the *first* gate's already-blocked value, and opening the first gate will unblock tiles the second still blocks. Not reachable with the current coordinates, but easy to trip over when adding a gate.

---

### Simplification (no behavior change)

**Monster numbers as `const`.** `MonsterNumber` / `FirstMonsterNumber` / `AttackMonsterNumber` etc. are `public static short { get; }` auto-properties, which aren't constants — that's what forces the `var number when number == X` switch workaround, 15 occurrences across `CastleSiegeNpcController`, `CastleSiegeNpcBuyAction`, `CastleSiegeNpcRepairAction` and `CastleSiegeNpcUpgradeAction`. As `public const short` they become plain `case CastleSiegeGate.MonsterNumber:` labels.

**Duplicated upgrade lookups.** The "which upgrade collection for this monster number" mapping appears five times (`CastleSiegeNpcController.GetMaximumHealth`, `.GetInitialHealth`, `CastleSiegeNpcBuyAction`, `CastleSiegeNpcUpgradeAction.GetDefinitions`, and `ApplyPersistedUpgrades` in both gate and statue), and the `definitions.FirstOrDefault(d => d.Level == level)?.Value` lookup is copy-pasted as a private `GetValue` in `CastleSiegeGate` and `CastleSiegeStatue` plus twice more in the controller. One helper on `CastleSiegeConfiguration` (e.g. `GetUpgrades(monsterNumber, upgradeType)`) plus one `GetValue(this IEnumerable<CastleSiegeUpgradeDefinition>, byte level)` extension would collapse all of it.

**Redundant casts.** 8× `checked((byte)definition.SpawnX/SpawnY)` — `CastleSiegeNpcDefinition.SpawnX`/`SpawnY` are already `byte` (`CastleSiegeNpcController` ×6, `CreateInfo` ×2).

### Smaller notes

- `FindDefenseStructure` / `FindGate` match on either the configured `InstanceId` **or** the live map object `Id`. This is unambiguous today (`InstanceId` is a `byte`; map object ids start at `ConstantPlayerId + 1` = 0x0201), but the client only ever echoes what the server sent it, so accepting both widens the input surface for no gain. Worth a comment explaining why the dual match is safe, or narrowing it.
- `CastleSiegeStatue.Regenerate()` and `RestoreFullHealth()` change health without notifying observers, so clients keep showing a stale HP bar until the next NPC-list request. Intentional?
- Re: Crown = 216, lever = 219 in the description — the reasoning holds against the existing Valley of Loren monster table, but flagging it so it lands as an explicit maintainer decision rather than an implicit one.

---

sven-n:
## Re-review (`f908adf` → `31d76cf`)

Went through the new commit. All seven substantive points from the previous round are addressed, and I verified each rather than taking the diff at face value. CI is green on the new head.

**Confirmed fixed:**

- **Regen race** — `AttackableNpcBase.RestoreHealth` is a correct CAS loop: it bails at `currentHealth <= 0` so it can't resurrect a dead structure, clamps to `maximumHealth - currentHealth`, and retries on contention. `Health` is now `Volatile.Read` / `Interlocked.Exchange`, which makes it consistent with `TryHit`'s `Interlocked.Add`. Note this one touches every NPC in the game, not just siege structures — it's correctness-positive and the cost is negligible, but worth being aware of.
- **Gates on startup** — `CloseGatesAsync()` added to the `Start` branch, with `ServerStartupDuringBattleClosesGatesAsync` covering it.
- **Overlapping gate terrain** — the reference-counted `_gateTerrain` is right; I traced the A-closes / B-closes / A-opens / B-opens sequence and the tile stays blocked exactly as long as it should, and `blockedAreas` correctly re-reports both the still-covered and the originally-blocked tiles.
- **Repair costs** — now configuration-driven, and the migration's `defaultValue: 5 / 3 / 1000000` means existing installs keep the current behavior. Initializer covers fresh installs.
- **Upgrade lookups consolidated** into `GetUpgrades`/`GetValue`; all 8 redundant `checked((byte)…)` casts gone (0 remaining).
- **Gate broadcasts** — `ForEachSiegePlayerAsync` replaces the per-gate full player-list copy. Worth noting for reassurance: `InvokeViewPlugInAsync` catches and logs per-player, so one broken connection can't abort a gate broadcast through the new `Task.WhenAll`.
- The explanatory comments on `GetTerrainArea` and `FindDefenseStructure` answer what I was asking for.

Four new things, one of which I think is a real regression introduced by the fix.

---

### 1. Crown/switch occupancy area silently widened from 5 tiles to 9

`CastleSiegeCrownIntelligence.cs` / `CastleSiegeSwitchIntelligence.cs`

The old filter was `player.GetDistanceTo(crown.Position) <= 1`, which is **Euclidean** (`LocateableExtensions.GetDistanceTo` → `EuclideanDistanceTo`) — at range 1 that admits only the 5 tiles `(0,0), (±1,0), (0,±1)`; the diagonals are √2 ≈ 1.414 and were excluded.

`GetAttackablesInRange(point, 1)` filters with `ILocateable.IsInRange`, which is **Chebyshev** (`xdiff <= range && ydiff <= range`) — the full 3×3, so the 4 diagonal tiles are now included.

The switch to a map-scoped query was the right call, but it changed the gameplay footprint as a side effect. A 3×3 is arguably the more sensible "standing at the switch" rule, so this may well be what you want — but it should be a deliberate choice, not a by-product. If the 5-tile shape was intended, re-adding `player.GetDistanceTo(...) <= 1` after the range query keeps both the performance win and the original semantics.

While you're there: the `.Where(player => player.CurrentMap == crown.CurrentMap)` that survived the refactor is now dead — the query is issued on `crown.CurrentMap`, so every result is on that map by construction. Same in the switch intelligence.

### 2. `WalkToAsync` no longer validates `target` at all

`GetWalkableStepCount` only inspects `steps[i].To`; the `target` parameter is used unchecked when the whole path is walkable. That's safe for the only production caller — `CharacterWalkBaseHandlerPlugIn.GetTarget` returns `steps[^1].To`, and `WalkToAsync` returns early on empty steps — so `target` always equals the last step. But `WalkToAsync` is `public`, and that invariant is now implicit where it used to be enforced by the explicit `WalkMap[target.X, target.Y]` check.

Simplest fix: derive it unconditionally rather than only in the truncation branch — `target = steps.Span[^1].To;` — which removes the possibility of the parameter disagreeing with the path.

### 3. A truncated walk sends no resync, so the client can stay desynced

The new test asserts `ObjectMovedAsync(player, MoveType.Instant)` is called `Times.Never` on truncation, so this is deliberate — flagging it to confirm the trade-off is understood.

The server stops the player at the last reachable tile, but the client is authoritative for its own movement and keeps walking to the original target locally, and nothing corrects it. The only recovery is the `MaxAllowedWalkStartOffset` check on the *next* walk request, which tolerates 5 tiles. So a player who paths across a closed gate can sit up to 5 tiles out of sync with the server until they issue a walk that exceeds that threshold — and a gate area is 6×2, which is right at the boundary.

Truncation is still better than the wholesale rejection, but sending the `IObjectMovedPlugIn` correction once the truncated walk completes would close the gap.

### 4. `_siegeMapPlayers` is unbounded and depends on the removal path firing

Defensive rather than a confirmed leak — I traced it and the normal paths do fire `ObjectRemoved`. But the dictionary holds strong `Player` keys for the context's lifetime, and the cleanup depends on `GameContext.RemovePlayerAsync:363`, which calls `player.CurrentMap?.RemoveAsync(player)` **without awaiting it** (pre-existing, not this PR). Any failure in that un-awaited path leaves a `Player` — and its whole object graph — pinned for the life of the server.

`ForEachSiegePlayerAsync` already computes the "is this player still on the siege map" predicate, so pruning the non-matching keys while it iterates costs nothing and removes the dependency on the removal path entirely.

---

### Minor

- `BlockGateTerrain` / `ReleaseGateTerrain` take a `map` parameter, but `_gateTerrain` is keyed by `Point` alone. Correct today (one siege map per context), though the signature implies a multi-map safety that isn't there. Either drop the parameter or include the map in the key.
- The updated walk test uses `await Task.Delay(TimeSpan.FromMilliseconds(500))` to wait for the background walker. That's timing-dependent and will be the first thing to go flaky on a loaded CI agent — polling `Player.Position` until it reaches the expected tile with a timeout would be sturdier.
- `MonsterNumber` and friends are still `public static short { get; }` rather than `const`, so the `var number when number == X` workaround remains — 16 occurrences now (was 15), in `CastleSiegeConfigurationExtensions.GetUpgrades`, `CastleSiegeNpcBuyAction`, `CastleSiegeNpcRepairAction` and `CastleSiegeNpcController`. The lookup consolidation was the bigger half of that point, so this is cosmetic at this stage.

Nothing here blocks from my side apart from wanting a decision on the crown/switch range change.

---

PR #873, opened 2026-08-10
Prevent duplicate map entry on repeated F3 12 packet

mrallanodreman:
Hi @sven-n — first, thank you for the detailed review on #867. The time you took
to explain the *why*, not just the *what*, is what let me step back and split the
useful parts out cleanly instead of pushing one big lump. This is the first of the
two pieces you said you'd take, on its own on top of current master.

## Problem

A repeated `F3 12` ("client ready after map change") packet re-runs
`ClientReadyAfterMapChangeAsync`, which unconditionally re-adds the player — and,
if present, its summon (`CurrentMap.AddAsync(summon)` + `summon.OnSpawn()`) — to
the current map. You suspected this in the review; I've confirmed it on master.

## Fix

Return early when `CurrentMap` is already set on entry. During a regular map
change `CurrentMap` is set to `null` (see the `// Will be set again, when the
client acknowledged the map change by F3 12 packet.` spots) and only reassigned
inside this handler, so a non-null value on entry is a reliable "this handler
already ran for the current map change" signal.

## Notes

- One file, +10 lines. No behaviour change for the normal single-packet path.
- Independent of the large-map work — nothing else from #867 rides along.
- Rebased on current `master`.

### DIFF
--- a/src/GameLogic/Player.cs
+++ b/src/GameLogic/Player.cs
@@ -949,6 +949,16 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
         this.ThrowNotInitializedProperty(this.SelectedCharacter is null, nameof(this.SelectedCharacter));
         this.SelectedCharacter.ThrowNotInitializedProperty(this.SelectedCharacter.CurrentMap is null, nameof(this.SelectedCharacter.CurrentMap));
 
+        if (this.CurrentMap is not null)
+        {
+            // Guard against a repeated F3 12 (client ready after map change) packet:
+            // during a regular map change CurrentMap is set to null and only assigned
+            // here, so a non-null value means this handler already ran for the current
+            // map change. Without this, a duplicate packet would add the player (and its
+            // summon) to the map a second time.
+            return;
+        }
+
         if (this.CurrentMiniGame is { } currentMiniGame)
         {
             this.CurrentMap = currentMiniGame.Map;
--- a/src/GameLogic/Player.cs
+++ b/src/GameLogic/Player.cs
@@ -951,11 +951,14 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
 
         if (this.CurrentMap is not null)
         {
-            // Guard against a repeated F3 12 (client ready after map change) packet:
-            // during a regular map change CurrentMap is set to null and only assigned
-            // here, so a non-null value means this handler already ran for the current
-            // map change. Without this, a duplicate packet would add the player (and its
-            // summon) to the map a second time.
+            // Guard against a repeated F3 12 (client ready after map change) packet.
+            // A map change usually leaves CurrentMap null until this handler assigns it,
+            // so a non-null value means the handler already ran. The exception is the
+            // IRespawnAfterDeathPlugIn branch of RespawnAtAsync, which assigns CurrentMap
+            // and adds the player itself; a trailing packet is redundant there as well.
+            // Without this guard, a duplicate packet adds the player (and its summon) to
+            // the area of interest a second time, which the bucket does not deduplicate.
+            this.Logger.LogWarning("Ignoring client-ready packet: player {0} is already on map {1}.", this, this.CurrentMap);
             return;
         }
 
--- a/tests/MUnique.OpenMU.Tests/ClientReadyAfterMapChangeTests.cs
+++ b/tests/MUnique.OpenMU.Tests/ClientReadyAfterMapChangeTests.cs
@@ -1,0 +1,63 @@
+// <copyright file="ClientReadyAfterMapChangeTests.cs" company="MUnique">
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// </copyright>
+
+namespace MUnique.OpenMU.Tests;
+
+using System.Linq;
+using System.Threading.Tasks;
+using MUnique.OpenMU.GameLogic;
+using MUnique.OpenMU.Pathfinding;
+using NUnit.Framework;
+
+/// <summary>
+/// Tests for the handling of the client-ready packet which is sent after a map change.
+/// </summary>
+[TestFixture]
+public class ClientReadyAfterMapChangeTests
+{
+    /// <summary>
+    /// Tests that a repeated client-ready packet does not add the player to the
+    /// area of interest a second time. The bucket does not deduplicate its entries,
+    /// and its removal only strips the first occurrence, so a duplicate entry would
+    /// outlive the player on the map.
+    /// </summary>
+    /// <returns>The task.</returns>
+    [Test]
+    public async Task RepeatedClientReadyDoesNotAddPlayerTwiceAsync()
+    {
+        var player = await PlayerTestHelper.CreatePlayerAsync().ConfigureAwait(false);
+        await player.ClientReadyAfterMapChangeAsync().ConfigureAwait(false);
+
+        var map = player.CurrentMap;
+        Assert.That(map, Is.Not.Null);
+        var occurrencesAfterFirst = map!.GetAttackablesInRange(player.Position, 1).Count(o => o == player);
+        Assert.That(occurrencesAfterFirst, Is.EqualTo(1));
+
+        await player.ClientReadyAfterMapChangeAsync().ConfigureAwait(false);
+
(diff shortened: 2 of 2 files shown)


src/GameLogic/Player.cs
```diff
@@ -949,6 +949,16 @@ public async ValueTask ClientReadyAfterMapChangeAsync()
...
 
+        if (this.CurrentMap is not null)
+        {
+            // Guard against a repeated F3 12 (client ready after map change) packet:
+            // during a regular map change CurrentMap is set to null and only assigned
+            // here, so a non-null value means this handler already ran for the current
+            // map change. Without this, a duplicate packet would add the player (and its
+            // summon) to the map a second time.
+            return;
+        }
```
sven-n, line 963: The guard itself is correct — I traced all four callers and `CurrentMap` is null on entry in every legitimate flow (details in the review summary).

Two small things on this block:

**1. The comment's premise isn't quite universal.** "during a regular map change `CurrentMap` is set to null" holds for `WarpToAsync` (line 882) and the else-branch of `RespawnAtAsync` (line 930), but *not* for the `IRespawnAfterDeathPlugIn` branch at 918-926, which assigns a non-null `CurrentMap` and adds the player itself. That branch is exactly the case where the guard changes behaviour rather than just deduplicating (see the summary), so it would be good for the comment to acknowledge it instead of implying null is the only pre-state.

**2. Consider logging.** A duplicate `F3 12` is client behaviour the server can't otherwise see, and this makes it entirely silent. Elsewhere in this file suspicious client input gets a warning (e.g. the walk-offset resync at line 1145). Something like:

```csharp
if (this.CurrentMap is not null)
{
    this.Logger.LogWarning("Ignoring client-ready packet: player {player} is already on map {map}.", this, this.CurrentMap);
    return;
}
```

would make an actual re-entrancy attempt visible without changing the fix. Happy to defer to @sven-n on whether that's warning- or debug-level noise.

---

sven-n (commented):
Reviewed against `master` at 8919944. The premise checks out and the fix looks right; a couple of notes below, one of which is worth a second look before merging.

## The bug is real, and it's worse than a no-op

Tracing what a second `F3 12` actually does today:

- `GameMap.AddAsync` (GameMap.cs:188) *does* partially dedupe — the `existing != locateable` check means the id isn't regenerated, `_playerCount` isn't double-incremented, and `_objectsInMap` isn't touched.
- But `_areaOfInterestManager.AddObjectAsync(locateable)` runs unconditionally, outside that `if`.
- `BucketAreaOfInterestManager.AddObjectAsync` → `Bucket<T>.AddAsync` → `this._innerList.Add(item)` with no dedupe (Bucket{T}.cs:53), and `Bucket<T>.RemoveAsync` uses `List.Remove`, which only removes the *first* occurrence (Bucket{T}.cs:73).

So a duplicate packet leaves the player twice in its bucket, fires `ItemAdded` twice at every observer, and — because removal only strips one entry — leaves a ghost behind when the player later leaves the map. Same for the summon, which additionally gets `OnSpawn()` called twice. Worth fixing.

Note the `CurrentMap` setter is a no-op on the duplicate (same map ⇒ `_currentMap != value` is false), so `RaisePlayerLeftMap`/`RaisePlayerEnteredMap` don't misfire. The damage is confined to the AOI/bucket layer.

## Call-site audit

All four callers of `ClientReadyAfterMapChangeAsync`, and `CurrentMap` on entry:

| Caller | `CurrentMap` on entry |
|---|---|
| `CharacterClientReadyPacketHandlerPlugIn` (F3 12) | null after `WarpToAsync` (Player.cs:882) or `RespawnAtAsync`'s else-branch (Player.cs:930) ✔ |
| `Player.cs:2340` (enter world) | null — fresh `Player`, and `RemoveFromCurrentMapAsync` sets `_currentMap = null` directly (Player.cs:1851) on the way back to character selection ✔ |
| `OfflinePlayer.StartAsync:201` | null — `SetupCharacterAsync` only does `AddPlayerAsync` + `SetSelectedCharacterAsync` ✔ |
| `OfflineMapChangePlugIn.MapChangeAsync` | null — invoked from the `IMapChangePlugIn` calls immediately after `CurrentMap = null` ✔ |

So no legitimate path regresses.

## The one case worth checking: `RespawnAtAsync` + `IRespawnAfterDeathPlugIn`

The branch at Player.cs:918-926 is the exception to "CurrentMap is null before this handler runs" — it sets `CurrentMap` non-null and calls `AddAsync(this)` itself. If an old client sends `F3 12` after that respawn packet, today that packet is what runs the summon block at the end of this method; with the guard it won't.

Concretely: `TryRemoveFromCurrentMapAsync` removes a live summon from the map (Player.cs:1819-1822), and the respawn-plugin branch re-adds only the player, never the summon.

- On the **death** path this is moot — the summon is removed, disposed and nulled at Player.cs:2167-2172 before `RespawnAtAsync` is reached.
- On the **duel** path it isn't: `RespawnOfDuelPartnerIfInDuelAsync` (Player.cs:2216) calls `duelPartner.RespawnAtAsync(...)` with no summon cleanup.

Either the summon is already orphaned there today (a pre-existing bug in that branch, independent of this PR), or a trailing `F3 12` has been quietly papering over it — in which case this guard surfaces it. Worth a quick check with an old client + an active summon in a duel; if it turns out to be pre-existing, it's a separate fix and shouldn't block this one.

## Minor

- No test. `SpeedHackAntiCheatTests` / `CastleSiegeNpcTests` show the fixtures exist to build a player on a map; a test asserting a second `ClientReadyAfterMapChangeAsync()` leaves the bucket count unchanged would lock this in cheaply and document the invariant.
- Inline note on the comment wording and on logging the duplicate.

Nice, tightly scoped PR otherwise — one file, one behaviour, clear rationale.

---

mrallanodreman:
Thanks — the bucket trace is the part I had missed. The partial dedupe in `GameMap.AddAsync` made this look like a no-op, and it isn't: `AddObjectAsync` running outside that `if`, plus `List.Remove` only stripping the first occurrence, is exactly the ghost-on-leave path. Pushed all three points:

**Comment.** Reworded to stop implying null is the only pre-state, and to name the `IRespawnAfterDeathPlugIn` branch explicitly.

**Logging.** Added a `LogWarning` on the ignored packet, in the same style as the walk-offset resync above it. Happy to drop it to debug if you'd rather not have it at warning level.

**Test.** `ClientReadyAfterMapChangeTests` covers both halves of the invariant: the player is added to the area of interest exactly once across two calls, and no entry survives `RemoveAsync` afterwards. Both fail with the guard removed — the second one is the one that would have caught the ghost.

On the duel case you flagged: it looks pre-existing to me rather than something this guard introduces. `RespawnAtAsync` drops the summon in `TryRemoveFromCurrentMapAsync`, and the `IRespawnAfterDeathPlugIn` branch only re-adds the player — the summon is only ever restored by the `F3 12` path at the end of `ClientReadyAfterMapChangeAsync`. So that branch never restores it on its own, in any scenario. It just doesn't show on the death path, where the summon is already disposed and nulled beforehand; `RespawnOfDuelPartnerIfInDuelAsync` is the only caller that reaches it with a live summon. If that reading is right, a trailing `F3 12` was papering over it and the fix belongs in that branch, separately from this PR. Happy to open an issue for it if you want it tracked.

sven-n:
Merged — thanks for the thorough follow-up. All three points landed well, and the reworded comment naming the `IRespawnAfterDeathPlugIn` branch is more useful than what I suggested.

Verified locally before merging, not just on CI: full suite is 705/705 green, and I checked the new tests aren't vacuous by reverting the guard — both fail without it, `PlayerLeavesNoGhostAfterRepeatedClientReadyAsync` on exactly the ghost path. That's the test I'd have wanted.

On the duel case: your reading is right, and it's pre-existing rather than introduced here. `RespawnAtAsync` drops a live summon in `TryRemoveFromCurrentMapAsync` (Player.cs:1819-1822) and the respawn-plugin branch re-adds only the player, so that branch never restores the summon on its own — the death path just hides it, since the summon is disposed and nulled at Player.cs:2167-2172 beforehand. `RespawnOfDuelPartnerIfInDuelAsync` (Player.cs:2216) is the only caller reaching it with a live summon. So a trailing `F3 12` was compensating for it, and this guard removes that accident. Please do open an issue — the fix belongs in that branch (re-add the summon alongside the player), and it shouldn't ride along here.

---

PR #874, opened 2026-08-10
Bounds-check client-supplied walk steps against the terrain grid

mrallanodreman:
Second of the two pieces from the #867 review, standalone on top of current
master and independent of the re-entrancy guard PR.

## Change

`GetWalkableStepCount` (used by `WalkToAsync`) indexes `terrain.WalkMap` with
coordinates taken directly from the client's walk request. This guards the access
with the grid dimensions and treats an out-of-bounds step as non-walkable,
truncating the path at that step rather than indexing past the terrain grid.

## Honest scope note

On master `Point` is `record struct Point(byte X, byte Y)` and `WalkMap` is
`[256, 256]`, so the index cannot currently exceed the bounds — this is defensive
hardening, not a live bug fix. I'm sending it as its own PR (per your review)
rather than folding it into the large-map proposal. Completely fine to close it if
you'd rather not carry a guard that's unreachable today; it becomes load-bearing
the moment coordinate widths change.

### DIFF
--- a/src/GameLogic/Player.cs
+++ b/src/GameLogic/Player.cs
@@ -1668,8 +1668,14 @@ public class Player : AsyncDisposable, IBucketMapObserver, IAttackable, IAttacke
         for (var index = 0; index < steps.Length; index++)
         {
             var target = steps[index].To;
-            if (!terrain.WalkMap[target.X, target.Y])
+            if (target.X >= terrain.WalkMap.GetLength(0)
+                || target.Y >= terrain.WalkMap.GetLength(1)
+                || !terrain.WalkMap[target.X, target.Y])
             {
+                // Treat an out-of-bounds step as non-walkable instead of indexing past
+                // the terrain grid. With today's byte-sized coordinates on a 256x256 grid
+                // this is unreachable, but it keeps the client-supplied path from indexing
+                // the map out of range once coordinate widths change.
                 return index;
             }
         }


src/GameLogic/Player.cs
```diff
@@ -1668,8 +1668,14 @@ private static int GetWalkableStepCount(GameMapTerrain terrain, ReadOnlySpan<Wal
         for (var index = 0; index < steps.Length; index++)
         {
             var target = steps[index].To;
-            if (!terrain.WalkMap[target.X, target.Y])
+            if (target.X >= terrain.WalkMap.GetLength(0)
+                || target.Y >= terrain.WalkMap.GetLength(1)
+                || !terrain.WalkMap[target.X, target.Y])
```
sven-n, line 1673: If this does get kept, two mechanical notes:

**`GetLength(0)`/`GetLength(1)` per iteration.** This is per walk-step, per walk packet, per player. `GetLength(int)` on a multi-dimensional array isn't an intrinsic the way `Array.Length` is, and the JIT won't reliably hoist it out of the loop here. Hoisting both into locals before the loop, or exposing the size from `GameMapTerrain` (it already has `private const int MapSize = 256` at GameMapTerrain.cs:18), avoids paying it 15+ times per packet for a check that is currently constant-false.

**Comparison is provably false today.** `target.X` is a `byte` widened to `int`, so `>= 256` cannot hold. That's your point in the description, but it also means neither branch of this condition is exercised by any test or runtime path — it's code that can only ever be validated by reading it.

---

sven-n (commented):
Reviewed against `master` at 8919944. Confirming your scope note, and adding three points that I think push this further toward "close it" than the description does — but that's @sven-n's call, not mine.

## Confirming the unreachability

`Point` is `record struct Point(byte X, byte Y)` (Pathfinding/Point.cs:15) and `WalkMap` is `new bool[MapSize, MapSize]` with `MapSize = 256` (GameMapTerrain.cs:18,66). `byte` maxes at 255, so both new conditions are constant-false. Agreed, no live bug.

## It isn't memory-safety hardening

The framing in the description — "indexing past the terrain grid" — is worth correcting, because it changes how the trade-off reads. `bool[,]` indexing in .NET is bounds-checked by the runtime; an out-of-range index throws `IndexOutOfRangeException`, it does not read past the array. So there's no unsafe read being prevented here.

What the patch actually does is convert a hypothetical loud exception into a silent path truncation. That's arguably the *less* desirable failure mode: if coordinate widths ever widen and something starts feeding out-of-range steps, an exception names the bug immediately, whereas a truncated walk just looks like a player rubber-banding for no reason. If the goal is future-proofing, a `Debug.Assert` — which fails loudly in dev and vanishes in release — gets the intent across without changing production behaviour.

## Guarding one call site of ~20 is a false signal

`WalkMap` is indexed in roughly twenty places, and this is far from the only one reachable from client input:

- `WizardTeleportAction.cs:39,65` — client-supplied teleport target
- `DropItemAction.cs:28` — client-supplied drop coordinates
- `TwistingSlashMasterySkillPlugIn.cs:47`, `PollutionSkillPlugIn.cs:47`, `EarthShakeSkillPlugIn.cs:45`, `SummonPartySkillPlugin.cs:116` — skill targets
- `Player.cs:965` — the safezone check in `ClientReadyAfterMapChangeAsync`
- `GuardIntelligence.cs:34`, `NonPlayerCharacter.cs:230`, `ChaosCastleContext.cs:478`, `GameMap.cs:279`, `MovementHandler.cs:105`, `BotNavigator.cs:1808`

If `Point` ever widens past `byte`, every one of these needs attention, and the real fix is upstream anyway: `WalkMap` would have to be sized from the map definition rather than a hardcoded 256, at which point a bounds check derived from the array is the natural consequence of that change, not a precursor to it. Hardening one of twenty sites now mostly risks giving a future reader the impression this area was audited.

## Suggestion

My read: close this one. The scenario it protects against is a coordinate-width change that would require a coordinated pass over `GameMapTerrain` and every consumer, and this guard wouldn't meaningfully shorten that work. If you'd rather leave a marker for that future change, a comment on `GameMapTerrain.WalkMap` noting the `byte`-coordinate assumption would age better than a runtime check that can never fire — and it would be visible to everyone touching the other nineteen sites, not just this one.

Left a couple of mechanical notes inline in case it's kept anyway. Either way, splitting it out rather than folding it into the large-map PR was the right call — it made this easy to evaluate on its own terms.

---

mrallanodreman:
Thanks for the detailed review — the reasoning holds up and I'm closing this.

Your first point settles it for `master`: with `Point` declared as `(byte X, byte Y)` and `WalkMap` allocated as `bool[256, 256]`, both new conditions are provably constant-false. Dead code here.

The second point is the one the description got wrong, and it's worth stating plainly: `bool[,]` indexing is bounds-checked by the runtime, so there was never an unsafe read to prevent. What the patch actually did was convert a loud `IndexOutOfRangeException` into a silently truncated path — and a player rubber-banding for no visible reason is far harder to trace than a stack trace naming the exact call site. That's a worse failure mode, and calling it hardening was misleading.

The third point is fair too. Guarding one of the ~20 `WalkMap` index sites reachable from client input suggests the area was audited when it wasn't.

On where the real fix lives:

> `WalkMap` would have to be sized from the map definition rather than a hardcoded 256, at which point a bounds check derived from the array is the natural consequence of that change, not a precursor to it.

That's exactly right, and it explains where this patch came from: it was extracted from a branch where `GameMapTerrain` already allocates `WalkMap` from the map definition and `Point` carries `ushort`. In that context the same check stops being dead code — but as you say, it only makes sense as a consequence of the sizing change, not ahead of it. Sending it on its own made it look like standalone hardening, which it wasn't. That's the design sketched in #875, whenever there's time for it.

Closing. Thanks for taking the time to trace it properly.

Issue #875, opened 2026-08-10
Design proposal: opt-in support for maps larger than 256×256


mrallanodreman:
# Design proposal: opt-in support for maps larger than 256×256

Hi @sven-n, and thank you — both for OpenMU itself and for the unusually
detailed review on #867. You were right on every count, and I want to start by
owning the mistake: I opened that PR far too early, as one big lump, before I
understood how a change like this should enter the project. Your feedback gave me
the map I was missing. So this is me stepping back and doing it the right way —
opening the design conversation first, before any more code.

I'm closing/leaving #867 aside as you suggested and splitting the useful parts
out (see "Immediate small PRs" below).

And honestly — I'm genuinely excited to share where this got to. We got large maps
**actually rendering in the client**, terrain **height sculpting** working, and it
looks great in-game. That's the part I want to put in front of the community: not a
promise, but something that already runs.

## The goal

Let servers **opt in** to maps larger than the current fixed 256×256 grid, so
map makers can finally build spaces the classic client never could — bigger
continents, raid zones, event arenas, sculpted terrain with real elevation —
without changing anything for the 99% of servers that stay on 256×256. This opens
a whole new canvas for map makers, and the exciting part is we've already seen it
work end to end: server → protocol → client → rendered on screen.

Why this is worth a design discussion rather than just a feature flag: coordinates
are 8-bit across the whole stack, so "bigger maps" touches the data model, the
terrain loader, the wire protocol, the pathfinding, and the tools. It has to be
done as a sequence of reviewable, backward-compatible steps, not one PR.

## The client story (the part #867 was missing)

You correctly pointed out that no stock client can use any of this, which is what
makes an out-of-tree protocol dialect unmaintainable for OpenMU. That gap is real,
and here's how I'd like to close it so the server side has a legitimate consumer:

- **An open client — and it's already public.** We maintain
  [**LinuxMu**](https://github.com/mrallanodreman/LinuxMu), a fork of the Season 6
  client (based on the [MuMain](https://github.com/sven-n/MuMain) sources) that runs
  against OpenMU on Linux and Windows, and large maps render in it today. It's out
  in the open, not a private binary only we can build — it lives on
  [`main`](https://github.com/mrallanodreman/LinuxMu), and you can read the entire
  client-side change in this
  [full diff](https://github.com/mrallanodreman/LinuxMu/commit/25e7e70343a99b2ec9b297b2b7a46478574c38d6).
[...]

PR #876, opened 2026-08-10
feat: global map coordinates for large maps (512x512+)

mrallanodreman:
Implements Global packet variants (C1:D5 walk, C3:CC teleport, C3:CD gate, C3:CE area-skill, C3:CF drop-item, C2:x:D0 scope/S2C variants) so that maps > 256x256 work end-to-end: server pathfinding, persistence, spawn placement, warp/gate, and client rendering.

Key changes:
- : size derived from sqrt(data.Length), arrays dynamically sized
- /: runtime grid dimensions, no byte.MaxValue limits
- : takes mapSize parameter
- Player coords serialized as ushort throughout (EF migration included)
- NPC spawn validation respects terrain bounds
- Walk, teleport, map-change, scope, respawn all use Global packets when coords > 255

Runtime verified: movement 254→256→511→0, teleport, warp index 0/2, gate, spawn center.
Hito 512 achieved — see mu-spec/MILESTONE-512.md for full validation.


mrallanodreman:
Closing this — it shouldn't have been opened, and that's on me.

This is the same branch as #867, which you closed on design grounds, not because of the
defect list. Re-sending it (and larger) ignores what you actually asked for. Sorry for
the noise.

The right place for this discussion is #875, where I wrote up the design proposal and the
phased plan starting with the data model and terrain-size handling, with no behaviour
change at 256x256. I'll wait for your thoughts there instead of pushing code ahead of it.

#873 and #874 stay open as the small, independent pieces you said you'd look at separately.

PR #878, opened 2026-08-12
Give bots their own log category and quiet their per-tick logging

nolt:
## Summary

Server-side bots now log under their own category
(`MUnique.OpenMU.GameLogic.Bots.BotPlayer`) instead of sharing
`MUnique.OpenMU.GameLogic.Player` with real players, and their recurring per-bot
log entries drop from Information to Debug. A single Serilog
`MinimumLevel.Override` can now silence bots — or turn them up for inspection —
without touching real players, whose logging stays byte-for-byte unchanged.

## Background

Every `Player` created its logger with `CreateLogger<Player>()`, so all
subclasses — the connected `RemotePlayer`, the offline MU Helper `OfflinePlayer`
and `BotPlayer` — logged under the single `...GameLogic.Player` category. On a
server running bots this made bot output indistinguishable from a real player's.

## The problem

Measured on a Season Six server with 500 bots (100 accounts × 5 characters, 3
game servers): 592 log lines per minute just after startup, 293 in the steady
state, every single one under `...GameLogic.Player`. `BotNavigator` alone was
74.6% of the volume, its "heading to hunting ground" line 58.7%. No log
configuration could quiet the bots without also quieting real players.

## The fix

Two independent parts.

**1. Own category.** A new `RuntimeCategoryLogger<T>` — an `ILogger<T>` adapter
that resolves its category from the owner's *runtime* type. `BotPlayer` installs
it for itself in its constructor, so bots log under `...GameLogic.Bots.BotPlayer`.
`Player`'s constructor is untouched (`CreateLogger<Player>()`), so connected and
offline real players keep the exact `...GameLogic.Player` category as before. The
base class has no knowledge of the bot subclass; `Player.Logger` merely gains a
`protected set` so a subclass can supply its own logger. The adapter is needed
because the `ILoggerOwner<Player>` contract requires an `ILogger<Player>` while
`CreateLogger(Type)` returns a non-generic `ILogger`.

**2. Quiet by default.** The 33 bot log entries that fire on a recurring per-bot
basis (hunting-ground walks, shopping trips, party churn, equipment and
progression milestones) move from Information to Debug. The eight operator-facing
events keep Information (start, stop, purge, reset, account generation, population
summary); all warnings and errors are unchanged.

The same 500 bots now produce about one log line per five minutes at the default
level.

## Configuration

- Default (`appsettings.json`: `Default=Information`, `MUnique=Information`): bots
  are effectively silent; real players are unaffected.
- Silence bots entirely:
  `Serilog__MinimumLevel__Override__MUnique.OpenMU.GameLogic.Bots=Warning`.
- Inspect bot behaviour: `...GameLogic.Bots=Debug` (verbose — dominated by the
  per-step `Player.WalkToAsync` trace; `| grep -v WalkToAsync` to skim).

`BotFeaturePlugIn`'s own operator lines use the periodic-plugin house-style
category (`BotFeaturePlugIn`, from `CreateLogger(GetType().Name)`, like the other
periodic plugins), so they are addressed by the `BotFeaturePlugIn` key, not the
`...Bots` prefix — deliberately left consistent with the sibling plugins.

## Verified live

Rebuilt and run with 200 generated bots (~65 concurrently animated). At the
default Information level: 0 lines under `...GameLogic.Player` and no per-tick bot
spam. Flipping only `...GameLogic.Bots=Debug` brought back 2600+ Debug lines, all
under `...Bots.BotPlayer`, still 0 under `...GameLogic.Player`.

## Changes

- `src/GameLogic/RuntimeCategoryLogger.cs` (new) — the `ILogger<T>` adapter that
  categorises by runtime type.
- `src/GameLogic/Bots/BotPlayer.cs` — the constructor installs the
  runtime-category logger.
- `src/GameLogic/Player.cs` — `Logger` gains a `protected set`; the constructor is
  otherwise unchanged.
- `src/GameLogic/Bots/*` — 33 recurring `LogInformation` → `LogDebug` across
  `BotNavigator`, `BotPartyHandler`, `BotShoppingHandler`, `BotMiniGameHandler`,
  `BotMasterHandler`, `BotSkillProgressionPlugIn`, `BotWingHandler`,
  `BotJewelHandler`, `BotEquipmentHandler`, `BotResetHandler`, `BotManager`,
  `BotFeaturePlugIn`. Two comments in `BotShoppingHandler` were reworded so they no
  longer imply the line is visible at the default level.

## No client impact

Server-side only; no packet or client change.

### DIFF
--- a/src/GameLogic/Bots/BotEquipmentHandler.cs
+++ b/src/GameLogic/Bots/BotEquipmentHandler.cs
@@ -114,7 +114,7 @@ internal static class BotEquipmentHandler
             return false;
         }
 
-        player.Logger.LogInformation(
+        player.Logger.LogDebug(
             "Bot '{Name}' equipped '{New}'{Replaced}.",
             player.Name,
             item,
--- a/src/GameLogic/Bots/BotFeaturePlugIn.cs
+++ b/src/GameLogic/Bots/BotFeaturePlugIn.cs
@@ -603,7 +603,7 @@ public class BotFeaturePlugIn : IFeaturePlugIn, IPeriodicTaskPlugIn, ISupportCus
             if (offline.SelectRandom() is { Login: not null } candidate
                 && await state.Manager.SpawnBotAsync(gameContext, candidate.Login, candidate.Slot).ConfigureAwait(false))
             {
-                logger.LogInformation("Bot presence rotation: +1 (online {Online}/{Target} of {Total}).", online + 1, targetOnline, totalPopulation);
+                logger.LogDebug("Bot presence rotation: +1 (online {Online}/{Target} of {Total}).", online + 1, targetOnline, totalPopulation);
             }
         }
         else if (online > targetOnline)
@@ -611,7 +611,7 @@ public class BotFeaturePlugIn : IFeaturePlugIn, IPeriodicTaskPlugIn, ISupportCus
             var stopped = await state.Manager.StopRandomBotAsync().ConfigureAwait(false);
             if (stopped is not null)
             {
-                logger.LogInformation("Bot presence rotation: -1 '{Name}' (online {Online}/{Target} of {Total}).", stopped, online - 1, targetOnline, totalPopulation);
+                logger.LogDebug("Bot presence rotation: -1 '{Name}' (online {Online}/{Target} of {Total}).", stopped, online - 1, targetOnline, totalPopulation);
             }
         }
     }
--- a/src/GameLogic/Bots/BotJewelHandler.cs
+++ b/src/GameLogic/Bots/BotJewelHandler.cs
@@ -358,7 +358,7 @@ internal static class BotJewelHandler
         var consumed = inventory.GetItem(jewelSlot) != jewel;
         if (consumed)
         {
-            player.Logger.LogInformation(
+            player.Logger.LogDebug(
                 "Bot '{Name}' used '{Jewel}' on '{Item}': level {Before} -> {After}.",
                 player.Name,
                 jewel.Definition?.Name,
--- a/src/GameLogic/Bots/BotManager.cs
+++ b/src/GameLogic/Bots/BotManager.cs
@@ -87,7 +87,7 @@ public sealed class BotManager
 
             BotSkillProgressionPlugIn.CatchUpPendingProgress(bot);
 
-            bot.Logger.LogInformation("Bot started for account '{LoginName}', character '{Character}'.", loginName, character.Name);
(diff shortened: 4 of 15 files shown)


sven-n:
## Review

Reviewed the full diff (15 files, +92/−39) against `master` @ `8919944`. **Looks good — approving and merging.**

### Correctness

**`RuntimeCategoryLogger<T>`** — a straight pass-through adapter over `ILoggerFactory.CreateLogger(Type)`. All three `ILogger` members (`BeginScope`, `IsEnabled`, `Log`) delegate to `_inner`, nothing is swallowed or reinterpreted, and the type is `internal` so it stays an implementation detail of `GameLogic`. The generic parameter exists only to satisfy the `ILoggerOwner<Player>` contract (`new ILogger<T> Logger { get; }` in `ILoggerOwner.cs:28`), which is exactly what the XML remarks explain.

**`Player.Logger` gaining `protected set`** (`Player.cs:156`) — the narrowest change that makes this work. The base constructor still does `CreateLogger<Player>()` (`Player.cs:104`), so `RemotePlayer` and `OfflinePlayer` keep the byte-identical `MUnique.OpenMU.GameLogic.Player` category, and the base class needs no knowledge of the bot subclass. Assignment order is safe: `BotPlayer`'s constructor body runs after the base constructor, and the two other consumers of `Logger` captured outside the constructor — `SkillHitValidator` (`Player.cs:186`, lazy `??=`) and `_accountLoggingScope` (`Player.cs:268`, set on account assignment) — are both resolved well after construction, so they pick up the bot logger, not the discarded base one.

**Level changes** — spot-checked all 33 sites. Every downgraded entry is on a recurring per-bot path (navigation, shopping, party churn, equipment/wing/jewel/master/reset milestones, per-bot spawn). Warnings and errors are untouched, and the operator-facing aggregates in `BotFeaturePlugIn` (`Bot feature started {Started} of {Total}`, stop, purge, reset, account generation) and `BotGenerator` remain at Information. The two reworded comments in `BotShoppingHandler` correctly no longer claim default-level visibility while preserving the reason those lines exist.

### Notes (non-blocking)

- `BotPlayer` construction now creates a `Player`-category logger in the base constructor and immediately replaces it. One throwaway resolution per bot spawn — irrelevant at these rates, and avoiding it would require touching `Player`'s constructor, which this PR deliberately leaves alone.
- `Bot presence rotation: ±1` moving to Debug means population drift is no longer visible at the default level. The periodic `Bot feature started {Started} of {Total}` line still reports the population, so this seems fine, but it is the one downgrade that is arguably operator-facing rather than per-bot.

### CI

Both checks green: `MUnique.OpenMU` (Azure build #4453) and Codacy Static Code Analysis. `mergeable_state: clean`.

---

PR #879, opened 2026-08-13
Add configurable server time zone for periodic event schedules

nolt:
## Problem

Periodic events (Golden/Red Dragon invasions, White Wizard Invasion, Blood Castle,
Devil Square, Chaos Castle, Happy Hour) decide when to start in
`PeriodicTaskConfiguration.IsItTimeToStart()`, which compares the configured
`Timetable` (a list of `TimeOnly`) against `DateTime.UtcNow`. The times of day are
therefore effectively interpreted as **UTC**, but nothing in the model or the Admin
Panel says so. Consequences:

- Admins enter what they think are local times; the server treats them as UTC.
- Anything presenting these times (e.g. a website) has to re-apply an offset, which
  is a second, separate source of truth.
- The local wall-clock time of an event drifts by an hour across daylight saving time.

## Solution

Introduce a single, optional `SystemConfiguration.TimeZoneId`. The scheduler
interprets the timetable in that zone; a website or any other consumer can read the
same value. Empty/unresolvable ⇒ UTC, i.e. exactly today's behavior.

## Changes

- **DataModel** — `SystemConfiguration.TimeZoneId` (nullable string) with a `[Display]`
  name/description, plus resources and an EF migration adding the nullable `text`
  column in the `config` schema.
- **GameLogic** — `IGameContext.ServerTimeZone` (defaults to `TimeZoneInfo.Utc`).
  `PeriodicTaskConfiguration.IsItTimeToStart(TimeZoneInfo)` converts `DateTime.UtcNow`
  into that zone before matching the timetable. The single caller in
  `PeriodicTaskBasePlugIn` passes `gameContext.ServerTimeZone`.
- **Startup** — `TimeZoneId` is resolved once (`Program.ResolveServerTimeZone`) and
  flowed into each game server context in `GameServerContainer`. Resolution accepts
  IANA and Windows ids (ICU maps them on both platforms since .NET 6) and falls back
  to UTC with a warning when the id is empty or unresolvable.
- **Admin Panel** — the generic editor for the **System** section shows a new
  *Server time zone* field. A dedicated component builder (same pattern as the
  existing `PasswordHashFieldBuilder`) renders a text input backed by a `<datalist>`
  of the available system time zones, normalized to IANA (`HasIanaId` /
  `TryConvertWindowsIdToIanaId`) so the suggestions look the same regardless of the
  host OS. The field validates the entered id (`FindSystemTimeZoneById`) and rejects
  garbage with a message, while still allowing any free-text value and an empty value
  (= UTC).

## Backward compatibility

`TimeZoneId` null/empty ⇒ everything is interpreted in UTC ⇒ **identical to current
behavior**. Existing databases are untouched until an admin sets a value.

## One-time step for admins who set a zone

After setting e.g. `Europe/Warsaw`, the existing timetable entries (which were UTC)
should be shifted once to the natural local time, because they are now interpreted in
the configured zone. New entries are entered directly as local time.

## Cross-platform

- `FindSystemTimeZoneById` accepts IANA ids on Windows too (no `tzdata` package
  needed there — Windows ships its own time zone data, and ICU maps IANA⇄Windows).
- Requires ICU (globalization not invariant), which OpenMU already enables
  (`DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false`).
- Without any time zone data at all, the field degrades to a plain text input and
  resolution falls back to UTC — nothing throws.

## Testing

- Fresh seed: migration applies, column present.
- Invalid id (`Bogus/DoesNotExist`) ⇒ `[Warning] Could not resolve configured server
  time zone ... Falling back to UTC`, events run in UTC.
- End-to-end: `TimeZoneId=Europe/Warsaw` + a non-whole-hour local slot ⇒ the event
  fired at the configured **local** time (e.g. Chaos Castle at 03:14 local = 01:14
  UTC), which UTC-based scheduling could not produce.
- Admin Panel field renders, suggests zones, validates garbage, accepts empty (= UTC).
- Build clean (0 warnings introduced; the earlier CS1574 from a stray doc cref was
  fixed during review).

## Companion

A separate, optional website PR (`openmu-simple-web`, branch
`events-timezone-from-config`) makes the events page read the same `TimeZoneId` from
the database instead of the container time zone. Deploy the server change first.

### DIFF
--- a/src/DataModel/Configuration/SystemConfiguration.cs
+++ b/src/DataModel/Configuration/SystemConfiguration.cs
@@ -76,6 +76,18 @@ public partial class SystemConfiguration
         ResourceType = typeof(Resources))]
     public bool ReadConsoleInput { get; set; }
 
+    /// <summary>
+    /// Gets or sets the time zone id used to interpret the time-of-day entries of the periodic
+    /// event schedules (e.g. invasions, Blood Castle, Devil Square). IANA ids are recommended
+    /// (for example "Europe/Warsaw"); when empty or unresolvable, UTC is used.
+    /// </summary>
+    [Display(
+        Order = 6,
+        Name = nameof(Resources.SystemConfiguration_TimeZoneId_Name),
+        Description = nameof(Resources.SystemConfiguration_TimeZoneId_Description),
+        ResourceType = typeof(Resources))]
+    public string? TimeZoneId { get; set; }
+
     /// <inheritdoc />
     public override string ToString()
     {
--- a/src/DataModel/Properties/Resources.Designer.cs
+++ b/src/DataModel/Properties/Resources.Designer.cs
@@ -217,5 +217,23 @@ namespace MUnique.OpenMU.DataModel.Properties {
                 return ResourceManager.GetString("SystemConfiguration_ReadConsoleInput_Name", resourceCulture);
             }
         }
+
+        /// <summary>
+        ///   Looks up a localized string similar to Server time zone.
+        /// </summary>
+        public static string SystemConfiguration_TimeZoneId_Name {
+            get {
+                return ResourceManager.GetString("SystemConfiguration_TimeZoneId_Name", resourceCulture);
+            }
+        }
+
+        /// <summary>
+        ///   Looks up a localized string similar to The time zone id used to interpret the times of periodic event schedules (e.g. invasions, Blood Castle, Devil Square). IANA ids are recommended, for example &apos;Europe/Warsaw&apos;. When empty, UTC is used..
+        /// </summary>
+        public static string SystemConfiguration_TimeZoneId_Description {
+            get {
+                return ResourceManager.GetString("SystemConfiguration_TimeZoneId_Description", resourceCulture);
+            }
+        }
     }
 }
--- a/src/DataModel/Properties/Resources.resx
+++ b/src/DataModel/Properties/Resources.resx
@@ -127,6 +127,12 @@ or behind a firewall) and you want to use it within your computer or private net
 	<data name="SystemConfiguration_ReadConsoleInput_Name" xml:space="preserve">
 		<value>Read console input</value>
 	</data>
+	<data name="SystemConfiguration_TimeZoneId_Name" xml:space="preserve">
+		<value>Server time zone</value>
+	</data>
+	<data name="SystemConfiguration_TimeZoneId_Description" xml:space="preserve">
+		<value>The time zone id used to interpret the times of periodic event schedules (e.g. invasions, Blood Castle, Devil Square). IANA ids are recommended, for example 'Europe/Warsaw'. When empty, UTC is used.</value>
+	</data>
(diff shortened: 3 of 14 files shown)


sven-n:
## Review

I went through the full diff (the 5.8k lines are the auto-generated migration designer snapshot; the hand-written change is ~150 lines across 13 files). Overall this is a clean, well-scoped change and I'm happy with it.

### What's good

- **Backward compatible by construction.** `TimeZoneId` null/empty ⇒ `TimeZoneInfo.Utc` ⇒ byte-identical behavior to today. The migration only adds a nullable `text` column, and the model snapshot is updated consistently with it.
- **Single source of truth.** Putting it on `SystemConfiguration` and exposing it through `IGameContext.ServerTimeZone` means the scheduler and any external consumer (website) read the same value, which is exactly the problem being solved.
- **Resolution is defensive.** `ResolveServerTimeZone` catches only `TimeZoneNotFoundException`/`InvalidTimeZoneException`, logs a warning and falls back to UTC — a bad id degrades instead of preventing startup. `GetSuggestibleTimeZoneIds()` likewise degrades to a plain text field if no tz data is present.
- **Fits existing patterns.** `TimeZoneIdFieldBuilder` mirrors `PasswordHashFieldBuilder` 1:1 and is registered before the generic `TextFieldBuilder`, which is the required order for it to win. The static `Program.ServerTimeZone` follows the existing static `_systemConfiguration` convention in that file.
- Only one call site of `IsItTimeToStart()` existed (`PeriodicTaskBasePlugIn`), and it's updated; `WanderingMerchantsPlugIn` overrides the `IGameContext` overload and is unaffected.

### Notes / follow-ups (none blocking)

1. **DST transition edges.** `ConvertTimeFromUtc` + a 5-second match window means a timetable slot that falls inside the spring-forward gap never fires that day, and a slot in the fall-back overlap can fire twice. That's inherent to wall-clock scheduling and is arguably the desired semantic here, but it's worth a line in the docs for admins picking times near 02:00–03:00.
2. **Signature change is source-breaking.** `PeriodicTaskConfiguration.IsItTimeToStart()` is `public virtual`; anyone with an out-of-tree plugin calling or overriding it will need to update. An overload forwarding to `TimeZoneInfo.Utc` would avoid that, but since it's a single in-repo caller I don't think it's worth the extra surface.
3. **Requires a restart to take effect.** The zone is resolved once during `ReadSystemConfigurationAsync` and pushed into contexts at game-server init, so changing it in the Admin Panel won't apply live. Same as the other `SystemConfiguration` values, so consistent — just something to mention to admins.
4. The one-time shift of existing timetable entries after setting a zone is real and easy to overlook; the PR description covers it well, and it'd be good to carry that note into the release notes.

### Verdict

Behavior-preserving by default, cross-platform handling is correct (ICU maps IANA⇄Windows ids since .NET 6, and OpenMU already runs non-invariant), CI is green. Approving and merging.

---

Issue #853, opened 2026-07-25
Fresh start bots configuration


wd40-h:
Would it be possible to add a configuration to create bots as fresh start players, beginning at level 1 with starter equipment? We could use it to verify how much time players will take to reach max level.

PR #880, opened 2026-08-15
feature: implement bot fresh start flag and receive elf solder buff

eduardosmaniotto:
## Summary

Implements #853. Bot improvements: fresh-character startup, friendlier combat targets for low-level bots, Elf Soldier buff trips, and real-player skill parity — bots learn exactly what a player could.

### Start as fresh characters (`StartAsFreshCharacters`)

- New bot option: freshly generated bots start like new players (level 1, level-0 starter gear, no reset history) instead of as leveled veterans.
- The veteran path becomes a `BotStartupProfile`, chosen at generation (`BotStartupProfile.For(...)`). Documented in `docs/Bots.md`; covered by `BotStartupProfileTests`.

### Combat level headroom

- `CombatHandler.IsSafeTarget` now engages monsters up to `LevelHeadroom` (5) levels above a bot's effective level, so fresh level-1 bots can actually find targets; true safety still comes from the existing damage/defense checks. Two new `CombatHandlerTests` cases.

### Elf Soldier buff trips

- Bots claim the free buff a dialog NPC offers (Elf Soldier): when configured and qualifying, they walk over (warping home first if needed), talk, request the buff and close the dialog like a player.
- `BotBuffHandler` caches the offers once per game configuration; trips are errands like shopping — they beat hunting/party-follow, and `BotPartyHandler` declines invites meanwhile. Orchestrated in `BotNavigator.TryBuffTripAsync`; a no-op on servers without a buff NPC.

### Skill parity with real players

- Bots no longer learn skills that only come from an item — orb, scroll, weapon, pet — when the skill itself carries no requirements: the gate lives on the item, so a fresh DK no longer gets Twisting Slash, an MG no Fire Slash/Power Slash, an Elf no Ice Arrow, a DL no Earthshake/Force Wave, an RF no Killing Blow/Beast Uppercut/Phoenix Shot.
- Item-granted skills with requirements of their own stay learnable once the bot qualifies (Rageful Blow at 170). The item-granted set is derived from the server config (`GetItemGrantedSkillNumbers`), so future items are covered automatically; one explicit entry stays for a skill whose granting weapon isn't in the config.
- Applied at generation and level-up; bot builds no longer differ between reset and classic servers.


src/GameLogic/Bots/BotNavigator.cs
```diff
@@ -733,6 +779,109 @@ private async ValueTask TradeAndUpgradeAsync(GameMap map, Point merchantPosition
...
+            // No buff NPC lives on this map at all. Like a player pulling a town scroll, the bot warps
+            // to its class hometown, where the buff is offered; the better-map logic takes it back
+            // hunting afterward.
+            if (this.TryGetHomeEscapeGate(out var homeGate, out var homeMap, out _)
+                && homeGate.Map != map.Definition)
+            {
+                this._travelPath = null;
+                this._hasDestination = false;
+                this._lastWarpUtc = DateTime.UtcNow;
+                this._nextBuffCheckUtc = DateTime.UtcNow; // start the trip on the new map right away
```
sven-n, line 849: This branch sets `_nextBuffCheckUtc = DateTime.UtcNow` and warps home. If no live buff NPC is found on the home map either, we fall through to `return false` with `_nextBuffCheckUtc` still at only `now + BuffCheckInterval` (30 s, set earlier in the method), instead of the long `NextBuffCheckUtc()` backoff that every other failed trip applies.

On a server where the buff NPC definition exists but has no reachable live instance, a level-eligible bot then gets dragged home from its hunting map every ~30 seconds indefinitely — and `_lastWarpUtc` is reset on each pass, so `WarpCooldown` doesn't dampen it either. Applying the `BuffCooldown` backoff on the failure path would fix both.

---

src/GameLogic/Bots/BotBuffHandler.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        // Eager snapshot, like the other readers of ActiveEffects (see MagicEffectsList): the list is
+        // mutated by the effect expiry timers, and a lazy enumeration from this (unsynchronized)
+        // helper tick raced them regularly at scale. A torn read simply counts as "active"; the next
+        // tick then skips the trip for a moment and re-evaluates.
+        try
+        {
+            return effects.Values.ToArray().Any(e => e?.Definition == effectDef);
+        }
+        catch (Exception)
```
sven-n: `catch (Exception) { return true; }` swallows every failure of the effect scan, not just the documented concurrent-mutation race. A `NullReferenceException` or a configuration error in `e.Definition` is permanently reported as "already buffed", so those bots silently never claim the buff and the real cause never surfaces in logs.

Catching the narrow collection-mutation case (`InvalidOperationException`), or at minimum logging at debug level here, would keep the intended resilience without hiding genuine errors.

---

sven-n (commented):
Reviewed the bot buff trips, startup profiles, and stat-weight changes. Two correctness concerns noted inline; the rest looks good.

---

sven-n (approved):
Both review findings are addressed in e96ec5e: the failed buff-trip path now applies the full `NextBuffCheckUtc()` backoff, and the effect-scan catch is narrowed to `InvalidOperationException` with a debug log. CI is green. LGTM.

---

PR #881, opened 2026-08-15
feature: add object level and master level to live map

eduardosmaniotto:
# Live map: select players from the list & show level info

Improves the live-map overview in the AdminPanel.

## What changed

- **Select players from the list** — clicking a row in the player list now selects
  that player's object on the map, exactly like clicking the object itself
  (highlight + info card).
  - New `OnSelectPlayer` callback from `MapPlayerList` → `Map` → new
    `SelectMapObject(...)` JS interop → `MapApp.selectObject(objectId)`.
  - Row action buttons (disconnect/ban/follow) use `stopPropagation`, so they no
    longer trigger row selection; rows get a pointer cursor.
  - Selection uses the map-object id, which matches the Three.js world key.
- **Show level / master level** in the object info card, carried through
  `MapObject`, `CreateMapObject`, and the `ObjectData`/`PlayerData` types.
- Rebuilt `MUnique.OpenMU.Web.Map.js` (+ sourcemap) from TypeScript sources.


src/Web/Map/Components/Map.razor
```diff
@@ -75,6 +79,18 @@
...
+    private async Task SelectPlayerAsync(int objectId)
+    {
+        try
+        {
+            await this.JsRuntime.InvokeVoidAsync("SelectMapObject", this.Server.Id, this.MapId, objectId).ConfigureAwait(false);
+        }
+        catch
+        {
+            // The player may not be loaded into the Three.js scene yet.
+        }
```
sven-n: The case in the comment never actually throws: `SelectMapObject` returns `false` when the app or the object is missing, it doesn't raise. What this bare `catch` really swallows is `JSDisconnectedException` and any genuine interop error, with no log line.

Suggest narrowing to `catch (JSDisconnectedException)` / `catch (JSException)`, and logging the unexpected case at debug level — the view plug-ins in this project already take an `ILoggerFactory`, so an injected logger here would be consistent.

---

src/Web/Map/Map/LocateableExtensions.cs
```diff
@@ -28,6 +29,18 @@ public static MapObject CreateMapObject(this ILocateable locateable)
...
             NpcObjectKind = (locateable as NonPlayerCharacter)?.Definition.ObjectKind,
+            Level = GetLevel(locateable),
+            MasterLevel = GetMasterLevel(locateable),
         };
     }
+
+    private static int GetLevel(ILocateable locateable) => locateable switch
+    {
+        Player player => player.Level,
+        IAttackable attackable => (int)attackable.Attributes[Stats.Level],
```
sven-n: The `Player` arm is subsumed by the `IAttackable` arm — `Player.Level` is defined as `(int)(this.Attributes?[Stats.Level] ?? 0)` (`Player.cs:223`), so both compute the same value. The difference is that the `IAttackable` arm dereferences `Attributes` without the null guard.

Neither is a live bug today (`MonsterAttributeHolder.GetValueOfAttribute` falls through to 0 for undefined attributes, and a `Player` always hits the first arm), but if you collapse this to one arm, keep the null-safe form.

---

src/Web/Map/Map/LocateableExtensions.cs
```diff
@@ -28,6 +29,18 @@ public static MapObject CreateMapObject(this ILocateable locateable)
             X = locateable.Position.X,
             Y = locateable.Position.Y,
             NpcObjectKind = (locateable as NonPlayerCharacter)?.Definition.ObjectKind,
+            Level = GetLevel(locateable),
```
sven-n, line 32: Worth noting in the description: the level is captured only when the object is created, and `CreateMapObject` is re-sent only on new-in-scope and on `MoveType.Teleport` (`ObjectMovedPlugIn.cs:53`) — ordinary walking goes through `moveTo`, which doesn't replace `obj.data`. So a player who levels up while already on the map keeps the level they had when they entered scope until they teleport or re-enter.

The card's X/Y already behaves this way, so this is consistent rather than a regression — just not obvious to someone reading the card.

---

src/Web/Map/Scripts/livemap/MapApp.ts
```diff
@@ -69,6 +71,22 @@ export class MapApp {
...
+    /**
+     * Selects the object with the specified id, as if it was clicked on the map.
+     * @param objectId - The id of the object to select.
+     * @returns True if the object was found and selected, false otherwise.
+     */
+    public selectObject(objectId: number): boolean {
+        const data = this.world.getObjectById(objectId)?.data as ObjectData | undefined;
+        if (data === undefined) {
+            return false;
+        }
```
sven-n, line 91: When the lookup misses (the object isn't in the scene yet), this returns `false` and nothing else happens — so any previously selected object keeps its highlight and its info card. From the user's side the row click looks like it worked, but the card is showing a different player.

Clearing the selection on a miss would be more honest, e.g. `this.onObjectPicked(null)` before returning `false` (the launcher's picker callback hides `#selected_info` on `null`, so the card would need the same treatment).

---

src/Web/Map/Components/Map.razor
```diff
@@ -59,6 +60,9 @@
     [Parameter]
     public EventCallback<string?> OnFollowPlayer { get; set; }
 
+    [Inject]
+    private IJSRuntime JsRuntime { get; set; } = null!;
```
sven-n, line 68: Nit: this is the only `[Inject]` member in this `@code` block without an XML doc comment — `MapFactory` and the parameters above all have one.

---

sven-n (commented):
Reviewed the change end to end — it's correct and cleanly scoped, no blocking issues. The comments below are minor.

What I verified:

- `MapPlayerList` rows key off `player.Id` from `ConcurrentDictionary<int, Player>`, which is the same `locateable.Id` written into `MapObject.Id` and used as the key in the Three.js `world.objects` map — so `selectObject(objectId)` resolves the right mesh.
- Blazor JS interop serializes with camelCase, so `Level`/`MasterLevel` arrive as `data.level`/`data.masterLevel`, matching the launcher.
- Neither `GetLevel` nor `GetMasterLevel` can throw or NRE for a player without a selected character (`Player.Level` is null-guarded; `MonsterAttributeHolder.GetValueOfAttribute` falls through to 0).
- `WorldObjectPicker` keeps no "last picked" state, so driving `onObjectPicked` from outside the picker can't desync it — a later canvas click still behaves correctly.
- The committed `MUnique.OpenMU.Web.Map.js` diff matches the TypeScript change exactly, with no unrelated bundle drift, and `MapPlayerList.razor.css` fits the CSS-isolation setup already used by `Map.razor.css` / `LiveMap.razor.css` in this library.

One thing not covered by an inline comment: the info card was moved above the player list, which is a visible layout change the description doesn't mention. It looks intentional, just worth calling out.

---

PR #886, opened 2026-08-18
remove bot boredoom

eduardosmaniotto:
## Remove bot party boredom and raise bot party share

Bots no longer get "bored" and leave a party with a human player after a few minutes.
Previously a bot would walk away from a player who grouped it once a random 10–20 minute
boredom timer ran out.

Additionally, more bots group together: the share of bots placed into parties during the
hourly re-formation rises from 60% to 80%, so bot parties are more common on the server.

### DIFF
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -253,8 +253,9 @@ whole session. A living player takes precedence over the bot's own company
 — a bot hunting with other bots leaves them for the inviter, and breaks that bot
 party up if it was leading it, so a player never has to guess which bot happens
 to be free. In a party the bot follows its leader and defers a due reset, leaving
-only on the engine's own terms: when the party disbands, when it is kicked, or
-when the leader enters a map it may not access.
+only on the engine's own terms: when the party disbands, when it is kicked,
+when the leader enters a map it may not access, or before the bot's own logout (the
+presence rotation or a fault restart stops it).
 
 A bot fights back when a player attacks it, but only as far as the game's own
 PvP rules allow: inside the active self-defense window, or against a player
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -254,8 +254,10 @@ whole session. A living player takes precedence over the bot's own company
 party up if it was leading it, so a player never has to guess which bot happens
 to be free. In a party the bot follows its leader and defers a due reset, leaving
 only on the engine's own terms: when the party disbands, when it is kicked,
-when the leader enters a map it may not access, or before the bot's own logout (the
-presence rotation or a fault restart stops it).
+when the leader enters a map it may not access, before the bot's own logout (the
+presence rotation or a fault restart stops it), or - once no live human is left
+in the party - ten minutes after its companion disconnects, so the bot is not
+stranded in a dead party forever.
 
 A bot fights back when a player attacks it, but only as far as the game's own
 PvP rules allow: inside the active self-defense window, or against a player
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -254,10 +254,9 @@ whole session. A living player takes precedence over the bot's own company
 party up if it was leading it, so a player never has to guess which bot happens
 to be free. In a party the bot follows its leader and defers a due reset, leaving
 only on the engine's own terms: when the party disbands, when it is kicked,
-when the leader enters a map it may not access, before the bot's own logout (the
-presence rotation or a fault restart stops it), or - once no live human is left
-in the party - ten minutes after its companion disconnects, so the bot is not
-stranded in a dead party forever.
+when the leader enters a map it may not access, ten minutes after every human
+companion has disconnected without reconnecting, or before the bot's own logout (the
+presence rotation or a fault restart stops it).
 
 A bot fights back when a player attacks it, but only as far as the game's own
 PvP rules allow: inside the active self-defense window, or against a player
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -255,8 +255,8 @@ party up if it was leading it, so a player never has to guess which bot happens
 to be free. In a party the bot follows its leader and defers a due reset, leaving
 only on the engine's own terms: when the party disbands, when it is kicked,
 when the leader enters a map it may not access, ten minutes after every human
-companion has disconnected without reconnecting, or before the bot's own logout (the
-presence rotation or a fault restart stops it).
+companion has disconnected without reconnecting, or before the bot's own logout
+(the presence rotation or a fault restart stops it).
 
 A bot fights back when a player attacks it, but only as far as the game's own
 PvP rules allow: inside the active self-defense window, or against a player
--- a/docs/Bots.md
+++ b/docs/Bots.md
@@ -233,7 +233,7 @@ requirement, the player-killer rule). A bot which does not qualify leaves the
 party and goes back to its own life instead of blocking the entry.
 
 Inside, its open-world routine is suspended: no shopping, no map changes, no
-boredom, no grudges. It fights what the event throws at it and keeps up with the
+grudges. It fights what the event throws at it and keeps up with the
 leader. Chaos Castle is a free-for-all, so there the other participants are
 targets like everyone else — and a fight inside leaves no grudge outside. A bot
 which dies respawns in the safezone like a player, which takes it out of the
@@ -248,13 +248,13 @@ experience bonus applies. Parties re-form every hour.
 A player may invite a bot into their own party: it accepts after a human-like
 pause of a few seconds, as long as it is not in the middle of an errand. There is
 no level gate — just like OpenMU's own party action, a bot accepts an inviter of
-any level, since it is the player who invites and the bot leaves once it gets
-bored. A living player takes precedence over the bot's own company
+any level, since it is the player who invites, and once it joins it stays for the
+whole session. A living player takes precedence over the bot's own company
 — a bot hunting with other bots leaves them for the inviter, and breaks that bot
 party up if it was leading it, so a player never has to guess which bot happens
-to be free. In a party the bot follows its leader, defers a due reset, and
-eventually leaves politely: when the leader enters a map it may not access,
-before its own logout, or simply when it gets bored.
+to be free. In a party the bot follows its leader and defers a due reset, leaving
+only on the engine's own terms: when the party disbands, when it is kicked, or
+when the leader enters a map it may not access.
 
 A bot fights back when a player attacks it, but only as far as the game's own
 PvP rules allow: inside the active self-defense window, or against a player
--- a/src/GameLogic/Bots/BotManager.cs
+++ b/src/GameLogic/Bots/BotManager.cs
@@ -230,7 +230,7 @@ public sealed class BotManager
         const int minPartySize = 2;
         const int maxPartySize = 5;
         const int maxLevelGap = 12;
-        const int partiedSharePercent = 60;
+        const int partiedSharePercent = 80;
 
         // Matched by the reset-aware effective level (see BotResetHandler.GetEffectiveLevel), so on a
         // reset server a freshly reset veteran groups with its peers instead of with real newbies.
--- a/src/GameLogic/Bots/BotNavigator.cs
+++ b/src/GameLogic/Bots/BotNavigator.cs
@@ -464,7 +464,7 @@ internal sealed class BotNavigator : AsyncDisposable
         }
 
         // Party bookkeeping: answer a pending invitation from a player once its human-like delay
-        // passed, and leave a party with a human again when the bot got bored (see BotPartyHandler).
+        // passed (see BotPartyHandler).
         await BotPartyHandler.ProcessAsync(this._player).ConfigureAwait(false);
 
         // Make sure the bot always carries healing and mana potions. Without them, the offline handlers
@@ -949,10 +949,6 @@ internal sealed class BotNavigator : AsyncDisposable
         // Fight from wherever the bot stands; the combat handler engages the targets around it.
         this._player.HuntingOrigin = this._player.Position;
 
-        // The party boredom timer must not run down (let alone fire) mid-event; the window
-        // restarts once the event is over.
-        this._player.PartyBoredomAtUtc = null;
-
         if (this._player.IsWalking)
         {
             return;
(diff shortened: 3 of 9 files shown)


sven-n:
## Review

The mechanical removal is clean and complete: `git grep -i bored` over `src`, `docs` and `tests` comes back empty, the mini-game safeguard that existed only to protect the timer goes with it, and the XML docs were rewritten instead of left stale. What follows is about the consequences of the removal, not the removal itself. #1 I consider a real bug, #2 and #3 are design decisions worth making explicitly, the rest are nits.

### 1. A bot is stranded in a dead party when its human companion logs out

This is where "stays for the whole session" meets the way the engine actually ends sessions. `Player.RemoveFromGameAsync` (`src/GameLogic/Player.cs:1263`) does not remove a disconnecting player from their party — it replaces them with an `OfflinePartyMember` snapshot via `Party.LeaveTemporarilyAsync` (`src/GameLogic/Party.cs:149`) to keep the slot reserved for a reconnect, and `PartyManager` never expires that snapshot.

For the bot left behind:

- `HasHumanCompanion` returns false — the snapshot is not a `Player` — so nothing in the bot code treats it as partied with a human any longer;
- `bot.Party` is still non-null, so the hourly re-formation never picks it up again: the candidate filter is `b.Party is null` (`src/GameLogic/Bots/BotManager.cs:238`);
- `GetPartyLeaderToFollow` returns null (the master is not a live `Player`), so it gets neither the follow behavior nor any party benefit.

The bot then hunts solo inside a one-member-plus-ghost party until it happens to be rotated out by the presence rotation, or until another player invites it. Before this PR the boredom timer usually fired first — 10–20 minutes is shorter than a typical play session — so the bot was normally back in circulation before its companion left. Now the zombie party is the *default* outcome of every player-bot party, because logging out is how sessions end. Over a long uptime this steadily drains the pool of party-less bots that `FormPartiesAsync` draws from, which works against the 60 → 80 % change in this same PR.

Suggestion: in `ProcessAsync` (`src/GameLogic/Bots/BotPartyHandler.cs:78`), let the bot leave a party that has no connected human member left and whose master is not a live player. That preserves the whole-session promise while the player is actually online, leaves bot-only parties untouched (their master is a live `BotPlayer`), and puts the bot back into circulation when the player is gone.

### 2. The reset and the master evolution are now deferred without a bound

Both progression gates skip while the bot has a human companion:

- `src/GameLogic/Bots/BotNavigator.cs:1147` (reset)
- `src/GameLogic/Bots/BotFeaturePlugIn.cs:479` (master evolution)

The old comment was explicit about what bounded that wait: "the boredom timer of BotPartyHandler bounds how long that takes". With the timer gone, the only bound is the party's lifetime. A bot parked in the party of an idle or AFK player sits at its reset level indefinitely — and at that level further hunting gains it nothing, so it simply stops progressing; the same holds for a level-400 bot waiting to evolve.

Since the permanent party is the goal here, the fix probably belongs on the deferral side rather than in a restored boredom timer: allow the reset/evolution once the deferral has lasted longer than some cap, or once the leader has been idle or out of range for a while (the follow code already tracks the leader's map and settle time).

### 3. "Abuse-free" versus an unlimited, permanent entourage

The class summary still claims the safeguards keep the feature "believable and abuse-free". With the timer gone there is no level gate, nothing caps how many bots a single player collects, and the bots never leave on their own. Party experience scales with the member count and derives its base from the party's *average* level (`CalculatePartyExperiencePerLevel`, `src/GameLogic/Party.cs:338`), so a permanent low-level entourage becomes a stable, zero-effort modifier on a player's exp curve rather than something they have to keep re-establishing. Boredom was weak friction — nothing stopped an immediate re-invite — but it was the only friction there was. Either cap the number of bots per player party, or drop "abuse-free" from that sentence.

### 4. The new leave-reason lists omit the bot's own logout

Both `docs/Bots.md` and the `BotPartyHandler` summary now say the bot leaves only "when the party disbands, when it is kicked, or when the leader enters a map it may not access". The presence rotation still pulls bots out from under players: `StopRandomBotAsync` (`src/GameLogic/Bots/BotManager.cs:137`) picks a random bot — including one partied with a human — and calls `KickMySelfAsync` before disconnecting it; the fault restart does the same. The previous text listed "before its own logout" for exactly this case, and from the player's side it is the most visible way a bot leaves. Worth putting back, since both lists read as exhaustive.

### 5. Minor

- The 60 → 80 % party share is an unrelated change riding along. Also, since `FormPartiesAsync` only considers party-less bots and bot parties are never disbanded, that constant governs how fast bots accumulate into parties rather than the steady-state share — which makes the neighbouring line "The rest of the bots keep hunting solo, so the population stays varied" (`src/GameLogic/Bots/BotManager.cs:225`) read increasingly optimistic at 80 %.
- `StaysInPartyWithHumanAsync` is tautological: with no pending invite, `ProcessAsync` is a no-op, so the test passes even against an empty method body. Fine as a marker against re-introduction, but it is not coverage of how long the bot stays.
- On the description's testability argument: `TryScheduleAcceptAsync` already carries an `acceptDelay` override for tests, so the boredom window could have been injected the same way. Removing the behavior is a defensible product call on its own — it just wasn't forced by testability.

---

sven-n:
## Re-review — `0933e41` ("add PartyDisconnectGracePeriod to BotPartyHandler")

The grace period is the right shape for the stranding problem, and the docs now list the bot's own logout again. CI is green on the new head. Three things about the new code, though; the first is the one I'd act on.

### A. The `IsOffLeveling` guard looks unreachable — and if it ever became reachable it would re-open the bug it sits in

`PartyAutoRejoinPlugIn` hooks `IPlayerStateChangedPlugIn` and calls `PartyManager.OnMemberReconnectedAsync` for *any* player that reaches `PlayerState.EnteredWorld` — the off-level ghost included, because `OfflinePlayer.InitializeAsync` → `ClientReadyAfterMapChangeAsync` walks the same world-entry path (`src/GameLogic/Player.cs:983`), and the state-changed hook is wired in the `Player` constructor for every player. It is active unless a configuration row explicitly sets `IsActive = false` (`src/PlugIns/PlugInManager.cs:451`).

So once an off-level session is actually running, the snapshot has already been swapped for the live `OfflinePlayer` and `party.PartyMaster is OfflinePartyMember` is false — the `IsOffLeveling` branch is never reached. The only window in which both conditions hold is between the real player's disconnect inside `TransitionToOfflineAsync` and the ghost's world entry a moment later, which the 10-minute grace covers anyway.

The mirror image is worse: on a server where that plug-in is switched off, the guard *does* become reachable, and it then keeps the bot parked in the dead party for the entire off-level session — hours or days. That is exactly the stranding this commit is fixing.

There is also a corollary worth knowing, independent of the guard: while the human off-levels, their ghost is an `OfflinePlayer`, so `HasHumanCompanion` is false. The bot will accept another player's invitation and walk away (`TryScheduleAcceptAsync` → `LeaveBotPartyAsync`), and a due reset can fire and teleport it home. An off-leveling player does not hold their bot, which is the opposite of what the guard's comment intends.

Suggestion: drop the `IsOffLeveling` check, the new public `OfflinePlayerManager.IsOffLeveling` and its two tests — the grace period alone covers the transition. If bots *should* stay with a player who switched to off-leveling, that is a separate behavior to write deliberately (and it currently half-happens by accident).

### B. Keying off the master slot lets one dead-party shape through

The condition is `party.PartyMaster is OfflinePartyMember`, but mastership moves: `ExitPartyAsync` reassigns it to `_partyMembers[0]` when the master leaves properly. Take a party joined in the order player P (master), bot B, player Q. P leaves through the normal party-exit action → members are `[B, Q]` and the master is now **B**. Q then disconnects and becomes a snapshot. There is no live human left, but the master is a live bot, so the new branch never fires and B is stranded with the ghost exactly as before.

Narrower than the original case, same failure. Deciding on "no live human left, and the party still holds a human's snapshot" rather than on the master's type would cover both — e.g. `!HasHumanCompanion(bot)` plus `party.PartyList.OfType<OfflinePartyMember>()`, taking the newest `DisconnectedAtUtc`. Bot-only parties carry no such snapshot as long as bots kick themselves before stopping, which both stop paths do (`StopRandomBotAsync`, `RemoveBotAsync`).

### C. Both leave-reason lists are exhaustive again, and again incomplete

The class summary still opens with "A bot stays in the party with its human companion for as long as the party exists", and neither its list nor the one in `docs/Bots.md` mentions the new reason: the bot also leaves ten minutes after its companion disconnects. Since the previous round of edits was about making those lists match reality, this one belongs in them too.

### Nits

- Put the cheap check first: compare `DisconnectedAtUtc` before calling `IsOffLeveling`, which scans `_activePlayers` on every tick of every bot sitting in a dead party. (Moot if A is applied.)
- `OfflinePlayerManager.IsOffLeveling(string characterName)` takes a **character** name, while `IsActive`, `TryGetPlayer` and `StopAsync` next to it all take **login** names and the dictionary is keyed by login name. A caller passing a login name gets a silent `false`. `IsCharacterOffLeveling`, or at least an XML remark, would prevent that.
- `DisconnectedAtUtc { get; internal set; }` exists only so the test can back-date it, on an otherwise immutable public snapshot type. `TryScheduleAcceptAsync`'s optional `acceptDelay` is the convention this file already has for that — an optional grace (or "now") parameter on `ProcessAsync` would do the same job without the settable property.

### Still open from the first round

- **The unbounded deferral of the reset and the master evolution** while the companion is *online* (`BotNavigator.cs:1147`, `BotFeaturePlugIn.cs:479`). The grace period fixes the disconnected case for free — `HasHumanCompanion` goes false the moment the snapshot appears — but an idle or AFK player still parks a bot at its reset level indefinitely.
- The 60 → 80 % share is still an unrelated change riding along, and the "the rest of the bots keep hunting solo" line above it (`BotManager.cs:225`) still describes a rate rather than a steady state.

Thanks for turning the first round around so quickly — the grace-period approach itself is the right call.

---

sven-n:
## Re-review — `78a064ff` ("address code review points")

All three points from the last round are addressed, and correctly: `IsOffLeveling` and its tests are gone, the condition now keys off "no live human left plus a lingering snapshot" instead of the master slot, the `internal set` gave way to the `utcNow` seam that matches `acceptDelay`, and both leave-reason lists mention the grace period. I traced `LeavesPartyWhenMasterIsBotButHumanSnapshotRemainsAsync` by hand — `ExitPartyAsync` does hand the master slot to `_partyMembers[0]`, so the party really does end up bot-mastered with a human snapshot seated in it, and the new condition catches it where the old one did not. Good test. CI is still running on this head.

One new thing, and it is the same complaint this PR opened with.

### The new check runs the heaviest work on the hot path, for the bots that never need it

```csharp
if (bot.Party is { } party && !HasHumanCompanion(bot))
{
    var mostRecentlyDisconnectedHuman = party.PartyList
        .OfType<OfflinePartyMember>()
        .OrderByDescending(member => member.DisconnectedAtUtc)
        .FirstOrDefault();
```

`HasHumanCompanion` is false for **every bot in a bot-only party**, which after the 60 → 80 % change is most of them. So for those bots the guard falls through into the body on every tick — `EvaluationInterval` is one second (`BotNavigator.cs:159`) — and each pass allocates an `OfType` iterator plus `OrderByDescending`'s buffer and key array, sorts a list that is almost always empty, and returns because `mostRecentlyDisconnectedHuman is null`. At a few hundred bots that is a few hundred throwaway sorts per second.

The PR description's case against the boredom timer was precisely this: "Every bot touched this wall-clock state on every evaluation tick, even bot-only parties that never used it." A null assignment has been replaced with a LINQ sort on the same path. Since only the newest `DisconnectedAtUtc` matters, a single `foreach` over `PartyList` tracking the max — bailing out the moment it finds none — does the same job allocation-free and reads no worse:

```csharp
DateTime? lastDisconnect = null;
foreach (var member in party.PartyList)
{
    if (member is OfflinePartyMember snapshot
        && (lastDisconnect is null || snapshot.DisconnectedAtUtc > lastDisconnect))
    {
        lastDisconnect = snapshot.DisconnectedAtUtc;
    }
}
```

### Nits

- The `ProcessAsync` summary still reads "Answers a pending invitation once its delay passed" — the method now also walks a bot out of a dead party. The `<param>` for `utcNow` was added right below it, so the doc was already in hand.
- `"leaves the party of the disconnected player '{Master}'"`: the property is named `{Master}`, but the point of this round's fix is that the member need not be the master — in the new test it explicitly is not. `{Player}` would keep the log honest. The local `mostRecentlyDisconnectedHuman` has the same slant; nothing restricts the snapshot to a human (a bot's own snapshot can survive a shutdown-path stop), so `mostRecentlyDisconnectedMember` is closer to what it holds.
- The comment above the block has a line break mid-sentence ("This is independent of who currently holds the / master slot") and then joins two unrelated claims with a dash. Two sentences would read cleaner.

### Worth knowing, not necessarily changing

With `IsOffLeveling` gone, a companion who switches to off-leveling still keeps their bot — the ghost rejoins the party as a live `OfflinePlayer`, so no snapshot exists and the new branch returns on `mostRecentlyDisconnectedHuman is null`. That is the same outcome the guard was reaching for, now reached implicitly. The flip side is unchanged: `HasHumanCompanion` is false for that ghost, so the bot can still be poached by another inviter and a due reset can still fire mid-party. Fine as a default — just no longer stated anywhere.

### Still open from round one

- **The unbounded deferral of the reset and the master evolution** while the companion is genuinely online (`BotNavigator.cs:1147`, `BotFeaturePlugIn.cs:479`). The grace period covers the disconnected case; an idle player parked in town still holds a bot at its reset level indefinitely.
- The 60 → 80 % share remains an unrelated change riding along, with "the rest of the bots keep hunting solo" (`BotManager.cs:225`) still describing a rate rather than a steady state.

Neither blocks this PR in my view — the party behavior itself is in good shape now.

---

eduardosmaniotto:
`The unbounded deferral of the reset and the master evolution while the companion is genuinely online (BotNavigator.cs:1147, BotFeaturePlugIn.cs:479). The grace period covers the disconnected case; an idle player parked in town still holds a bot at its reset level indefinitely.`

I think this needs a step back. What is the final purpose of bots? I would say stress testing and helping out starting players that have no one to play with and don't want to run 5 simultaneous accounts. 

As long as the player is online, there should be no need for the bot to abandon them; they are not competing to be the top players in the server.

sven-n:
Alright, then I think it's ready to merge :D

PR #887, opened 2026-08-18
Fix concurrency crash in attribute value recalculation

### LINKED ISSUE #883: Race condition in `ComposableAttribute`: `NullReferenceException` when a magic effect expires while an attribute is recalculated
## Summary

`ComposableAttribute` stores its elements in a plain, unsynchronized `List<IElement>` and enumerates it live while recomputing a value. Magic-effect expiration mutates that same list (`RemoveElement`) from a `System.Threading.Timer` callback on a thread-pool thread, without any lock shared with the reader. When an effect expires at the exact moment an attribute in the same dependency chain is being recalculated, the concurrent `List` mutation transiently exposes a `null` slot to the enumeration, throwing a `NullReferenceException`.

It is a long-standing latent thread-safety gap (the unsynchronized element list dates back years), but it was effectively never hit until server-side offline bots started reading attributes in a tight loop and buffing themselves. The offline MU Helper catches the exception, so bots keep running, but the server log is flooded with errors. Online players are theoretically exposed to the same race (far less frequently), and for them the exception is **not** caught by the helper's safety net.

## Environment

- Branch: `master`
- Affected files: `src/AttributeSystem/ComposableAttribute.cs`, `src/AttributeSystem/AttributeSystem.cs`, `src/GameLogic/MagicEffectsList.cs`, `src/GameLogic/MagicEffect.cs`

## Observed error

Logged repeatedly (bursts of several bots at once), swallowed by `OfflinePlayerMuHelper.SafeTickAsync`:

```
[Error] [MUnique.OpenMU.GameLogic.Bots.BotPlayer] Error in offline player helper tick for "bot0019".
System.NullReferenceException: Object reference not set to an instance of an object.
   at MUnique.OpenMU.AttributeSystem.ComposableAttribute.<>c.<GetAndCacheValue>b__10_0(IElement e) in ComposableAttribute.cs:line 65
   at MUnique.OpenMU.AttributeSystem.ComposableAttribute.GetAndCacheValue() in ComposableAttribute.cs:line 59
   at MUnique.OpenMU.AttributeSystem.AttributeRelationshipElement.CalculateValue() in AttributeRelationshipElement.cs:line 70
   at MUnique.OpenMU.AttributeSystem.AttributeRelationshipElement.get_Value() in AttributeRelationshipElement.cs:line 54
   ... (nested AttributeRelationshipElement / ComposableAttribute frames) ...
   at MUnique.OpenMU.GameLogic.Offline.HealingHandler.PerformSelfHealingAsync()
   at MUnique.OpenMU.GameLogic.Offline.HealingHandler.PerformHealthRecoveryAsync() in HealingHandler.cs:line 73
   at MUnique.OpenMU.GameLogic.Offline.OfflinePlayerMuHelper.TickAsync(...) in OfflinePlayerMuHelper.cs:line 207
```

The leaf frame is the `Where` predicate in `GetAndCacheValue` (`e => e.AggregateType == ...`) dereferencing a `null` element `e`.

## Root cause

1. **Unsynchronized shared list, live enumeration (reader).**
   `ComposableAttribute` keeps `private readonly IList<IElement> _elementList = new List<IElement>();` and `GetAndCacheValue()` enumerates it directly:

   ```csharp
   var rawValues = this.Elements.Where(e => e.AggregateType == AggregateType.AddRaw).Sum(e => e.Value);
   ```

   `this.Elements` returns the live `_elementList`; there is no snapshot and no lock.

2. **Off-thread mutation (writer).**
   A `MagicEffect` schedules expiry with `System.Threading.Timer` (`MagicEffect.cs`). On timeout, `MagicEffectsList.OnEffectTimeOutAsync` runs on a thread-pool thread and calls `this._owner.Attributes.RemoveElement(powerUp.Element, powerUp.Target)` — **after releasing `_addLock`** (`MagicEffectsList.cs`). `AttributeSystem.RemoveElement` forwards to `ComposableAttribute.RemoveElement` → `_elementList.Remove(element)`. Nothing serializes this against readers.

3. **The race.**
   `List<T>.Remove` shifts the backing array and clears the last slot. A reader enumerating with the pre-shift `_size` can observe a `null` in that slot, so `e.AggregateType` throws `NullReferenceException`. Because this happens deep in a chain of `AttributeRelationshipElement`s (e.g. Maximum Health / Current Health derivations), any expiring buff that contributes to the recalculated attribute can trigger it.

**Ruled out (not a data problem):** a buff never contributes a `null` element — `MagicEffectPowerUpExtensions` throws when `Boost` is missing and `CreateElement` never returns `null`; the leaf element types (`SimpleElement`, `ConstantElement`) have trivial `Value` getters that cannot throw. The `null` only exists transiently during concurrent mutation.

## How to reproduce / why it surfaced now

Any holder that (a) reads an attribute frequently and (b) has magic effects expiring concurrently will eventually hit it. Server-side offline bots make this reliable:

- The offline MU Helper ticks every ~500 ms and reads `MaximumHealth` / `CurrentHealth` (health-threshold checks in `HealingHandler`).
- Bots buff themselves, so they hold magic effects that expire on timers.
- Buffs applied around the same time expire around the same time → several timers fire in one window → multiple bot ticks catch the torn read together (hence the bursts).

Before offline bots, offline players had no self-buffs and no tight read loop, so the pre-existing race was essentially never exercised.

## Impact

- **Bots:** non-fatal. `OfflinePlayerMuHelper.SafeTickAsync` catches the exception, logs it, and the bot continues; the affected 500 ms tick simply skips healing. Main harm is log spam (can fill disk on busy servers).
- **Online players:** the same race is possible (e.g. damage calculation reading defense/max HP as a buff expires). It is much rarer, but there is **no** MU-Helper safety net around a player's attribute reads, so the exception can bubble up into the action/packet handler.

## Suggested fix

Make `ComposableAttribute` thread-safe, since it is the shared point where both the mutation and the enumeration meet:

- Guard `AddElement` / `RemoveElement` and the enumeration inside `GetAndCacheValue` with a private lock (`lock` / `System.Threading.Lock`), or
- At minimum, snapshot the list before enumerating in `GetAndCacheValue` (`_elementList.ToArray()`), combined with serialized mutations so two concurrent writers cannot corrupt the list.

The lock overhead is negligible: it is only taken on element add/remove and on an actual recompute (cache miss); reads that hit `_cachedValue` don't touch the list. Fixing it in `ComposableAttribute` protects every attribute holder (players included), not just the offline path.

A complementary hardening would be to perform the `Attributes.RemoveElement` loop in `MagicEffectsList.OnEffectTimeOutAsync` under the same lock used for the effect bookkeeping, so effect expiry never mutates attribute state lock-free.

nolt:
Fixes #883.

## Problem

`ComposableAttribute` keeps its elements in a plain, unsynchronized `List<IElement>` and enumerates it live while recomputing a value (`GetAndCacheValue`). A `MagicEffect` schedules its expiry on a `System.Threading.Timer`; when it fires, `MagicEffectsList.OnEffectTimeOutAsync` runs on a thread-pool thread and calls `Attributes.RemoveElement(...)`, mutating that same list with no lock shared with the reader.

When an effect expires at the moment an attribute in the same dependency chain is recalculated, the concurrent `List<T>.Remove` shifts the backing array and the reader can observe a transiently `null` slot, throwing a `NullReferenceException` at the `Where(e => e.AggregateType == ...)` predicate.

This was effectively never hit until server-side offline bots started reading attributes in a tight loop (the offline MU Helper ticks ~every 500 ms and reads `MaximumHealth`/`CurrentHealth`) while buffing themselves (buffs expire on timers). On a server with many offline bots the log fills with:

```
[Error] [MUnique.OpenMU.GameLogic.Bots.BotPlayer] Error in offline player helper tick for "botNNNN".
System.NullReferenceException: Object reference not set to an instance of an object.
   at MUnique.OpenMU.AttributeSystem.ComposableAttribute.<>c.<GetAndCacheValue>b__10_0(IElement e) in ComposableAttribute.cs:line 65
   at MUnique.OpenMU.AttributeSystem.ComposableAttribute.GetAndCacheValue()
   at MUnique.OpenMU.AttributeSystem.AttributeRelationshipElement.CalculateValue()
   ... (nested AttributeRelationshipElement / ComposableAttribute frames) ...
   at MUnique.OpenMU.GameLogic.Offline.HealingHandler.PerformSelfHealingAsync()
   at MUnique.OpenMU.GameLogic.Offline.OfflinePlayerMuHelper.TickAsync(...)
```

The offline helper's `SafeTickAsync` swallows the exception, so bots keep running, but online players are exposed to the same race (e.g. damage calculation reading defense/max HP as a buff expires) with no such safety net.

`AttributeRelationshipElement` — the element type that sits between attributes in the dependency chain — has the **same class of race** in a second place: its `GetAndCacheValue` did `this._cachedValue = this.CalculateValue(); return this._cachedValue.Value;`, and a concurrent `ElementChanged` (from an input attribute whose element was just removed) could reset `_cachedValue` to `null` between the assignment and the read, making `_cachedValue.Value` throw `InvalidOperationException: Nullable object must have a value`.

## Fix

- **`ComposableAttribute`** — the element list is now **copy-on-write** (`private volatile IElement[] _elements`): mutations replace the whole array under a lock, so a reader can aggregate an immutable snapshot with a single volatile read. The cached fast path of `Value` and the `Elements` getter are therefore lock-free and allocation-free; only a cache miss briefly takes the lock (to capture a consistent snapshot and to write the cache) and the aggregation itself runs outside the lock. Element subscribe/unsubscribe now happen inside the lock, so a concurrent add/remove of the same element cannot leave a live subscription on an element no longer in the list.
- **Cache coherency (version counter)** — both `ComposableAttribute` and `AttributeRelationshipElement` now carry a version counter, bumped under the lock whenever the composition changes or the cache is invalidated. A recomputation captures the version before aggregating and only writes its result into `_cachedValue` if the version is still unchanged. This closes a lost-update where a recomputation that raced a concurrent removal would **latch a stale value into the cache** and keep serving it until some unrelated change happened to invalidate it (e.g. an offline bot sitting on an expired buff's value indefinitely). It also removes the `InvalidOperationException` in `AttributeRelationshipElement` by returning the freshly computed local value.
- **`AttributeSystem`** — every access to the `_attributes` dictionary is now guarded by a lock. It was a plain `Dictionary` mutated from the magic-effect timer thread (`RemoveElement`) while read on the game-logic thread — the same class of race, with a worse failure mode (undefined behavior / an infinite spin inside a resizing lookup). The get-or-create is now atomic, and an empty attribute is only dropped when it is still empty **and** still the same instance (a `ReferenceEquals` guard). Element addition/removal and the value-changed dispatch stay **outside** the lock, so it is never held across a call that could reach another attribute system.

### Known follow-ups (deliberately out of scope for this crash fix)

- The fast-path read of the cached value (`Value => this._cachedValue ?? ...`) is still unsynchronized. A torn or late read of the 8-byte `Nullable<float>` can return a transiently wrong value; it is self-correcting and never crashes. Making the read strictly coherent would need `volatile` (not possible on `Nullable<float>`) or a lock on every read (contention on a very hot path).
- An empty attribute can still be dropped from `AttributeSystem._attributes` in a narrow window while another thread is adding an element to it, orphaning that contribution. Closing this cleanly requires holding the lock across `composable.AddElement`, whose dispatch synchronously raises the public `AttributeValueChanged` event — a reentrancy/cross-system-lock risk not worth taking in a crash fix.
- Relatedly, `RemoveElement` does not call `OnAttributeRemoved` when it drops an emptied attribute (pre-existing), so an `ItemAwareAttributeSystem` subscription on that instance is only cleaned up at `Dispose`.

These are safe to address in a separate change if they ever prove worthwhile.

## Tests

`MUnique.OpenMU.AttributeSystem.Tests`: **44/44**.

- `ConcurrentAddRemoveWhileReadingValueDoesNotThrow` — single attribute; reproduces the `ComposableAttribute` list race. Reads `Value` on one thread while another adds/removes elements.
- `ConcurrentRecalculationOfDependentAttributesDoesNotDeadlock` — a parent attribute whose value depends on a child attribute (via `AttributeRelationshipElement`) that is mutated concurrently; reproduces the nested race. Both concurrency tests fail with a 30 s timeout instead of hanging should the locking ever deadlock.
- `ConcurrentRemovalDuringReadDoesNotLatchStaleValue` — deterministic (single-threaded, reentrant): an element removes another mid-aggregation, and the test asserts the pre-removal value is not latched into the cache. Confirmed to fail (`Expected 0, but was 100`) without the version guard.

The two threaded tests were confirmed to fail (NRE / `Collection was modified` / `InvalidOperationException`) against the unpatched code and to pass with the fix; the suite was run repeatedly to confirm there is no flakiness.

## Affected files

- `src/AttributeSystem/ComposableAttribute.cs`
- `src/AttributeSystem/AttributeRelationshipElement.cs`
- `src/AttributeSystem/AttributeSystem.cs`
- `src/AttributeSystem/IComposableAttribute.cs` (doc note that `Elements` may be a snapshot)
- `tests/MUnique.OpenMU.AttributeSystem.Tests/ComposableAttributeTests.cs`

### DIFF
--- a/src/AttributeSystem/AttributeRelationshipElement.cs
+++ b/src/AttributeSystem/AttributeRelationshipElement.cs
@@ -61,8 +61,13 @@ public class AttributeRelationshipElement : SimpleElement
 
     private float GetAndCacheValue()
     {
-        this._cachedValue = this.CalculateValue();
-        return this._cachedValue.Value;
+        // Return the freshly computed value rather than re-reading the field: a concurrent ElementChanged
+        // (e.g. from an input attribute whose element is removed when a magic effect expires) can reset
+        // _cachedValue to null between the assignment and the read, which would make _cachedValue.Value
+        // throw an InvalidOperationException.
+        var value = this.CalculateValue();
+        this._cachedValue = value;
+        return value;
     }
 
     private float CalculateValue()
--- a/src/AttributeSystem/AttributeRelationshipElement.cs
+++ b/src/AttributeSystem/AttributeRelationshipElement.cs
@@ -10,6 +10,15 @@ namespace MUnique.OpenMU.AttributeSystem;
 /// </summary>
 public class AttributeRelationshipElement : SimpleElement
 {
+    /// <summary>
+    /// Serializes the <see cref="_version"/> bump with the guarded cache write, so a recalculation that
+    /// races a concurrent input change (e.g. an input attribute whose element is removed when a magic
+    /// effect expires) cannot latch its stale result into <see cref="_cachedValue"/>.
+    /// </summary>
+    private readonly object _cacheLock = new();
+
+    private long _version;
+
     private float? _cachedValue;
 
     /// <summary>
@@ -55,7 +64,12 @@ public class AttributeRelationshipElement : SimpleElement
 
     private void ElementChanged(object? sender, EventArgs eventArgs)
     {
-        this._cachedValue = null;
+        lock (this._cacheLock)
+        {
+            this._version++;
+            this._cachedValue = null;
+        }
+
         this.RaiseValueChanged();
     }
 
@@ -64,9 +78,25 @@ public class AttributeRelationshipElement : SimpleElement
         // Return the freshly computed value rather than re-reading the field: a concurrent ElementChanged
         // (e.g. from an input attribute whose element is removed when a magic effect expires) can reset
         // _cachedValue to null between the assignment and the read, which would make _cachedValue.Value
-        // throw an InvalidOperationException.
+        // throw an InvalidOperationException. The version captured before calculating guards the cache
+        // write: if an input changed while CalculateValue was running, the fresh value is still returned
+        // to this caller but not cached, so the lost update cannot latch a stale value.
+        long versionAtStart;
+        lock (this._cacheLock)
+        {
+            versionAtStart = this._version;
+        }
+
         var value = this.CalculateValue();
-        this._cachedValue = value;
+
+        lock (this._cacheLock)
+        {
+            if (this._version == versionAtStart)
+            {
+                this._cachedValue = value;
+            }
+        }
+
         return value;
     }
 
--- a/src/AttributeSystem/AttributeSystem.cs
+++ b/src/AttributeSystem/AttributeSystem.cs
@@ -14,6 +14,17 @@ public class AttributeSystem : IAttributeSystem, IEnumerable<IAttribute>
     private readonly IDictionary<AttributeDefinition, IAttribute> _attributes = new Dictionary<AttributeDefinition, IAttribute>();
 
     /// <summary>
+    /// Synchronizes every access to <see cref="_attributes"/>. The dictionary is read on the game logic
+    /// thread (damage calculations, stat lookups) while an attribute can be removed from a thread pool
+    /// thread when a magic effect expires on its timer (see <see cref="RemoveElement"/>). Concurrent
+    /// mutation of a plain <see cref="Dictionary{TKey,TValue}"/> during reads is undefined - wrong
+    /// results, exceptions, or an infinite spin inside a resizing lookup. Only the dictionary operations
+    /// are guarded; element aggregation and value-changed dispatch run outside this lock, so it never
+    /// nests into another attribute system's lock and cannot deadlock.
+    /// </summary>
+    private readonly object _attributesLock = new();
+
+    /// <summary>
     /// Initializes a new instance of the <see cref="AttributeSystem" /> class.
     /// </summary>
     /// <param name="statAttributes">The stat attributes. These attributes are added just as-is and are not wrapped by a <see cref="ComposableAttribute"/>.</param>
@@ -86,8 +97,13 @@ public class AttributeSystem : IAttributeSystem, IEnumerable<IAttribute>
             return false;
         }
 
-        if (this._attributes.TryGetValue(attributeDefinition, out var attribute)
-            && attribute is StatAttribute statAttribute)
+        IAttribute? attribute;
+        lock (this._attributesLock)
+        {
+            this._attributes.TryGetValue(attributeDefinition, out attribute);
+        }
+
+        if (attribute is StatAttribute statAttribute)
         {
             statAttribute.Value = newValue;
 
@@ -128,39 +144,62 @@ public class AttributeSystem : IAttributeSystem, IEnumerable<IAttribute>
     /// <inheritdoc/>
     public void AddElement(IElement element, AttributeDefinition targetAttribute)
     {
-        if (!this._attributes.TryGetValue(targetAttribute, out var attribute))
+        // The get-or-create is atomic under the lock so two threads racing the first add of the same
+        // attribute cannot both create it (which would make the second Dictionary.Add throw). The
+        // element is then added - and its ValueChanged dispatched - outside the lock, so the lock is
+        // never held across a call that could reach another attribute system's lock.
+        IComposableAttribute composableAttribute;
+        ComposableAttribute? created = null;
+        lock (this._attributesLock)
         {
-            attribute = new ComposableAttribute(targetAttribute);
-            this._attributes.Add(targetAttribute, attribute);
-            this.OnAttributeAdded(attribute);
+            if (this._attributes.TryGetValue(targetAttribute, out var attribute))
+            {
+                composableAttribute = attribute as IComposableAttribute
+                    ?? throw new ArgumentException($"Attribute {targetAttribute} is not a composable attribute.");
+            }
+            else
+            {
+                var composable = new ComposableAttribute(targetAttribute);
+                this._attributes.Add(targetAttribute, composable);
+                composableAttribute = composable;
+                created = composable;
+            }
         }
 
-        if (attribute is IComposableAttribute composableAttribute)
+        if (created is not null)
         {
-            composableAttribute.AddElement(element);
+            this.OnAttributeAdded(created);
         }
-        else
-        {
-            throw new ArgumentException($"Attribute {targetAttribute} is not a composable attribute.");
-        }
+
+        composableAttribute.AddElement(element);
     }
 
     /// <inheritdoc/>
     public void RemoveElement(IElement element, AttributeDefinition targetAttribute)
     {
-        if (this._attributes.TryGetValue(targetAttribute, out var attribute))
+        IComposableAttribute composableAttribute;
+        lock (this._attributesLock)
         {
-            if (attribute is IComposableAttribute composableAttribute)
+            if (!this._attributes.TryGetValue(targetAttribute, out var attribute))
             {
-                composableAttribute.RemoveElement(element);
-                if (!composableAttribute.Elements.Any())
-                {
-                    this._attributes.Remove(targetAttribute);
-                }
+                return;
             }
-            else
+
+            composableAttribute = attribute as IComposableAttribute
+                ?? throw new ArgumentException($"Attribute {targetAttribute} is not a composable attribute.");
+        }
+
+        composableAttribute.RemoveElement(element);
+
+        lock (this._attributesLock)
+        {
+            // Only drop the attribute when it is still empty and still the very instance we just mutated,
+            // so a concurrent AddElement that repopulated it does not get its element discarded (TOCTOU).
+            if (!composableAttribute.Elements.Any()
+                && this._attributes.TryGetValue(targetAttribute, out var current)
+                && ReferenceEquals(current, composableAttribute))
             {
-                throw new ArgumentException($"Attribute {targetAttribute} is not a composable attribute.");
+                this._attributes.Remove(targetAttribute);
             }
         }
     }
@@ -168,15 +207,16 @@ public class AttributeSystem : IAttributeSystem, IEnumerable<IAttribute>
     /// <inheritdoc/>
     public override string ToString()
     {
+        var snapshot = this.GetAttributeSnapshot();
         var stringBuilder = new StringBuilder();
         stringBuilder.AppendLine("Stat Attributes:");
-        foreach (var statAttribute in this._attributes.Values.OfType<StatAttribute>())
+        foreach (var statAttribute in snapshot.OfType<StatAttribute>())
         {
             stringBuilder.AppendLine($"  {statAttribute.Definition}: {statAttribute.Value}");
         }
 
         stringBuilder.AppendLine("Others:");
-        foreach (var attribute in this._attributes.Values.OfType<IComposableAttribute>())
+        foreach (var attribute in snapshot.OfType<IComposableAttribute>())
         {
             stringBuilder.AppendLine($"  {attribute.Definition}: {attribute.Value}");
         }
@@ -187,7 +227,7 @@ public class AttributeSystem : IAttributeSystem, IEnumerable<IAttribute>
     /// <inheritdoc />
     public IEnumerator<IAttribute> GetEnumerator()
     {
-        return this._attributes.Values.GetEnumerator();
+        return ((IEnumerable<IAttribute>)this.GetAttributeSnapshot()).GetEnumerator();
     }
 
     /// <inheritdoc />
@@ -203,13 +243,26 @@ public class AttributeSystem : IAttributeSystem, IEnumerable<IAttribute>
     /// <returns>The element of the attribute.</returns>
     public IElement GetOrCreateAttribute(AttributeDefinition attributeDefinition)
     {
-        var element = this.GetAttribute(attributeDefinition);
-        if (element is null)
+        IElement element;
+        ComposableAttribute? created = null;
+        lock (this._attributesLock)
         {
-            var composableAttribute = new ComposableAttribute(attributeDefinition);
-            element = composableAttribute;
-            this._attributes.Add(attributeDefinition, composableAttribute);
-            this.OnAttributeAdded(composableAttribute);
+            if (this._attributes.TryGetValue(attributeDefinition, out var existing))
+            {
+                element = existing;
+            }
+            else
+            {
+                var composableAttribute = new ComposableAttribute(attributeDefinition);
+                this._attributes.Add(attributeDefinition, composableAttribute);
+                element = composableAttribute;
+                created = composableAttribute;
+            }
+        }
+
+        if (created is not null)
+        {
+            this.OnAttributeAdded(created);
         }
 
         return element;
@@ -249,11 +302,17 @@ public class AttributeSystem : IAttributeSystem, IEnumerable<IAttribute>
             return null;
         }
 
-        if (this._attributes.TryGetValue(attributeDefinition, out var attribute))
+        lock (this._attributesLock)
         {
-            return attribute;
+            return this._attributes.TryGetValue(attributeDefinition, out var attribute) ? attribute : null;
         }
+    }
 
-        return null;
+    private IAttribute[] GetAttributeSnapshot()
+    {
+        lock (this._attributesLock)
+        {
+            return this._attributes.Values.ToArray();
+        }
     }
 }
--- a/src/AttributeSystem/ComposableAttribute.cs
+++ b/src/AttributeSystem/ComposableAttribute.cs
@@ -9,16 +9,32 @@ namespace MUnique.OpenMU.AttributeSystem;
 /// </summary>
 public class ComposableAttribute : BaseAttribute, IComposableAttribute
 {
-    private readonly IList<IElement> _elementList;
-
     /// <summary>
-    /// Synchronizes access to <see cref="_elementList"/>. Elements are usually added on the game logic
-    /// thread, but they can also be removed from a thread pool thread when a magic effect expires on its
-    /// timer, while a value recalculation may be enumerating the list at the same time. Without this lock
-    /// the concurrent <see cref="List{T}.Remove"/> can transiently expose a <c>null</c> slot to the
-    /// enumeration in <see cref="GetAndCacheValue"/>, causing a <see cref="NullReferenceException"/>.
+    /// Serializes the copy-on-write mutations of <see cref="_elements"/>, the element subscription
+    /// changes and the <see cref="_version"/> bump. Elements are usually added on the game logic thread,
+    /// but they can also be removed from a thread pool thread when a magic effect expires on its timer,
+    /// while a value recalculation is aggregating at the same time. The cached fast path of
+    /// <see cref="Value"/> and the <see cref="Elements"/> getter are lock-free (a single volatile read of
+    /// <see cref="_elements"/>); on a cache miss, <see cref="GetAndCacheValue"/> takes this lock only
+    /// briefly to capture a consistent (snapshot, version) pair and again to write the cache, and does the
+    /// aggregation itself outside the lock on the immutable snapshot.
     /// </summary>
     private readonly object _elementLock = new();
+
+    /// <summary>
+    /// The elements this attribute is composed of. Treated as immutable: mutations replace the whole
+    /// array under <see cref="_elementLock"/> (copy-on-write), so a reader that captured the reference
+    /// can keep enumerating it safely while another thread adds or removes an element.
+    /// </summary>
+    private volatile IElement[] _elements = [];
+
+    /// <summary>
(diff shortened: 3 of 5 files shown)


sven-n:
## Review

The diagnosis is right and the fix is minimal and well-documented. I reproduced everything locally against head `2e59c0c`:

- Full `MUnique.OpenMU.AttributeSystem.Tests` on the PR head: **43/43 pass** (~150 ms).
- Reverting *only* `ComposableAttribute.cs` + `AttributeRelationshipElement.cs` to `master` and re-running the two new tests: both fail reproducibly (`NullReferenceException` / `Collection was modified; enumeration operation may not execute`). So these are real regression tests, not decoration.

I also confirmed the call path the description claims: `MagicEffect._finishTimer` (`MagicEffect.cs:47`) → `OnTimerTimeout` on a thread-pool thread → `MagicEffectsList.OnEffectTimeOutAsync` (`MagicEffectsList.cs:154`) → `Attributes.RemoveElement(...)`. That is genuinely off the game-logic thread.

A few things below. Nothing here blocks the crash fix — item 1 is a correction to the PR description's scope claim, item 2 is an incompleteness worth naming explicitly.

---

### 1. The stale-cache claim in "Scope / deliberate non-goal" is not accurate — the stale value is *latched*, not self-correcting

The description says:

> an occasional stale or torn read of the cached value can still happen and corrects itself on the next recalculation

There is no next recalculation. The interleaving is:

1. Reader takes the snapshot (still contains the buff element) and starts aggregating.
2. Writer removes the buff → `ElementChanged` → `_cachedValue = null`.
3. Reader finishes and writes its pre-removal result into `_cachedValue`.

From then on every read is a **cache hit on the stale value**, so the expired buff stays applied until some *unrelated* element change happens to invalidate the cache again.

I made this deterministic rather than relying on timing, using an element that mutates the attribute while its `Value` is being read (exactly what step 2 does, just without the thread):

```csharp
private sealed class EvilElement : IElement
{
    private readonly Action _onRead;
    private bool _fired;
    public EvilElement(Action onRead) => this._onRead = onRead;
    public event EventHandler? ValueChanged;
    public AggregateType AggregateType => AggregateType.AddRaw;
    public float Value
    {
        get
        {
            if (!this._fired) { this._fired = true; this._onRead(); }
            return 0;
        }
    }
}

var attribute = new ComposableAttribute(def);
var buff = new SimpleElement { Value = 100, AggregateType = AggregateType.AddRaw };
attribute.AddElement(buff);
attribute.AddElement(new EvilElement(() => attribute.RemoveElement(buff)));

// buff has been removed, but:
//   value during race:      100
//   value on next read:     100
//   value after another:    100
```

For #883 specifically this means an offline bot polling `MaximumHealth` every 500 ms can sit on an inflated `MaximumHealth` indefinitely after a buff expires.

To be clear: **this is pre-existing, not a regression from this PR** — the `_cachedValue = newValue` lost update is structurally identical on `master`. (On `master` my harness throws `Collection was modified` before it can even demonstrate the staleness, which is its own confirmation that the PR fixes the crash.) But the description currently justifies leaving it on the grounds that it self-heals, and it doesn't.

The good news is that with the lock now present, closing it is cheap — a generation counter, no extra contention, no lock held across the recursive walk:

```csharp
private long _version;   // incremented in ElementChanged

private float GetAndCacheValue()
{
    IElement[] elements;
    long versionAtStart;
    lock (this._elementLock)
    {
        if (this._elementList.Count == 0) { this._cachedValue = 0; return 0; }
        versionAtStart = this._version;
        elements = [.. this._elementList];
    }

    // ... aggregate outside the lock, unchanged ...

    lock (this._elementLock)
    {
        if (this._version == versionAtStart)
        {
            this._cachedValue = newValue;
        }
    }

    return newValue;   // caller still gets its own freshly computed value
}
```

The caller keeps getting a coherent-for-its-snapshot value; the cache just never gets poisoned. The same applies to `AttributeRelationshipElement.GetAndCacheValue`. Either do this, or reword the section to say the stale value persists until the next element change — but please don't leave the current wording, it will mislead whoever picks this up next.

### 2. Same root cause, one level up: `AttributeSystem._attributes` is an unsynchronized `Dictionary` on the very same call path

`AttributeSystem._attributes` is a plain `Dictionary<AttributeDefinition, IAttribute>` (`AttributeSystem.cs:14`). The timer thread reaches it through the exact path this PR is fixing:

`OnEffectTimeOutAsync` → `AttributeSystem.RemoveElement` → `this._attributes.Remove(targetAttribute)` (`AttributeSystem.cs:158`, taken whenever the expiring buff was the attribute's last element)

…while the game-logic thread is doing `TryGetValue` (`:131`, `:151`, `:252`) and `Add` (`:134`, `:211`) on the same dictionary. Concurrent mutation of a `Dictionary` during reads is undefined: wrong results, exceptions, or an infinite spin inside `FindEntry` if a resizing `Add` races a lookup. That's the same class of bug as the one being fixed, on the same trigger, and the failure mode is worse than the list race.

There's also a TOCTOU in the same method — `RemoveElement` → `if (!composableAttribute.Elements.Any())` → `_attributes.Remove` isn't atomic, so an element added in between is silently dropped along with the attribute, and the next `GetOrCreateAttribute` hands back a fresh empty `ComposableAttribute` with all relationship subscriptions lost.

`MonsterAttributeHolder` already guards its dictionary with `lock (this._attributesLock)`, so there's precedent in the codebase for the same treatment here.

Totally fine to leave for a follow-up — but "This change removes the **crashes**" currently reads as if the path is now safe, and it isn't. Worth an explicit sentence in the description (and ideally a follow-up issue).

### 3. Every cache miss now allocates — copy-on-write would be strictly better here

`GetAndCacheValue` allocates an `IElement[]` per recompute, and `Elements` now allocates per access. Two existing callers pay for that on every element removal: `AttributeSystem.cs:156` (`Elements.Any()` materializes the whole list to answer "is it empty") and `MonsterAttributeHolder.cs:135` (`Elements.Skip(1).Any()` materializes it to answer "count > 1").

Given reads massively outnumber mutations here (buffs apply/expire, items get equipped — versus attribute reads in every damage calc and every bot tick), copy-on-write fits better and takes the lock off the read path entirely:

```csharp
private volatile IElement[] _elements = [];

public IEnumerable<IElement> Elements => this._elements;   // no lock, no allocation

// AddElement:    lock (this._elementLock) { this._elements = [.. this._elements, element]; }
// RemoveElement: lock (this._elementLock) { /* copy without the element */ }
// GetAndCacheValue: var elements = this._elements;        // single volatile read
```

The reader becomes wait-free *and* allocation-free, only the rare mutation copies, and the "aggregation runs outside the lock on purpose" caveat in the comment stops being needed at all — there's nothing to hold. It composes fine with the version counter from item 1.

If you'd rather keep the lock, at least consider giving `IComposableAttribute` an `ElementCount` (or `bool HasElements`) so the two `Any()`/`Skip(1).Any()` callers don't have to materialize an array to count.

### 4. Minor: the `+=` / `-=` are still outside the lock

`AddElement` adds to the list under the lock but subscribes after; `RemoveElement` removes under the lock but unsubscribes after. Concurrent `Add(e)` / `Remove(e)` of the same element can interleave as: Add inserts → Remove deletes and observes `removed == true` → Remove's `-=` no-ops (nothing subscribed yet) → Add's `+=` lands. Result: a live `ValueChanged` subscription on an element that is no longer in the list, invalidating this attribute's cache forever.

Very narrow and pre-existing, but since you're introducing the lock anyway it's free to close — `+=`/`-=` don't call into user code, so they can move inside the lock while `ElementChanged(...)`/`RaiseValueChanged` correctly stay outside.

### 5. Nit: test asymmetry on the timeout

`ConcurrentRecalculationOfDependentAttributesDoesNotDeadlock` has `.WaitAsync(TimeSpan.FromSeconds(30))`; `ConcurrentAddRemoveWhileReadingValueDoesNotThrow` doesn't. If a future locking change deadlocks the single-attribute path, that test hangs the CI run indefinitely instead of failing. Same guard on both, please.

Related observation: running the two tests in a tight loop under load, I occasionally saw only one of the two fail against unpatched code. They're good detectors but not deterministic ones — inherent to timing-based race tests. The reentrancy trick from item 1 is one way to get a deterministic companion test if you want belt-and-braces coverage.

### 6. Doc note on `Elements`

Good call documenting the snapshot semantics on `IComposableAttribute`. Worth also stating that callers must not expect to observe later mutations through a previously obtained sequence. FWIW `SkinChatCommandPlugIn.cs:37` already does `Elements.ToList().ForEach(RemoveElement)` — now a redundant copy, but harmless and still correct.

---

**Summary:** the two code changes are correct and I'd take them as-is for the crash. Please fix the wording in item 1 (or implement the version counter — it's ~8 lines), and add a sentence acknowledging item 2 so the dictionary race doesn't get lost. 3–6 are optional polish.

---

nolt:
Thanks for the review I will look into it.

nolt:
All six points: version counter so a raced recalculation no longer latches a stale value (1), copy-on-write element list for a lock-free/alloc-free read path (3), a lock around AttributeSystem._attributes with TOCTOU-safe empty-attribute removal (2), plus the symmetric test timeout and a deterministic stale-cache regression test (5) and the Elements doc note (6). Left as noted follow-ups: the fast-path read of the cached float? stays unsynchronized (transient, self-correcting) and a narrow empty-attribute TOCTOU remains — both would need the lock held across the value-changed dispatch, which synchronously raises a public event. Updated the description per item 1.

sven-n:
## Re-review of `13e5eb0`

All six points are addressed, and addressed well — the copy-on-write list, the version guard and the dictionary lock are each done the way I'd hoped, and the new follow-ups section is honest about what's left. I re-verified from scratch:

| Check | Result |
|---|---|
| `dotnet build src/MUnique.OpenMU.sln` | clean — 0 `error CS`, no new analyzer warnings (only the pre-existing NU1902/NU1903 package advisories) |
| `MUnique.OpenMU.AttributeSystem.Tests` | **44/44** |
| `MUnique.OpenMU.Tests` (the big GameLogic suite) | **750/750** — the `AttributeSystem` locking refactor regresses nothing |
| New deterministic test actually detects the bug | ✅ removed *only* the version guard from `ComposableAttribute.GetAndCacheValue` → `ConcurrentRemovalDuringReadDoesNotLatchStaleValue` fails with `Expected: 0, But was: 100.0f` |
| Both threaded tests now time out instead of hanging | ✅ |

Suite time went 153 ms → 443 ms, which is the expected consequence of the version guard: under sustained mutation the reader legitimately never caches and recomputes every pass. Not a concern.

Two more things I checked while I was in there, both fine: `AttributeRelationshipElement.InputElements` is an array at all three construction sites (`AttributeSystem.cs:81`, `AttributeSystemExtensions.cs:75`, `SoulBarrierStrengSkillAction.cs:83`), so the `InputElements.Sum(...)` on the crash stack has no live-collection hazard of its own; and there is no lock-ordering cycle between `_attributesLock` and `_elementLock` — `RemoveElement` holds the former while calling `Elements`, which is now lock-free, so the two never nest.

One substantive finding below, then nits.

---

### 1. The `ReferenceEquals` guard is right, but the *identity* hazard it guards against needs no concurrency at all

The follow-up note frames the remaining problem as a narrow concurrent window:

> An empty attribute can still be dropped from `AttributeSystem._attributes` in a narrow window while another thread is adding an element to it, orphaning that contribution.

The window is narrow, but the underlying breakage isn't a race at all — dropping an emptied attribute orphans **every reference captured earlier**, deterministically, single-threaded. That matters here specifically because capturing a long-lived reference is exactly how relationship inputs are wired: `CreateRelatedAttribute` (`AttributeSystem.cs:81`) and `AttributeSystemExtensions.CreateElement` (`:75`) both do `GetOrCreateAttribute(...)` and hold the returned instance inside an `AttributeRelationshipElement` for the lifetime of the relationship.

Straight-line repro, no threads:

```csharp
var system = new AttributeSystem([], [], []);
var captured = system.GetOrCreateAttribute(InputDef);   // what CreateRelatedAttribute captures

var buff = new SimpleElement(5, AggregateType.AddRaw);
system.AddElement(buff, InputDef);
system.RemoveElement(buff, InputDef);                   // now empty -> dropped from the dictionary

system.AddElement(new SimpleElement(7, AggregateType.AddRaw), InputDef);
```
```
captured.Value with buff:      5
same instance after re-create: False     <- GetOrCreateAttribute hands out a NEW instance
system value after re-add:     7
captured.Value after re-add:   0         <- the captured reference is permanently dead
```

So any relationship whose input attribute ever goes empty silently stops tracking that input. It behaves identically on `master`, so **this is pre-existing and not a regression from this PR** — but the PR is hardening this exact method and adding a `ReferenceEquals` identity guard to it, so it seems like the right moment to name it properly.

**Suggestion: just don't drop the entry.** An empty `ComposableAttribute` evaluates to `0`, which is precisely what `GetValueOfAttribute` returns for a missing attribute — so the removal is a pure memory optimization that costs instance identity. Deleting it closes the disclosed TOCTOU follow-up *and* this one, with no reentrancy risk, and without the option you rightly rejected (holding the lock across the value-changed dispatch):

```csharp
composableAttribute.RemoveElement(element);
// nothing else needed - an emptied attribute evaluates to 0, same as a missing one,
// and keeping it preserves the identity that relationship elements captured.
```

The cost is one empty `ComposableAttribute` per definition ever touched, per system — bounded by the number of `AttributeDefinition`s (a few hundred), so tens of KB per player at worst. It also incidentally removes the reason `OnAttributeRemoved` is never called (third follow-up bullet).

If you'd rather keep the drop, then please just widen the follow-up note — "in a narrow window while another thread is adding" undersells it; the orphaning happens on a plain sequential add/remove/add.

---

### Nits

**2. `OnAttributeAdded` runs outside the lock, so the get-or-create isn't quite as atomic as the description says.** In both `AddElement` and `GetOrCreateAttribute` the new attribute is published into the dictionary *inside* the lock but `OnAttributeAdded(created)` runs after it. A second thread can therefore find the attribute and add an element to it before `ItemAwareAttributeSystem.OnAttributeAdded` (`:104-107`) has subscribed `OnAttributeValueChanged` — that one change is then never forwarded to `AttributeValueChanged` (a missed client stat update, not a crash). Moving the call inside would mean invoking a `protected virtual` under the lock, so leaving it is defensible; it just belongs in the follow-ups list, since the PR now claims the get-or-create is atomic.

**3. "It does not call into user code, so it is safe to hold the lock across it"** (`ComposableAttribute.AddElement`) — event accessors *are* user code, and `ConstantElement` in this very repo defines custom `add`/`remove` (empty bodies). The claim holds today because every `IElement` implementation uses either a field-like event (which compiles to an `Interlocked.CompareExchange` loop, no lock) or those empty accessors — but as written the reasoning would bless a future `IElement` whose accessor takes a lock, which is a lock-ordering hazard. Suggest: *"every `IElement` in the tree uses a field-like (interlocked) or empty event accessor, so this cannot acquire another lock."*

**4. `_attributesLock` doc says "Synchronizes every access to `_attributes`"** — the constructor's `this._attributes.Add(statAttribute.Definition, statAttribute)` (`AttributeSystem.cs:36`) is unguarded. It's safe (the instance isn't published yet), but the comment reads as an invariant a future edit might trust. Worth a half-sentence.

**5. Optional: the first lock in both `GetAndCacheValue`s is removable.** Reading a *stale* `_version` only ever makes the guard more conservative (skip the cache write) — never wrong — so `Volatile.Read(ref this._version)` suffices, and in `ComposableAttribute` the `(snapshot, version)` pair doesn't need to be captured atomically either: whichever of the two reads is staler, the mismatch just fails the guard. That would halve the lock traffic on relationship-element cache misses, which are frequent in damage calculation. Uncontended locks are ~20 ns, so this is genuinely optional — and the current form is more obviously correct, which is worth something.

**6. `Elements => this._elements`** hands the shared array out as `IEnumerable<IElement>`; a caller can cast it back to `IElement[]` and write through it, which would now corrupt a snapshot other threads are mid-aggregation on (previously, mutating the returned `IList` was merely wrong for the one owner). Nothing in the repo does this, and `ImmutableArray` can't be `volatile`, so there's no free fix — just noting that the XML doc says "Treated as immutable" while the type doesn't enforce it.

---

**Summary:** LGTM for the crash fix and the coherency fix — I'd merge this. Item 1 is the only one I'd like a response on, and even there the minimum is a wording change to the follow-up bullet; deleting the four lines that drop the emptied attribute would be the stronger answer. 2–6 are optional.

---

PR #888, opened 2026-08-19
Regens refactoring & add tier 2 chars default movement speeds

ze-dom:
**To-do**

- [x] Comment changes with sources
- [x] UpdatePlugins

**Developments**
- Added resting spots recovery logic
- Added default running/swimming speed attributes for tier 2 chars (MG, DL, RF)
- Simplified some attribute relationships to avoid the use of "temporary" stats

**Bugfixes**
- Character pose not being updated
- _SdRecoverySpeedInc_ & _IncreaseSdRecoveryRate_ master skills 
- Reviewed values for health, mana, shield and ability passive regeneration
- `Stats.NearbyPartyMemberCount` not being updated

**Note**: at some point more passive regenerations were added for low level chars ([MU KR](https://www-muonline-co-kr.translate.goog/news/patch-note/detail/6027?page=62&category=1701&_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_) => [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18628), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L17729)). I didn't add these (this would only affect Season6 version).


src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,20 +135,22 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
         attributeRelationships.Add(this.CreateConditionalRelationship(Stats.DefenseFinal, Stats.IsShieldEquipped, Stats.BonusDefenseWithShield, AggregateType.AddFinal));
         attributeRelationships.Add(this.CreateConditionalRelationship(Stats.DefenseRatePvm, Stats.IsShieldEquipped, Stats.BonusDefenseRateWithShield, AggregateType.AddFinal));
 
-        var tempInnovDefDec = this.Context.CreateNew<AttributeDefinition>(Guid.NewGuid(), "Temp Innovation defense decrement", string.Empty);
-        this.GameConfiguration.Attributes.Add(tempInnovDefDec);
-        attributeRelationships.Add(this.CreateAttributeRelationship(tempInnovDefDec, -1, Stats.InnovationDefDecrement));
-        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.DefenseDecrement, 1, tempInnovDefDec, InputOperator.Add, AggregateType.Multiplicate));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.DefenseDecrement, -1, Stats.InnovationDefDecrement));
+
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.HealthRecoveryMultiplier, 0.03f, Stats.IsRecovering));
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18999), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18089)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,20 +135,22 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
...
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ManaRecoveryMultiplier, 0.03f, Stats.IsRecovering));
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18872), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L17967)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,20 +135,21 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
         attributeRelationships.Add(this.CreateConditionalRelationship(Stats.DefenseFinal, Stats.IsShieldEquipped, Stats.BonusDefenseWithShield, AggregateType.AddFinal));
         attributeRelationships.Add(this.CreateConditionalRelationship(Stats.DefenseRatePvm, Stats.IsShieldEquipped, Stats.BonusDefenseRateWithShield, AggregateType.AddFinal));
 
-        var tempInnovDefDec = this.Context.CreateNew<AttributeDefinition>(Guid.NewGuid(), "Temp Innovation defense decrement", string.Empty);
-        this.GameConfiguration.Attributes.Add(tempInnovDefDec);
-        attributeRelationships.Add(this.CreateAttributeRelationship(tempInnovDefDec, -1, Stats.InnovationDefDecrement));
-        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.DefenseDecrement, 1, tempInnovDefDec, InputOperator.Add, AggregateType.Multiplicate));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.DefenseDecrement, -1, Stats.InnovationDefDecrement));
+
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.HealthRecoveryMultiplier, 0.03f, Stats.IsResting));
```
ze-dom, line 140: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18999), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18089)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,20 +135,21 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
...
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ManaRecoveryMultiplier, 0.03f, Stats.IsResting));
```
ze-dom, line 141: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L19063), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18149)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,20 +135,21 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
...
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.AbilityRecoveryAbsolute, 3, Stats.IsInSafezone));
```
ze-dom, line 142: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18924), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18019)

src/GameServer/MessageHandler/AnimationHandlerPlugIn.cs
```diff
@@ -47,6 +48,11 @@ public async ValueTask HandlePacketAsync(Player player, Memory<byte> packet)
             _ => default,
         };
 
+        if (player.Pose > CharacterPose.Standing)
+        {
+            player.Attributes?.SetStatAttribute(Stats.IsResting, 1.0f);
+        }
+
```
ze-dom, line 55: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/protocol.cpp#L10892), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/protocol.cpp#L11907)

A great game design detail. Resting spots can be found both within (sitting, leaning, hanging) towns as well as outside (sitting, leaning; in lower level maps - where they may be handy). Think hardcore, first versions, old school MU, where you are scrapping by at starting levels 😅 

The reason `Stats.IsResting` is set only when `player.Pose > CharacterPose.Standing` and reset in _UpdateInSafezoneAfterPlayerMoved.cs_ is because when hanging, the player can still rotate by moving the cursor, and in that case it will send a `Character.Standing` pose despite still hanging (maybe "floating" would be a better term 😅).

src/GameLogic/GameContext.cs
```diff
@@ -82,6 +87,8 @@ public GameContext(GameConfiguration configuration, IPersistenceContextProvider
             this.ItemPowerUpFactory = new ItemPowerUpFactory(loggerFactory.CreateLogger<ItemPowerUpFactory>());
             this.PartyManager = new PartyManager(configuration.MaximumPartySize, loggerFactory.CreateLogger<Party>());
             this._recoverTimer = new Timer(this.RecoverTimerElapsed, null, this.Configuration.RecoveryInterval, this.Configuration.RecoveryInterval);
+            this._restingRecoveryInterval = (int)Math.Round(this.Configuration.RecoveryInterval * (1 + (2f / 3)));  // Originally, resting recovery interval is 5s
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L19003), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18092)
3000 * 5f / 3 = 5000

This way it's always proportional to `Configuration.RecoveryInterval`.

sven-n:
Nice work on the regen rework — the behavior it produces looks right to me. My concern is with *where* the numbers live: `Player.RegenerateAsync` now carries a lot of game design that a server owner can't touch without recompiling.

### What's hardcoded

1. **A mode flag in a public API.** `RegenerateAsync(bool isRestingCycle)` — the flag then filters `Stats.IntervalRegenerationAttributes` down to health + mana via reference comparison against static `Stats` members.
2. **Compensation multipliers as literals.** `3f/7f` for health, `1` for mana/ability, `* 3` for shield. They encode "originally every X seconds" relative to `GameConfiguration.RecoveryInterval`, which is admin-configurable — set it to 1000 ms and every one of those constants is silently wrong.
3. **The shield ramp table.** `>= 25 → 3`, `>= 15 → 2.5`, `>= 10 → 2`, else skip, as a `switch` inside the generic loop.
4. **The gate conditions.** The safezone / `ShieldRecoveryEverywhere` check and the `CurrentHealth && !isRestingCycle` check select behavior by comparing against `Stats.CurrentShield` / `Stats.CurrentHealth`.
5. **A second `Timer` in `GameContext`** at `RecoveryInterval * (1 + 2/3)`, not configurable, sharing a callback through a boxed `bool` in the timer state.
6. **Split ownership of `IsResting`.** `AnimationHandlerPlugIn` sets it to 1 when `Pose > Standing` but never back to 0 — a subsequent animation packet resets `player.Pose` to `Standing` through the `_ => default` arm while `IsResting` stays 1, so only movement clears it. And clearing it inside `UpdateIsInSafezoneAfterPlayerMoved` gives that plug-in two unrelated jobs, so it can't be deactivated independently.

Also worth deciding explicitly: since the resting cycle keeps mana at multiplier 1, mana regenerates on *both* cycles (~1.53x while sitting). That may well be intended, but right now it's neither configurable nor obvious from the code.

### Suggestion: rate-based regeneration + everything else as attributes

**1. Make regeneration rate-based instead of tick-based.** Give each `Stats.Regeneration` its own interval and apply the elapsed fraction:

```csharp
public class Regeneration
{
    // ...
    public TimeSpan Interval { get; init; } = TimeSpan.FromSeconds(3);
    public AttributeDefinition? EnabledAttribute { get; init; }
}

private static Regeneration HealthRegeneration { get; } =
    new(HealthRecoveryMultiplier, MaximumHealth, CurrentHealth, HealthRecoveryAbsolute)
    { Interval = TimeSpan.FromSeconds(7) };
```

```csharp
var factor = (float)((now - this._lastRegeneration[r]) / r.Interval);
attributes[r.CurrentAttribute] = Math.Min(
    attributes[r.CurrentAttribute] + (((max * multiplier) + absolute) * factor),
    max);
```

`3f/7f` then *falls out of* `3s tick / 7s interval` instead of being written down, and stays correct for any `RecoveryInterval`. This removes items 2 and 5 — one timer, no `bool` parameter, `RegenerateAsync()` keeps its original signature. (If discrete steps feel better for the client, the same timestamp supports "apply only when `elapsed >= Interval`".)

**2. Gate conditions as attributes instead of branches.** With the optional `EnabledAttribute` above, the shield special case becomes configuration:

```csharp
// ShieldRecoveryActive = Maximum(IsInSafezone, ShieldRecoveryEverywhere)
this.CreateAttributeRelationship(Stats.ShieldRecoveryActive, 1, Stats.IsInSafezone, InputOperator.Maximum, Stats.ShieldRecoveryEverywhere)
```

and the loop body is just `if (r.EnabledAttribute is { } flag && attributes[flag] < 1) continue;`.

**3. Replace the hard steps with a linear ramp driven by a duration attribute.** The attribute system already gives us the cap for free — `AttributeDefinition.MaximumValue` is enforced in `ComposableAttribute.CalculateValue()` and in `StatAttribute` — and `CharacterClassInitialization.cs:143-144` already shows the `temp attribute + Multiplicate` pattern. Your `ConstValueAttribute.AggregateType` change is what lets the `1 +` base term be expressed without a temp stat, so it earns its keep here.

New input attributes, written by code: `Stats.RestingDuration` (seconds since sitting/leaning/hanging) and `Stats.ShieldRecoveryDuration` (seconds of uninterrupted recovery). Everything else is data:

```csharp
// Stats.cs — the cap lives on the definition
public static AttributeDefinition ShieldRecoveryRampFactor { get; } =
    new(new Guid("..."), "Shield recovery ramp factor", "Rises linearly with the uninterrupted shield recovery duration.")
    {
        MaximumValue = 3,
    };
```

```csharp
// CharacterClassInitialization — factor = 1 + (duration / 15), capped at 3
baseAttributeValues.Add(this.CreateConstValueAttribute(1, Stats.ShieldRecoveryRampFactor));
attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ShieldRecoveryRampFactor, 1f / 15f, Stats.ShieldRecoveryDuration));
attributeRelationships.Add(this.CreateAttributeRelationship(
    Stats.ShieldRecoveryMultiplier, 1, Stats.ShieldRecoveryRampFactor, InputOperator.Multiply, AggregateType.Multiplicate));
```

The resting bonus takes the same shape (and there's already precedent one line above it, `HealthRecoveryMultiplier += 0.01 * IsInSafezone`):

```csharp
attributeRelationships.Add(this.CreateAttributeRelationship(Stats.RestingRecoveryFactor, 1f / 30f, Stats.RestingDuration));
attributeRelationships.Add(this.CreateAttributeRelationship(
    Stats.HealthRecoveryMultiplier, 1, Stats.RestingRecoveryFactor, InputOperator.Multiply, AggregateType.Multiplicate));
```

That deletes the `switch`, the `_lastShieldRegenerateStop` field, the resting filter and the second timer — and since slope and cap are per-character-class relationships, they're tunable in the admin panel and can differ per class.

Mapping the current numbers: either keep the delay (const base `2`, slope `1/15`, cap `3`, with a configurable `Stats.ShieldRecoveryDelay` before the duration starts counting), or drop it (slope `3/25`, cap `3`, counting from 0 — recovery starts immediately but weak, full ramp at 25s). I'd default to the second and keep the delay attribute at hand for servers that want the original hard cut-off.

**4. One small plug-in keeps the durations current** — no game rules inside it:

```csharp
[PlugIn]
[Guid("...")]
public class RegenerationDurationPlugIn : IPeriodicTaskPlugIn, IAttackableMovedPlugIn, IAttackableGotHitPlugIn
```

tick: `duration += elapsed` (the existing 1s `_tasksTimer`, no third timer); reset `RestingDuration` on pose → `Standing` and on movement; reset `ShieldRecoveryDuration` on hit, on leaving the safezone without `ShieldRecoveryEverywhere`, and when the shield is full.

One caveat: I wouldn't make the durations persisted stat attributes the way `IsInSafezone` is. Those are materialized as EF-tracked entities (`Player.AddMissingStatAttributes` → `PersistenceContext.CreateNew<StatAttribute>`), so a value changing every second dirties the character on every tick and gets written out by `PeriodicSaveProgressPlugIn`. Adding a non-persistent `SimpleElement` to the composable attribute when the player enters the world (the way power-ups attach) avoids that entirely. The remaining churn is cheap: each change raises `AttributeValueChanged` → `UpdateStatsBasePlugIn.UpdateStatsAsync`, which is a `FrozenDictionary` miss for an unknown definition, so no packets; dependent composables invalidate lazily and stop changing once the duration saturates.

**5. Give `IsResting` one owner** — the `Player.Pose` setter (or a dedicated `IPlayerPoseChangedPlugIn` if it should stay overridable):

```csharp
character.Pose = value;
this.Attributes?.SetStatAttribute(Stats.IsResting, value > CharacterPose.Standing ? 1f : 0f);
```

which fixes the never-cleared case and lets `UpdateIsInSafezoneAfterPlayerMoved` go back to doing one thing. With `RestingDuration` in place, `IsResting` could even be derived (`Minimum(RestingDuration, 1)` with `MaximumValue = 1`), though keeping it explicit as the gate flag is clearer.

### Net effect

`RegenerateAsync()` goes back to one signature and one loop with no `Stats.X ==` comparisons:

```csharp
foreach (var r in Stats.IntervalRegenerationAttributes)
{
    if (r.EnabledAttribute is { } flag && attributes[flag] < 1) { continue; }
    // current += ((max * multiplier) + absolute) * elapsedFactor(r)
}
```

Everything the PR currently hardcodes lands in one of two configurable places: the `Regeneration` definitions (interval, gate attribute) and attribute relationships seeded by `Persistence.Initialization` (ramp slopes, caps, resting/safezone bonuses — per character class). Defaults can reproduce the current numbers closely enough that this stays a refactor of the same behavior, and it fits the `UpdatePlugIn` you already have on the to-do list: it would seed the new attributes and relationships for existing databases.

As a bonus, once the durations are real attributes other systems can consume them — an item option that raises the ramp cap, or a skill that scales with resting time — all without touching `Player`.

---

src/GameLogic/GameContext.cs
```diff
@@ -82,6 +87,8 @@ public GameContext(GameConfiguration configuration, IPersistenceContextProvider
             this.ItemPowerUpFactory = new ItemPowerUpFactory(loggerFactory.CreateLogger<ItemPowerUpFactory>());
             this.PartyManager = new PartyManager(configuration.MaximumPartySize, loggerFactory.CreateLogger<Party>());
             this._recoverTimer = new Timer(this.RecoverTimerElapsed, null, this.Configuration.RecoveryInterval, this.Configuration.RecoveryInterval);
+            this._restingRecoveryInterval = (int)Math.Round(this.Configuration.RecoveryInterval * 5f / 3);  // Originally, resting recovery interval is 5s
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L19003), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18092)

3000 * 5f / 3 = 5000
This way it's always proportional to `Configuration.RecoveryInterval`.

src/GameLogic/Player.cs
```diff
@@ -1045,31 +1045,82 @@ public async ValueTask ClientReadyAfterMapChangeAsync()
...
+                            break;
+                        case >= 15:
+                            multiplier = 2.5f;
+                            break;
+                        case >= 10:
+                            multiplier = 2;
+                            break;
+                        default:
+                            continue;
+                    }
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L25856), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L26144)

src/GameLogic/Player.cs
```diff
@@ -1045,31 +1045,82 @@ public async ValueTask ClientReadyAfterMapChangeAsync()
...
+                else if (r.CurrentAttribute == Stats.CurrentHealth && !isRestingCycle)
+                {
+                    if (attributes[Stats.IsResting] > 0)
+                    {
+                        // We only regen health on the quicker resting cycle (for mana both cycles apply).
+                        continue;
+                    }
+
+                    // Originally, health is recovered every 7s
+                    multiplier = 3f / 7f;
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L19146), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18234)

src/GameLogic/Player.cs
```diff
@@ -1045,31 +1045,82 @@ public async ValueTask ClientReadyAfterMapChangeAsync()
...
+                }
+                else
+                {
+                    // Originally, mana and ability are recovered every 3s (the default recovery interval).
+                    multiplier = 1;
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18862), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L17957)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -184,7 +177,8 @@ private void AddCommonBaseAttributeValues(ICollection<ConstValueAttribute> baseA
 
         if (!this.UseClassicPvp)
         {
-            baseAttributeValues.Add(this.CreateConstValueAttribute(0.01f, Stats.ShieldRecoveryMultiplier));
+            baseAttributeValues.Add(this.CreateConstValueAttribute(100, Stats.ShieldRecoveryMultiplier));
+            baseAttributeValues.Add(this.CreateConstValueAttribute(1f / 75000, Stats.ShieldRecoveryMultiplier, AggregateType.Multiplicate)); // 1 / (30 * 100 * 25)
```
ze-dom, line 188: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L25851), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L26137)

src/Persistence/Initialization/CharacterClasses/ClassDarkLord.cs
```diff
@@ -128,8 +122,10 @@ private CharacterClass CreateDarkLord(CharacterClassNumber number, string name,
         result.BaseAttributeValues.Add(this.CreateConstValueAttribute(38, Stats.MaximumMana));
         result.BaseAttributeValues.Add(this.CreateConstValueAttribute(48.5f, Stats.MaximumHealth));
         result.BaseAttributeValues.Add(this.CreateConstValueAttribute(2, Stats.SkillMultiplier));
-        result.BaseAttributeValues.Add(this.CreateConstValueAttribute(1.0f / 33f, Stats.AbilityRecoveryMultiplier));
+        result.BaseAttributeValues.Add(this.CreateConstValueAttribute(0.03f, Stats.AbilityRecoveryMultiplier));
```
ze-dom, line 126: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18943), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18038)

Yes, 1/33 is pretty much the same, but it's more obvious at a glance to have the exact number as the sources imo.
Note that DK is slightly higher at 5% :) (already correct)

src/Persistence/Initialization/VersionSeasonSix/Items/GuardianOptions.cs
```diff
@@ -79,7 +79,7 @@ private void CreateHelmOption()
         definition.AddsRandomly = false;
 
         definition.PossibleOptions.Add(this.CreateOption(ItemGroups.Helm, Stats.DefenseRatePvp, 10, AggregateType.AddRaw, ItemOptionDefinitionNumbers.GuardianOption1));
-        definition.PossibleOptions.Add(this.CreateOption(ItemGroups.Helm, Stats.ShieldRecoveryMultiplier, 20, AggregateType.AddRaw, ItemOptionDefinitionNumbers.GuardianOption2)); // 20 absolute, need test
+        definition.PossibleOptions.Add(this.CreateOption(ItemGroups.Helm, Stats.ShieldRecoveryMultiplier, 20, AggregateType.AddRaw, ItemOptionDefinitionNumbers.GuardianOption2));
```
ze-dom, line 82: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L25847), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L26133)

src/Persistence/Initialization/VersionSeasonSix/SkillsInitializer.cs
```diff
@@ -769,7 +769,7 @@ private void InitializeMasterSkillData()
         this.AddPassiveMasterSkillDefinition(SkillNumber.AutomaticManaRecInc, Stats.ManaRecoveryMultiplier, AggregateType.AddRaw, FormulaRecoveryIncrease181, Formula181, 2, 1, SkillNumber.Undefined, SkillNumber.Undefined, 20);
         this.AddPassiveMasterSkillDefinition(SkillNumber.PoisonResistanceInc, Stats.PoisonResistance, AggregateType.AddRaw, Formula120, Formula120, 2, 1);
         this.AddPassiveMasterSkillDefinition(SkillNumber.DurabilityReduction2, Stats.ItemDurationIncrease, AggregateType.Multiplicate, Formula1204, 3, 1, SkillNumber.DurabilityReduction1);
-        this.AddMasterSkillDefinition(SkillNumber.SdRecoverySpeedInc, SkillNumber.MaximumSDincrease, SkillNumber.Undefined, 1, 3, SkillNumber.Undefined, 20, Formula120);
+        this.AddPassiveMasterSkillDefinition(SkillNumber.SdRecoverySpeedInc, Stats.ShieldRecoveryMultiplier, AggregateType.Multiplicate, $"1 + {FormulaRecoveryIncrease120}", Formula120, 3, 1, SkillNumber.MaximumSDincrease, SkillNumber.Undefined, 20);
```
ze-dom, line 772: _user.cpp_: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L25882), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L26166)
_MasterLevelSkillTreeSystem.cpp_: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L399), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L387)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,29 +135,25 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
...
-        }
 
         attributeRelationships.Add(this.CreateAttributeRelationship(Stats.MaximumGuildSize, 0.1f, Stats.Level));
         attributeRelationships.Add(this.CreateAttributeRelationship(Stats.CanFly, 1, Stats.IsDinorantEquipped));
     }
 
     private void AddCommonBaseAttributeValues(ICollection<ConstValueAttribute> baseAttributeValues, bool isMaster)
     {
-        baseAttributeValues.Add(this.CreateConstValueAttribute(1.0f / 27.5f, Stats.ManaRecoveryMultiplier));
+        baseAttributeValues.Add(this.CreateConstValueAttribute(0.037f, Stats.ManaRecoveryMultiplier));
```
ze-dom, line 163: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18872), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L17967)

Yes, practically the same, but easier to crosscheck with sources imo. Up to you.

src/GameLogic/Player.cs
```diff
@@ -1824,7 +1854,6 @@ private async ValueTask OnPlayerEnteredWorldAsync()
         this.AddMissingStatAttributes();
 
         this.Attributes = new ItemAwareAttributeSystem(this.Account!, selectedCharacter, this.GameContext.Configuration);
-        this.Attributes[Stats.NearbyPartyMemberCount] = 0;
```
ze-dom: Bug. This was not a `StatAttribute`, so the setter didn't work.

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,29 +135,29 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
...
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.DefenseDecrement, -1, Stats.InnovationDefDecrement));
+
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.HealthRecoveryMultiplier, 0.03f, Stats.IsResting));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ManaRecoveryMultiplier, 0.03f, Stats.IsResting));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.AbilityRecoveryAbsolute, 3, Stats.IsInSafezone));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.IsShieldRecoveryActive, 1, Stats.IsInSafezone));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.IsShieldRecoveryActive, 1, Stats.ShieldRecoveryEverywhere));
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ShieldRecoveryRampFactor, 1f / 15f, Stats.ShieldRecoveryHiatus));  // (3 - 2) / (25 - 10)
+        attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ShieldRecoveryMultiplier, 1, Stats.ShieldRecoveryRampFactor, aggregateType: AggregateType.Multiplicate));
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L25856), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L26144)

src/GameLogic/Attributes/Stats.cs
```diff
@@ -1647,5 +1701,36 @@ public Regeneration(AttributeDefinition regenerationMultiplier, AttributeDefinit
...
+        /// To avoid having different timers each with a regeneration interval, we run just one set to <see cref="GameConfiguration.RecoveryInterval"/>.
+        /// Then, we apply a compensation factor which is the ratio between elapsed time and <see cref="Interval"/>.
+        /// This way we can mimic original regeneration over time while still keeping it configurable.
+        /// </remarks>
+        public TimeSpan Interval { get; init; } = TimeSpan.FromSeconds(3);
+
+        /// <summary>
+        /// Gets the interval at which the attribute would complete a full regeneration cycle when the player is resting (<see cref="IsResting"/>).
+        /// </summary>
+        public TimeSpan IntervalResting { get; init; } = TimeSpan.FromSeconds(3);
```
ze-dom, line 1724: Default for `Interval` and `IntervalResting` as 3 for mana and ability: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L18862), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L17957)

src/GameLogic/Attributes/Stats.cs
```diff
@@ -1588,13 +1617,38 @@ public static IEnumerable<Regeneration> AfterMonsterKillRegenerationAttributes
...
+    };
 
-    private static Regeneration HealthRegeneration { get; } = new(HealthRecoveryMultiplier, MaximumHealth, CurrentHealth, HealthRecoveryAbsolute);
+    /// <summary>
+    /// Gets the health regeneration.
+    /// </summary>
+    public static Regeneration HealthRegeneration { get; } = new(HealthRecoveryMultiplier, MaximumHealth, CurrentHealth, HealthRecoveryAbsolute)
+    {
+        Interval = TimeSpan.FromSeconds(7),
+        IntervalResting = TimeSpan.FromSeconds(5),
```
ze-dom, line 1634: Interval: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L19146), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18234)
IntervalResting (includes mana too): [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L19003), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L18092)

src/Persistence/Initialization/CharacterClasses/CharacterClassInitialization.cs
```diff
@@ -138,20 +135,23 @@ private void AddCommonAttributeRelationships(ICollection<AttributeRelationship>
...
             attributeRelationships.Add(this.CreateAttributeRelationship(Stats.DefenseRatePvp, 1, Stats.DefenseRatePvm));
             attributeRelationships.Add(this.CreateAttributeRelationship(Stats.AttackRatePvp, 1, Stats.AttackRatePvm));
         }
         else
         {
-            attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ShieldRecoveryMultiplier, 0.01f, Stats.IsInSafezone));
+            attributeRelationships.Add(this.CreateAttributeRelationship(Stats.IsShieldRecoveryActive, 1, Stats.IsInSafezone));
+            attributeRelationships.Add(this.CreateAttributeRelationship(Stats.IsShieldRecoveryActive, 1, Stats.ShieldRecoveryEverywhere));
+            attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ShieldRecoveryRampFactor, 1f / 15f, Stats.ShieldRecoveryHiatus));  // (3 - 2) / (25 - 10)
+            attributeRelationships.Add(this.CreateAttributeRelationship(Stats.ShieldRecoveryMultiplier, 1, Stats.ShieldRecoveryRampFactor, aggregateType: AggregateType.Multiplicate));
```
ze-dom, line 154: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L25856), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L26144)

src/GameLogic/Player.cs
```diff
@@ -1824,7 +1855,7 @@ private async ValueTask OnPlayerEnteredWorldAsync()
...
```
ze-dom, line 1597: This was not a `StatAttribute`, so any assignment didn't work.

src/GameServer/MessageHandler/AnimationHandlerPlugIn.cs
```diff
@@ -41,12 +42,17 @@ public async ValueTask HandlePacketAsync(Player player, Memory<byte> packet)
 
         player.Pose = animation switch
         {
-            0x80 => CharacterPose.Sitting,
-            0x81 => CharacterPose.Leaning,
-            0x82 => CharacterPose.Hanging,
+            0x80 or 0x6C => CharacterPose.Sitting,
+            0x81 or 0x6D => CharacterPose.Leaning,
+            0x82 or 0x6E => CharacterPose.Hanging,
```
ze-dom, line 47: For 0.75 and 0.95d

ze-dom:
All set!

sven-n:
Reviewed the current head (`4eeb7314`). The design landed well — `Interval`/`IntervalResting` per regeneration is a better generalization than what I proposed, and the ramp math (`4/3 + hiatus/15`, capped at 3) reproduces the original 2x→3x window between 10s and 25s exactly. Solution builds and the tests pass (773 in `MUnique.OpenMU.Tests`, 12 in `Persistence.Initialization.Tests`).

Seven things I'd raise, roughly in severity order.

### 1. The hiatus attribute grows by one element per second and is never bounded

`ShieldRecoveryHiatusPlugIn.cs:31`

```csharp
attributes.AddElement(new SimpleElement(1, AggregateType.AddRaw), Stats.ShieldRecoveryHiatus);
```

Every tick adds a *new element* instead of updating one. `ComposableAttribute.AddElement` does `_elements = [.. _elements, element]`, so each add is O(n) and the session total is O(n²); each add also invalidates the cache and raises `ValueChanged`, and the regeneration loop re-reads the value every recovery tick, so the aggregation (five LINQ passes over the array) re-runs each second over an ever-growing list.

Resets only happen on movement or on being hit, so an idle character in a safezone with a full shield — MU Helper, AFK, the most common case — never resets: ~3,600 elements after an hour, ~29,000 after eight. And `RemoveAllElements()` removes one at a time (`Array.IndexOf` + two array copies + an event cascade each), so the eventual reset is O(n²) *on the game logic thread*, since `AttackableMoved` is called synchronously from the `Player.Position` setter.

A single `SimpleElement` per player whose `Value` is set (`element.Value = elapsedSeconds`, reset to `0`) keeps the same semantics at constant cost, and also fixes the tick-counting drift mentioned below.

### 2. Upgraded databases lose shield recovery entirely

`RegenerationsRefactorPlugInSeason6.cs`

The initialization creates two relationships feeding the new gate:

```csharp
attributeRelationships.Add(this.CreateAttributeRelationship(Stats.IsShieldRecoveryActive, 1, Stats.IsInSafezone));
attributeRelationships.Add(this.CreateAttributeRelationship(Stats.IsShieldRecoveryActive, 1, Stats.ShieldRecoveryEverywhere));
```

The update plug-in only creates the `ShieldRecoveryEverywhere` one. On an upgraded database `IsShieldRecoveryActive` stays 0 for every character without the level 380 armor Guardian option, so the `EnablerAttribute` check blocks shield regeneration permanently. The old `IsInSafezone → ShieldRecoveryMultiplier` (0.01) combination also isn't removed, so it stays behind as dead configuration (and nudges the base from 100 to 100.01).

### 3. `IsResting` can stay set while the character stands

`AnimationHandlerPlugIn.cs:51-54`

The flag is set when the pose is above `Standing`, but nothing clears it when a later animation packet falls into the `_ => default` arm and resets the pose — only movement clears it. So: sit (0x80), then send any other animation, and the character keeps the resting recovery (health on the 5s interval, mana on both cycles, `+0.03` on both multipliers) while standing and attacking, for as long as it doesn't move.

Setting the flag in the `Player.Pose` setter gives it a single owner and covers both directions in one line; the minimal alternative is an `else` branch setting `0.0f`.

### 4. The `NearbyPartyMemberCount` login reset was dropped

`Player.cs:1628`

Master had `this.Attributes[Stats.NearbyPartyMemberCount] = 0;` right after the attribute system is created; it's now replaced by the `IsResting` reset rather than joined by it. That matters more than it used to, because this PR makes `NearbyPartyMemberCount` a real per-class stat attribute, so it is now persisted where the write used to be a silent no-op. A character that logs out with party members nearby keeps the stale count on the next login and gets an inflated Swell Life bonus until a party event corrects it. Both lines should stay.

### 5. The 0.95d update plug-in reports the Season 6 version

`RegenerationsRefactorPlugIn095d.cs:28`

```csharp
public override UpdateVersion Version => UpdateVersion.RegenerationsRefactorSeason6;
```

Should be `RegenerationsRefactor095d`. `UpdatePlugInBase.Key` is `(int)Version` and the value is written into the `ConfigurationUpdate` entry, so 0.95d installations record 107 and enum value 106 stays unreachable.

### 6. Monster hits don't interrupt the shield hiatus

`ShieldRecoveryHiatusPlugIn.cs:48-49`

```csharp
var attackerPlayer = attacker as Player ?? (attacker as Monster)?.SummonedBy;
if (defender is null || attackerPlayer is null || defender == attackerPlayer || defender.Attributes is not { } attributes)
```

The attacker-must-be-a-player check makes sense in `SelfDefensePlugIn`, where it came from, but here it means PvM shield damage never resets the ramp — precisely the `ShieldRecoveryEverywhere` case, which is the one where recovery runs outside a safezone. The `SummonedBy` fallback on the defender side also resets the *owner's* hiatus when their summoned monster is hit. `if (attackable is not Player player) return;` covers what this plug-in needs.

### 7. `_lastRegeneration` is pre-seeded and never re-seeded

`Player.cs:69`

The dictionary is initialized with exactly the four known regenerations and the loop indexes it directly (`this._lastRegeneration[r]`), so any future entry in `IntervalRegenerationAttributes` — version-specific or from a plug-in — throws `KeyNotFoundException` inside the recovery tick. `TryGetValue` keeps that safe.

Separately, the timestamps are seeded at *Player construction* rather than at character selection, so a player who spends a few minutes on the character selection screen gets one large burst factor on the first tick after entering the world. Re-seeding them when the attribute system is created would avoid it.

### Smaller things

- The resting mana rule is back to `r.CurrentAttribute == Stats.CurrentMana` inside the loop (`Player.cs:906`). A `bool RestingAppliesAdditionally { get; init; }` on `Regeneration` would keep it declarative like the rest.
- The hiatus counts ticks, not elapsed time, so it drifts whenever the periodic task runs late; using elapsed seconds also makes the threshold and the slope honest.
- The update plug-ins create attribute definitions by hand instead of using the existing `UpdatePlugInBase.AddStatIfNotExists`, which already handles the duplicate check and copies `MaximumValue`.
- `RegenerationsRefactorPlugIn075.ApplyAsync` only calls `base.ApplyAsync` — it can be removed.
- In `RegenerationsRefactorPlugInBase`, `tempAttackSpeeds` and `tempInnovationDefDecrements` are deferred queries over `gameConfiguration.Attributes`, which is mutated inside the same loop. It works as written, but a `.ToList()` would remove the hazard.
- Typo in `Stats.cs`: "against wich `HiatusAttribute` is checked".

I'd treat 1-3 as blockers; 4-6 are small, contained fixes. Happy to push any of them if that's useful.

---

ze-dom:
Ok, still some stuff to address :)

ze-dom:
From Claude's latest list:
1, 2, 4, 5 - fixed
3 - ignored, explained in the comments
6 - ignored, monsters don't damage shield
7 - now using `_lastRegenerate` instead, which already exists and serves the same purposed; added reseeding on entering world

All set!

PR #902, opened 2026-08-25
API keys for the public API, managed in the admin panel

sven-n:
## Why

The public API under `/api` — `status`, `is-online/{account}` and `send/{server}` — exists for external applications: a game launcher, a status page, a website. Since the admin panel authenticates its users, `ConfigureAdminPanel` puts every controller behind `RequireAuthorization()`, so the only way into that API is the panel's authentication cookie.

An external application can't get that cookie. It would have to go through the login form and, where it's set up, a TOTP second factor — which a server-to-server client can't do, and shouldn't be able to do with a secret it holds.

## What

A second authentication scheme next to the cookie, and a page to manage its keys.

```http
GET /api/status HTTP/1.1
X-Api-Key: <the key>
```

`Authorization: Bearer <the key>` works too, for clients which only speak that.

### The page

**API keys**, next to *Users*, requiring the administrator role. Create a key for an application, pick its role, and copy it — it's shown exactly once, because only its SHA-256 hash is stored, the same way recovery codes are.

The list shows each key by its name and its leading characters, so keys can be told apart without knowing them, plus the last time each was used — a key whose *Last used* stays empty is one you can delete. A key can be **disabled** rather than deleted, which stops it immediately without touching the application's configuration, and its role can be changed without handing out a new key.

### Roles

A key carries the same roles as a user and defaults to the least privileged one:

| Endpoint | Needs |
|---|---|
| `GET /api/status` | Viewer |
| `GET /api/is-online/{account}` | Viewer |
| `GET /api/send/{server}?msg=` | Operator |

`send` is the only endpoint which *does* something rather than reporting something, so it asks for the operator role. That's a tightening compared to before, where any authenticated panel user could use it.

A signed in panel user can still use the whole API with its own roles, which is handy while trying things out in the browser.

### Configured keys still work

The panel needs a database to store keys in, and the API has to work before that database exists — the same reason the panel has a bootstrap user. So `AdminPanel:Api:Keys` and `OPENMU_API_KEY` / `OPENMU_API_KEY_ROLES` are still read, and the compose files pass the variables through. Configured keys can't be managed in the panel, which the docs say.

## Details worth a look in review

- **Hashing.** A plain unsalted SHA-256 is deliberate: a generated key is 32 random bytes and therefore not guessable, so a slow hash buys nothing, and it has to be computed on every API request. Same rationale as `AdminUser.RecoveryCodeHashes`.
- **Lookup.** Stored keys are found by an indexed lookup on that hash. Configured keys can't be, so they are all compared with `CryptographicOperations.FixedTimeEquals` without an early break — neither the duration nor the number of comparisons says how much of a guessed key was right.
- **Schema.** `admin.ApiKey`, next to `admin.AdminUser` and for the same reason: no game server database role may read it. Both tables share the context, so the migration is still driven by the admin user repository.
- **Last used.** Written at most once per minute per key, not awaited, and the repository swallows its own errors — it must never slow or fail an API request.
- The key itself is never written to the log.
- A request without any key is `NoResult()`, not a failure, so the cookie still gets its chance on the same request.
- The cookie handler answers `401`/`403` below `/api` instead of redirecting to the login page, which an API client can't use.
- Like the panel itself, the API is open while no admin user exists at all (the initial setup mode). That's existing `AdminAccessRequirement` behaviour, and the docs mention it.

## Possible follow-ups

Per-key IP allowlists, an expiry date, and rate limiting. None of them are needed to use the API, and each adds UI, so they seemed better kept separate.

## Testing

- `dotnet build src/MUnique.OpenMU.sln -p:ci=true` — succeeds, 0 errors.
- `dotnet test tests/MUnique.OpenMU.Web.Tests` — 38 passed, 0 failed. The 12 new tests cover configured and stored key lookup, rejection of unknown, too-short and disabled keys, the role build-up, that only a hash is stored, that the last-usage write is throttled, that generated keys are unique and prefixed, both header forms, and the no-key / wrong-key results.
- The migration was generated with `dotnet ef migrations add`, not written by hand.

https://claude.ai/code/session_01VZLARthjVEssPfKLacpaWw


sven-n:
Few things to improve:
* The field for 'Last used' is not required.
* When adding a new api key, it's not made visible to the user. Only a part (of the hash?) is visible at the api key table.
  * We want it to be displayed ONCE after it has been created, with the clear message that this key will never be displayed again. There should be a button to copy it to clipboard.
  * The role of the API key should not be modifyable afterwards.
* AdminPanel:Api:Keys and OPENMU_API_KEY / OPENMU_API_KEY_ROLES are not required. They don't have to exist when the system is installed initially. Only a few people need it and it's okay that they need to set this up first manually in the admin panel.

Issue #895, opened 2026-08-24
Network Analyzer as an Admin Panel page, incl. per-account traffic observation


sven-n:
**Is your feature request related to a problem? Please describe.**

The network analyzer (`src/Network/Analyzer`) is a WinForms tool which acts as a man-in-the-middle proxy. That works well for development, but an admin who wants to look at the traffic of a suspicious player usually can't set such a proxy up on the server machine. The traffic is already flowing through our own servers — we just don't expose it anywhere.

**Status:** done. All phases are merged: #896, #897, #901, #907, #916, #917 and #922.

**Describe the solution you'd like**

A `/network-analyzer` page in the admin panel:

* A collapsible connection list on the left, grouped by server type (connect server, game server, chat server) and by server instance.
* An entry shows the account and character name where we know it, and falls back to the remote endpoint (IP:port) where we don't.
* Selecting an entry starts capturing and shows the packets in a live-updating grid, with the same field extraction the WinForms tool does. The protocol version is applied automatically — no manual version dropdown.
* An account flag `IsNetworkObservationActive` makes the game server archive the traffic of that account's sessions on the file system. Archived sessions are browsable in the same page — while the player is online *and* after they went offline — and can be deleted.
* A button next to an online player (live map player list, online accounts page) navigates to the analyzer page with that connection preselected.

The WinForms tool **stays**. This is an addition, not a replacement — both use the same analysis code after the split described below.

---

#### 1.1 The analyzer tool

Only a few of its 15 files are actually WinForms:

| File | UI-bound? | Notes |
|---|---|---|
| `MainForm.*`, `PacketSenderForm.*`, `Program.cs` | yes | the app itself |
| `LiveConnection.cs`, `LiveConnectionListener.cs`, `ClientConnectedEventArgs.cs` | proxy only | only needed for the MITM mode |
| `PacketAnalyzer.cs` | no | resolves a packet to its definition and extracts the fields |
| `Packet.cs`, `ICapturedConnection.cs`, `SavedConnection.cs`, `CapturedConnectionExtensions.cs`, `FieldExtensions.cs` | no | model + `.mucap` save/load |

#### 1.2 The servers

All three server types create their client connections through the very same code path — `MUnique.OpenMU.Network.Listener.CreateConnection` returns a `MUnique.OpenMU.Network.Connection`:

* `GameServer/DefaultTcpGameServerListener.cs:81`
* `ConnectServer/ClientListener.cs:67`
[...]

PR #907, opened 2026-08-27
Add the network analyzer page to the admin panel

sven-n:
Phase 3 of #895 — the first part which is actually visible to an admin. Builds on the capture hook (#897) and the capture service (#901), both on master now.

## The page

`/network-analyzer` shows the connections of all servers of this process on the left and the traffic of the selected one on the right.

* The **connection list** is a collapsible sidebar, grouped by server (using the server description) and searchable by name or endpoint. Each entry can be selected or disconnected.
* Selecting a connection starts a capture and shows its packets in a **virtualized `QuickGrid`** with timestamp, direction, size, code and the extracted short information. Selecting a packet shows its **raw data and all extracted fields**.
* The packets can be filtered by direction and by a text which is matched against code, raw data and message. The view can be **paused** (the capture keeps running) and **cleared**.
* The **packet definition set and the client version come from the connection**, so connect server traffic is analyzed with the connect server definitions without asking the user — that's what the phase 1 changes to `PacketAnalyzer` were for.

The page and its menu entry only appear when an `IPacketCaptureService` is registered, which is the case in the all-in-one deployment. In the distributed deployment the page shows an explanatory note instead. The service is resolved through `IServiceProvider.GetService` rather than `[Inject]`, because Blazor throws for an unregistered service even on a nullable property — the same pattern `Servers.razor` uses for the optional game server dictionary.

## Two fixes found by running it

I ran the server in demo mode and drove the page in a browser, which turned up two things:

* **`PacketAnalyzer` loaded its definition files relative to the working directory.** They are copied next to the binaries, which is a different directory whenever the server isn't started from its own folder — the analyzer then threw `FileNotFoundException` and took the Blazor circuit down with it. It resolves them against `AppContext.BaseDirectory` now, and a missing file leaves the definitions empty instead of throwing (the grid then still shows the raw data). This affects the deployed server generally, not just the page.
* **`ICapturedConnectionInfo` now provides the description of its server**, so the groups are captioned "Connect Server (Season 6 Episode 3 GMO Client)" instead of "Connect Server 65536" — the connect server ids start at `SpecialServerIds.ConnectServer`.

## Verification

Besides the build and the test suite, I ran the whole thing: `dotnet run -- -demo -resolveIP:local`, opened TCP clients against a connect server and a game server, and drove the page in Chromium.

Both connections were listed under their servers, and selecting the connect server client showed its traffic — including a `C2` response, so the reassembly of the outgoing pipe from phase 1 is confirmed under real conditions:

```
00:00:00.746  C→S   4  F4  ServerListRequest
00:00:02.245  C→S   4  F4  ServerListRequest
00:00:02.247  S→C  19  F4  ServerListResponse (ServerCount: 3)
```

The detail pane showed `C2 00 13 F4 06 00 03 00 00 00 00 01 00 00 00 02 00 00 00` with the extracted `ServerListResponse / ServerCount: 3`. No Blazor errors, no console errors.

* `dotnet build MUnique.OpenMU.sln -p:ci=true` → 0 errors; no warning originates in a file this PR adds or changes.
* Complete test suite: 1601 tests, 0 failures.

## Notes

* There are no automated component tests — that would need bUnit as a new dependency. Say the word if you want it and I'll add the test project.
* The plan mentioned an autoscroll button; it needs JS interop for the grid's scroll container, so I left it out rather than half-doing it.
* `.gitignore` gained `data-protection-keys/`, which ASP.NET creates in the working directory when you run the server locally.

https://claude.ai/code/session_01Pb82LmoaUVdZtBtQs7xrtA

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pb82LmoaUVdZtBtQs7xrtA)_


sven-n:
grafik
I'm impressed, that's what claude came up with. He even tested it end-to-end inside a cloud environment - that's the original screenshot from claude.

PR #910, opened 2026-08-28
Fix crash on startup when the database contains no game configuration

### LINKED ISSUE #908: Server crashes on startup with empty database
**Describe the bug**
Server crashes without a `GameConfiguration`.

**To Reproduce**
I was able to get the database to a state where all the tables etc. were created, but the GameConfiguration was missing - by a failed import.
When starting the Startup project, the process crashes.

**Expected behavior**
The server should start. The admin panel should behave as if the database isn't created yet.

sven-n:
Fixes #908.

When the database schema exists but the `GameConfiguration` is missing (e.g. after a failed import), the Startup project crashed while building the host.

## Root causes

* `Program.CreateHostAsync`: the `ByDataSourceReferenceHandler` factory dereferenced the default game configuration id (`configId!.Value`) without a null check → `NullReferenceException`. It's resolved while constructing the `PlugInManager` singleton, so the host never started.
* `Program.CreateMissingPlugInConfigurations`: `context.GetAsync<GameConfiguration>()...First()` threw an `InvalidOperationException`, because `PlugInConfigurationsFactory` always tries to create the plugin configurations it finds missing — which is all of them on an empty database.

### Startup

* `Startup/Program.cs`: both paths handle a missing game configuration now. The configuration context in the reference handler factory is disposed properly, too.
* Since the server can now boot into an uninitialized state, the `PlugInManager` singleton starts without plugin configurations. `Program` subscribes to `SetupService.DatabaseInitialized` (before the host, and with it the server containers, is started) and reads the created plugin configurations afterwards.
* The data source of the plugin manager's reference handler is reloaded at that point. Otherwise the references within the custom plugin configurations would be resolved on a data source which is either not loaded at all (empty database) or holds the previous, deleted game configuration (re-install). The configurations are taken from that reloaded data source, so they stay alive as long as the reference handler does, and the registered `ICollection<PlugInConfiguration>` is updated as well.

### Plugin manager

* The configuration loop of the constructor moved into a public `ReadConfigurations(IEnumerable<PlugInConfiguration>)`, which can be called again when the configurations become available later.
* Reading the configurations again can also activate a plugin now, which is required when it was inactive in the previously read data. Previously read configurations aren't observed anymore, so the `PropertyChanged` handlers don't pile up.
* A failing configuration doesn't stop the remaining ones from being read anymore.

### Setup service and page

* `SetupService.CreateDatabaseAsync` awaits the subscribers of its `DatabaseInitialized` event one after another. A multicast delegate only returns the task of the last subscriber, so the plugin configurations were applied while the server containers were already restarting.
* The new `DataInitializationState` (`Initialized`, `NotInitialized`, `Unknown`) tells a failed or timed out check apart from an actually empty database. The setup page only offers the unconfirmed **Create** button for `NotInitialized`; `Unknown` keeps the confirmed **ReInstall** button.
* The page determines the state again after a schema update (it can't be determined on an outdated schema) and loads it before it hides the installer.

## Testing

* Full solution builds without errors.
* `MUnique.OpenMU.Tests` (984), `MUnique.OpenMU.Web.Tests` (95), `MUnique.OpenMU.PlugIns.Tests` (41), `MUnique.OpenMU.ChatServer.Tests` (26) and `MUnique.OpenMU.Persistence.Initialization.Tests` (21, 6 skipped) pass.
* Not verified by an end-to-end startup against a real PostgreSQL instance - that wasn't available in the used environment.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SF8VVTJtUjimJfVKa32Lxm)_

### DIFF
--- a/src/PlugIns/PlugInManager.cs
+++ b/src/PlugIns/PlugInManager.cs
@@ -48,11 +48,7 @@ public class PlugInManager
         if (configurations is not null)
         {
             this.DiscoverAndRegisterPlugIns();
-            var loadedAssemblies = new HashSet<string>();
-            foreach (var configuration in configurations)
-            {
-                this.ReadConfiguration(configuration, loadedAssemblies);
-            }
+            this.ReadConfigurations(configurations);
         }
     }
 
@@ -83,6 +79,21 @@ public class PlugInManager
     /// Gets the reference handler for references in custom plugin configurations.
     /// </summary>
     public ReferenceHandler? CustomConfigReferenceHandler { get; }
+
+    /// <summary>
+    /// Reads the given plugin configurations and applies them to the known plugins.
+    /// It can be called again later, e.g. when the configurations became available
+    /// after the database has been initialized.
+    /// </summary>
+    /// <param name="configurations">The plugin configurations.</param>
+    public void ReadConfigurations(IEnumerable<PlugInConfiguration> configurations)
+    {
+        var loadedAssemblies = new HashSet<string>();
+        foreach (var configuration in configurations)
+        {
+            this.ReadConfiguration(configuration, loadedAssemblies);
+        }
+    }
 
     /// <summary>
     /// Discovers and registers all plugins of all loaded assemblies.
--- a/src/Startup/Program.cs
+++ b/src/Startup/Program.cs
@@ -304,8 +304,19 @@ internal sealed class Program : IDisposable
                 var dataSource = new GameConfigurationDataSource(
                     provider.GetService<ILogger<GameConfigurationDataSource>>()!,
                     persistenceContextProvider!);
-                var configId = persistenceContextProvider!.CreateNewConfigurationContext().GetDefaultGameConfigurationIdAsync(default).AsTask().WaitAndUnwrapException();
-                dataSource.GetOwnerAsync(configId!.Value).AsTask().WaitAndUnwrapException();
+                using var configurationContext = persistenceContextProvider!.CreateNewConfigurationContext();
+                var configId = configurationContext.GetDefaultGameConfigurationIdAsync(default).AsTask().WaitAndUnwrapException();
+                if (configId is { } gameConfigurationId)
+                {
+                    dataSource.GetOwnerAsync(gameConfigurationId).AsTask().WaitAndUnwrapException();
+                }
+                else
+                {
+                    // The database doesn't contain a game configuration yet. It can be created later,
+                    // e.g. through the admin panel, so we just don't load it here.
+                    this._logger.Warning("No game configuration found in the database.");
+                }
+
                 var referenceHandler = new ByDataSourceReferenceHandler(dataSource);
                 return referenceHandler;
             })
@@ -325,6 +336,12 @@ internal sealed class Program : IDisposable
         {
             host.ConfigureAdminPanel();
         }
+
+        // When the server is started with an uninitialized database, the plugin configurations
+        // are not available yet. They're created during the data initialization, so we have to
+        // load them afterwards. We subscribe before the host is started, so that this handler
+        // runs before the server containers are restarted.
+        host.Services.GetRequiredService<SetupService>().DatabaseInitialized += () => this.OnDatabaseInitializedAsync(host.Services);
 
         this._logger.Information("Starting host...");
         var stopwatch = new Stopwatch();
@@ -378,18 +395,60 @@ internal sealed class Program : IDisposable
             return configs;
         }
 
+        if (!this.IsGameConfigurationAvailable(persistenceContextProvider))
+        {
+            // Without a game configuration, we can't add the missing plugin configurations,
+            // because they are part of it. This is the case when the database is not initialized yet.
+            this._logger.Warning("No game configuration found in the database, so the missing plugin configurations can't be created yet.");
+            return configs;
+        }
+
         configs.AddRange(this.CreateMissingPlugInConfigurations(typesWithMissingConfigs, persistenceContextProvider, referenceHandler));
         _ = context.SaveChangesAsync().AsTask().WaitAndUnwrapException();
         return configs;
     }
 
+    private async ValueTask OnDatabaseInitializedAsync(IServiceProvider services)
+    {
+        try
+        {
+            if (services.GetService<PlugInManager>() is not { } plugInManager
+                || services.GetService<IPersistenceContextProvider>() is not { } persistenceContextProvider)
+            {
+                return;
+            }
+
+            using var context = persistenceContextProvider.CreateNewTypedContext(typeof(PlugInConfiguration), false);
+            var configurations = (await context.GetAsync<PlugInConfiguration>().ConfigureAwait(false)).ToList();
+            plugInManager.ReadConfigurations(configurations);
+            this._logger.Information("Applied {count} plugin configurations after the database initialization.", configurations.Count);
+        }
+        catch (Exception ex)
+        {
+            this._logger.Error(ex, "Error when applying the plugin configurations after the database initialization.");
+        }
+    }
+
+    private bool IsGameConfigurationAvailable(IPersistenceContextProvider persistenceContextProvider)
+    {
+        using var context = persistenceContextProvider.CreateNewConfigurationContext();
+        return context.GetDefaultGameConfigurationIdAsync(default).AsTask().WaitAndUnwrapException() is not null;
+    }
+
     private IEnumerable<PlugInConfiguration> CreateMissingPlugInConfigurations(IEnumerable<Type> plugInTypes, IPersistenceContextProvider persistenceContextProvider, ReferenceHandler referenceHandler)
     {
-        GameConfiguration gameConfiguration;
+        GameConfiguration? gameConfiguration;
 
         using (var context = persistenceContextProvider.CreateNewContext())
         {
-            gameConfiguration = context.GetAsync<GameConfiguration>().AsTask().WaitAndUnwrapException().First();
+            gameConfiguration = context.GetAsync<GameConfiguration>().AsTask().WaitAndUnwrapException().FirstOrDefault();
+        }
+
+        if (gameConfiguration is null)
+        {
+            // The database is not initialized yet - the plugin configurations are created
+            // together with the game configuration.
+            yield break;
         }
 
         using var saveContext = persistenceContextProvider.CreateNewContext(gameConfiguration);
--- a/src/Web/AdminPanel/Pages/Setup.razor
+++ b/src/Web/AdminPanel/Pages/Setup.razor
@@ -7,7 +7,7 @@
 
 @if (this.ShowInstall)
 {
-    <Install InstallationFinished="() => this.ShowInstall = false" />
+    <Install InstallationFinished="this.OnInstallationFinishedAsync" />
 }
 else if (!this.SetupService.CanConnectToDatabase)
 {
@@ -26,17 +26,19 @@ else if (this.SetupService.IsUpdateRequired)
     <p>@Resources.DatabaseStatus: <span class="badge bg-warning text-dark">@Resources.UpdateRequired</span></p>
     <button class="btn btn-primary" @onclick="this.OnUpdateClickAsync">@Resources.Update</button>
 }
+else if (!this._isDataInitialized)
+{
+    @* The database exists, but doesn't contain any data yet. For the user, it's the same
+       as if the database wasn't created yet - it just needs to be installed. *@
+    <p>@Resources.DatabaseStatus: <span class="badge bg-success">@Resources.UpToDate</span></p>
+    <p>@Resources.InitializedGameVersion: <span class="badge bg-warning text-dark">@Resources.NoInitializedDataFound</span></p>
+
+    <button class="btn btn-primary" @onclick="this.OnInstallClick">@Resources.Create</button>
+}
 else
 {
     <p>@Resources.DatabaseStatus: <span class="badge bg-success">@Resources.UpToDate</span></p>
-    @if (!this._isDataInitialized)
-    {
-        <p>@Resources.InitializedGameVersion: <span class="badge bg-warning text-dark">@Resources.NoInitializedDataFound</span></p>
-    }
-    else
-    {
-        <p>@Resources.InitializedGameVersion: <span class="badge bg-success">@this._gameClientVersion</span></p>
-    }
+    <p>@Resources.InitializedGameVersion: <span class="badge bg-success">@this._gameClientVersion</span></p>
 
     <button class="btn btn-warning" @onclick="this.OnReInstallClickAsync">@Resources.ReInstall</button>
 }
--- a/src/Web/AdminPanel/Pages/Setup.razor.cs
+++ b/src/Web/AdminPanel/Pages/Setup.razor.cs
@@ -39,13 +39,23 @@ public partial class Setup
     public IJSRuntime JsRuntime { get; set; } = null!;
 
     /// <inheritdoc />
-    protected override async Task OnInitializedAsync()
+    protected override Task OnInitializedAsync()
+    {
+        return this.LoadDataStateAsync();
+    }
+
+    private async Task LoadDataStateAsync()
     {
         this._isDataInitialized = await this.SetupService.IsDataInitializedAsync().ConfigureAwait(false);
-        if (this._isDataInitialized)
-        {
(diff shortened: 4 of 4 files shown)


src/Startup/Program.cs
```diff
@@ -378,18 +395,60 @@ private ICollection<PlugInConfiguration> PlugInConfigurationsFactory(IServicePro
...
+        {
+            if (services.GetService<PlugInManager>() is not { } plugInManager
+                || services.GetService<IPersistenceContextProvider>() is not { } persistenceContextProvider)
+            {
+                return;
+            }
+
+            using var context = persistenceContextProvider.CreateNewTypedContext(typeof(PlugInConfiguration), false);
+            var configurations = (await context.GetAsync<PlugInConfiguration>().ConfigureAwait(false)).ToList();
+            plugInManager.ReadConfigurations(configurations);
```
sven-n, line 469: **Plugin configurations containing references fail here after an install, which stops the whole loop.**

`PlugInManager` is a singleton. It got its `CustomConfigReferenceHandler` at startup, when the database was still empty, so the handler's `GameConfigurationDataSource` never loaded an owner (the new `else` branch in the `ReferenceHandler` factory skips `GetOwnerAsync`).

Path: `ReadConfigurations` → `ConfigurePlugIn` → `PlugInContainerBase.OnPlugInConfigurationChanged` → `GetConfiguration(configType, this.Manager.CustomConfigReferenceHandler)` → `ByDataSourceReferenceResolver.ResolveReference` → `DataSourceBase.Owner` → `InvalidOperationException("owner is not loaded.")`.

The installed data does contain a reference: `DataInitializationBase` adds the Fenrir item to the Bless Jewel plugin's configuration. Nothing on that path catches the exception, so it ends `ReadConfigurations`. Every configuration after that one is never applied, including the disabled-by-default plugins, which stay active. Your catch below only logs it.

On a ReInstall of a populated database, the handler instead looks up references in the old, deleted game configuration.

Suggestion: reload the handler's data source when `DatabaseInitialized` fires (e.g. call `GetOwnerAsync` with the new default config id before `ReadConfigurations`), or have `GameConfigurationDataSource` load its owner on first use.

---

src/Web/AdminPanel/Pages/Setup.razor
```diff
@@ -26,17 +26,19 @@ else if (this.SetupService.IsUpdateRequired)
...
     <button class="btn btn-primary" @onclick="this.OnUpdateClickAsync">@Resources.Update</button>
 }
+else if (!this._isDataInitialized)
+{
+    @* The database exists, but doesn't contain any data yet. For the user, it's the same
+       as if the database wasn't created yet - it just needs to be installed. *@
+    <p>@Resources.DatabaseStatus: <span class="badge bg-success">@Resources.UpToDate</span></p>
+    <p>@Resources.InitializedGameVersion: <span class="badge bg-warning text-dark">@Resources.NoInitializedDataFound</span></p>
+
+    <button class="btn btn-primary" @onclick="this.OnInstallClick">@Resources.Create</button>
```
sven-n: **This "Create" button wipes the database with no confirmation.**

`OnInstallClick` opens the installer, which runs `ReCreateDatabaseAsync` directly. ReInstall asks for confirmation (`ReinstallConfirmation`) first.

`_isDataInitialized` comes from `SetupService.IsDataInitializedAsync`, which returns `false` on any exception and on its 3-second timeout. A slow or failing query on a fully populated database therefore shows this branch, and one click drops the data. Before this PR, that case still showed ReInstall, which asks.

Suggestion: keep the confirmation dialog on this path too, or make `IsDataInitializedAsync` tell "no configuration found" apart from "check failed", and only offer Create in the first case.

---

src/Startup/Program.cs
```diff
@@ -326,6 +337,12 @@ private async Task<IHost> CreateHostAsync(string[] args)
             host.ConfigureAdminPanel();
         }
 
+        // When the server is started with an uninitialized database, the plugin configurations
+        // are not available yet. They're created during the data initialization, so we have to
+        // load them afterwards. We subscribe before the host is started, so that this handler
+        // runs before the server containers are restarted.
+        host.Services.GetRequiredService<SetupService>().DatabaseInitialized += () => this.OnDatabaseInitializedAsync(host.Services);
```
sven-n, line 367: **Subscribing first doesn't make this handler finish before the containers restart.**

`SetupService.CreateDatabaseAsync` does `await eventHandler.Invoke()` on a multicast `AsyncEventHandler`. That calls every subscriber in turn but only returns the **last** subscriber's `ValueTask`.

This handler runs until its first real await (`context.GetAsync<PlugInConfiguration>()`), and then `GameServerContainer`/`ConnectServerContainer.OnDatabaseInitializedAsync` start `RestartAllAsync` at the same time. Game servers can come up before the disabled-by-default plugins are switched off, and `PlugInManager` state (the `_plugInPoints` dictionary isn't thread-safe) is changed at the same time as server startup.

Suggestion: make `CreateDatabaseAsync` await each subscriber in turn (`GetInvocationList()`), or call this step explicitly before the event is raised. Either way, the comment should match what actually happens.

---

src/Web/AdminPanel/Pages/Setup.razor
```diff
@@ -26,17 +26,19 @@ else if (this.SetupService.IsUpdateRequired)
     <p>@Resources.DatabaseStatus: <span class="badge bg-warning text-dark">@Resources.UpdateRequired</span></p>
     <button class="btn btn-primary" @onclick="this.OnUpdateClickAsync">@Resources.Update</button>
 }
+else if (!this._isDataInitialized)
```
sven-n: **`_isDataInitialized` isn't reloaded after an update.**

It's only loaded in `OnInitializedAsync` and after an install. Suppose a database needs a schema update: `IsDataInitializedAsync` fails against the old schema, and the exception is swallowed, so it returns `false`. After the admin clicks Update, `IsUpdateRequired` becomes `false` and the page falls into this branch. It then offers the unconfirmed "Create" button for a database that actually has data.

Suggestion: call `LoadDataStateAsync()` after `InstallUpdatesAsync` in `OnUpdateClickAsync`.

---

src/PlugIns/PlugInManager.cs
```diff
@@ -84,6 +80,21 @@ public PlugInManager(ICollection<PlugInConfiguration>? configurations, ILoggerFa
...
+    /// It can be called again later, e.g. when the configurations became available
+    /// after the database has been initialized.
+    /// </summary>
+    /// <param name="configurations">The plugin configurations.</param>
+    public void ReadConfigurations(IEnumerable<PlugInConfiguration> configurations)
+    {
+        var loadedAssemblies = new HashSet<string>();
+        foreach (var configuration in configurations)
+        {
+            this.ReadConfiguration(configuration, loadedAssemblies);
```
sven-n: **Calling this a second time can't turn a plugin back on.**

`ReadConfiguration` only calls `DeactivatePlugIn` when `IsActive == false`, and never `ActivatePlugIn`. That's fine at construction time, but now it also runs after a ReInstall. If plugin X was turned off in the old data and is on in the new data, X stays off until the process restarts, even though the log says the configurations were applied.

Suggestion: when called again, activate plugins whose configuration has `IsActive == true`, as `OnConfigurationChanged` / `PlugInConfigurationChangeApplier` already do.

---

src/Startup/Program.cs
```diff
@@ -378,18 +395,60 @@ private ICollection<PlugInConfiguration> PlugInConfigurationsFactory(IServicePro
...
+    {
+        try
+        {
+            if (services.GetService<PlugInManager>() is not { } plugInManager
+                || services.GetService<IPersistenceContextProvider>() is not { } persistenceContextProvider)
+            {
+                return;
+            }
+
+            using var context = persistenceContextProvider.CreateNewTypedContext(typeof(PlugInConfiguration), false);
```
sven-n: **These configurations come from a context that is disposed right away. `ReadConfiguration` attaches a `PropertyChanged` handler to each of them** (`PlugInManager.cs`, `configuration.PropertyChanged += ...`).

Those objects are thrown away, so the handlers never react to later admin-panel edits. Each ReInstall adds another set of handlers that keep the old objects and the manager alive. The singleton `ICollection<PlugInConfiguration>` also still holds the startup (empty) list.

Minor, but worth checking whether the handlers belong on these instances at all.

---

src/Web/AdminPanel/Pages/Setup.razor.cs
```diff
@@ -39,13 +39,23 @@ public partial class Setup
...
-            this._gameClientVersion = await this.SetupService.GetCurrentGameClientVersionAsync().ConfigureAwait(false);
-        }
+        this._gameClientVersion = this._isDataInitialized
+            ? await this.SetupService.GetCurrentGameClientVersionAsync().ConfigureAwait(false)
+            : null;
+    }
+
+    private async Task OnInstallationFinishedAsync()
+    {
+        this.ShowInstall = false;
```
sven-n: **The page can briefly show the wrong state here.**

`ShowInstall` is set to `false` before `LoadDataStateAsync` finishes. Blazor can re-render at the await with the stale `_isDataInitialized == false`, so the page briefly shows "No initialized data found" and the unconfirmed Create button for the database that was just installed.

Suggestion: reload the state first, then hide the installer.

---

src/Startup/Program.cs
```diff
@@ -378,18 +395,60 @@ private ICollection<PlugInConfiguration> PlugInConfigurationsFactory(IServicePro
             return configs;
         }
 
+        if (!this.IsGameConfigurationAvailable(persistenceContextProvider))
```
sven-n: Nit: this check and the new `FirstOrDefault()` / `yield break` in `CreateMissingPlugInConfigurations` test the same thing with different queries (default config id vs. the first `GameConfiguration` row). That costs an extra database round trip, and the two checks could disagree. I'd keep just one of them.

---

src/Startup/Program.cs
```diff
@@ -304,8 +304,19 @@ private async Task<IHost> CreateHostAsync(string[] args)
...
+                var configId = configurationContext.GetDefaultGameConfigurationIdAsync(default).AsTask().WaitAndUnwrapException();
+                if (configId is { } gameConfigurationId)
+                {
+                    dataSource.GetOwnerAsync(gameConfigurationId).AsTask().WaitAndUnwrapException();
+                }
+                else
+                {
+                    // The database doesn't contain a game configuration yet. It can be created later,
+                    // e.g. through the admin panel, so we just don't load it here.
+                    this._logger.Warning("No game configuration found in the database.");
```
sven-n: Nit: this factory is registered as transient. On an empty database, every resolve logs this warning and returns a handler around a data source with no owner, which fails later far from here ("owner is not loaded"). This only moves the crash from startup to first use. Fixing the root cause (see the comment on `ReadConfigurations`) would also remove this.

---

sven-n (commented):
The startup fix works: an empty database no longer crashes the server. The follow-up after an install has problems, though:

- **Plugin configurations fail after an install.** `PlugInManager` keeps the reference handler it got at startup, whose data source never loaded a game configuration. The first custom configuration with a reference throws and stops the loop (see `Program.cs:423`).
- **The new "Create" button wipes the database without asking.** `IsDataInitializedAsync` returns `false` on a timeout or error, so a populated database can land on this branch (see `Setup.razor:36`).
- **This handler doesn't finish before the servers restart.** The multicast `AsyncEventHandler` only returns the last subscriber's task, so the handler races `RestartAllAsync` (see `Program.cs:344`).

The rest are smaller follow-ups and nits.

---

src/Startup/Program.cs
sven-n: Fixed by reloading the data source. `ByDataSourceReferenceHandler` now exposes its `DataSource`, and `LoadPlugInConfigurationsAsync` calls `ForceDiscardChangesAsync()` (to drop the old context and owner) followed by `GetOwnerAsync()` before the configurations are read. That covers the re-install case with the deleted game configuration, too.

The configurations themselves are now taken from that reloaded data source (`gameConfiguration.PlugInConfigurations`) instead of a separate, immediately disposed context.

Additionally, `ReadConfigurations` catches and logs per configuration now, so a single failing one doesn't stop the remaining ones.

---

src/Web/AdminPanel/Pages/Setup.razor
sven-n: Fixed by telling the two cases apart: `SetupService.GetDataInitializationStateAsync` returns a new `DataInitializationState` (`Initialized`, `NotInitialized`, `Unknown`), where a failed or timed out check results in `Unknown`.

The unconfirmed Create button is only shown for `NotInitialized`. `Unknown` falls into the last branch with the confirmed ReInstall button, like before this PR. The field also starts as `Unknown`, so the first render (before `OnInitializedAsync` finished) can't show the Create button either.

`IsDataInitializedAsync` is kept for the updates page and now delegates to the new method.

---

src/Startup/Program.cs
sven-n: You're right, thanks. `CreateDatabaseAsync` now awaits the subscribers one after another over `GetInvocationList().OfType<AsyncEventHandler>()`, so the registration order actually decides the execution order. The comment at the subscription says that now, too.

---

src/Web/AdminPanel/Pages/Setup.razor
sven-n: Fixed - `OnUpdateClickAsync` awaits `InstallUpdatesAsync` and then calls `LoadDataStateAsync()`. Together with the new `Unknown` state, a failing check on the outdated schema doesn't lead to the unconfirmed Create button anymore either.

---

src/PlugIns/PlugInManager.cs
sven-n: Fixed - `ReadConfiguration` now activates the plugin when the configuration is active and deactivates it otherwise, like `OnConfigurationChanged` does. At construction time that's a no-op for active plugins (they're registered as active, and `PlugInContainerBase.OnPlugInActivated` returns early when the plugin is already in the container).

---

src/Startup/Program.cs
sven-n: Good catch, both parts are addressed:

* The configurations are now taken from the reference handler's data source, whose context stays alive as long as the handler does - so they aren't objects of a context that was disposed right away.
* `ReadConfigurations` unsubscribes from the previously read configurations before it reads the new ones (the handlers are kept in `_configurationSubscriptions`), so they don't pile up with each re-install.
* The registered `ICollection<PlugInConfiguration>` is refilled with the new configurations in `OnDatabaseInitializedAsync`.

Whether those `PropertyChanged` handlers are needed at all is a separate question - admin panel edits are delivered through `IConfigurationChangePublisher` on different instances - so I left the mechanism as it is.

---

src/Web/AdminPanel/Pages/Setup.razor.cs
sven-n: Fixed - `OnInstallationFinishedAsync` loads the state first and hides the installer afterwards.

---

src/Startup/Program.cs
sven-n: Removed - only the `FirstOrDefault()` check in `CreateMissingPlugInConfigurations` is left, which also logs the warning now. That's one query instead of two, and the two checks can't disagree anymore.

---

src/Startup/Program.cs
sven-n: The root cause is fixed now (see the reply on `ReadConfigurations`): the data source is reloaded as soon as the database got initialized, so a handler that was created while the database was still empty doesn't stay unloaded.

The log message here is now a `Debug` one and says that the data source is loaded later, so it doesn't spam a warning per resolve. I kept the eager loading itself, because the synchronous `Get`/`GetAll` of `DataSourceBase` require a loaded owner - making the data source load itself on first use would mean blocking on an async load inside the JSON deserialization, which I'd rather not add here.

---

src/Startup/Program.cs
sven-n: Follow-up on the unsubscription: Codacy's quality gate flagged the lambda I used for it (unused `sender` parameter, and the `using System.ComponentModel;` it needed), so `da8fd980` subscribes with a method group instead.

That removes the need to keep the created delegate - remembering the observed configurations is enough, because `+=` and `-=` of the same method group are equal delegates. The plugin type is looked up by the type id of the sender, which `PlugInConfiguration.OnPropertyChanged` passes when it raises the event. It also saves a closure per configuration.

---

PR #913, opened 2026-08-29
Stop orphaned walk loops from spinning a CPU core forever

### LINKED ISSUE #911: An orphaned Walker walk loop spins a CPU core for the lifetime of the process
## What happens

A `Walker` whose walk loop gets orphaned never stops. The loop keeps running at full speed
until the process exits, pinning one core and keeping its walk supporter (a player, a monster,
a summon) alive with it. Because each orphaned loop is permanent, the damage accumulates: CPU
usage rises in *steps* — one step per stranded loop — and memory rises linearly.

Observed on a two-core server running a bot population: both cores pegged (~196%), RSS growing
from 1188 MB to 3104 MB over three days (~640 MB/day, monotonic) until the box went into swap.

## Why bots make it worse, and why it is not a bot bug

`Walker` is used by every moving object. Bots just hit it far more often: the bot navigator
re-paths roughly every 2 seconds across every animated bot, so concurrent walk requests on the
same object are routine. Any server with enough movement traffic can reach the same state.

## Root cause — two defects in `src/GameLogic/Walker.cs`

**1. `StartWalkAsync` replaces `_walkCts` without cancelling the previous source.**
The token derived from that source is the only thing that can end the loop it belongs to. Once
the field points at a newer source, the older loop can never be stopped — by anyone.

**2. `WalkLoopAsync` uses `continue` (not `break`) when `WalkStepAsync` returns null.**
That relies on the preceding `StopAsync()` having cancelled *this loop's own* token, so the
while-condition ends it on the next pass. But `StopAsync` is a silent no-op once `_walkCts` no
longer refers to this loop's source. The null step is reproducible, so the loop retries it
**without ever awaiting anything** — a tight, non-yielding spin.

**3. A related race in `StopAsync` itself.** It cancels whatever `_walkCts` happens to be when
it acquires the writer lock, which is not necessarily the walk the caller started. A loop that
has run out of steps enters `StopAsync` and blocks on `WriterLockAsync()` — an acquisition that
takes no token, so cancelling that loop cannot release it. If `StartWalkAsync` wins the lock
first, it cancels the old loop, installs a new walk and releases; the old loop then acquires the
lock and cancels the walk that just started. The player or bot stops moving one step after being
told to walk.

**4. `InitializeWalkToAsync` replaces the step queue without ending the walk that is reading it.**
A loop still running from the previous request can dequeue and execute the steps of the walk that
has not been started yet, moving the supporter along the new path early and leaving the new loop
one step short. Both callers in the engine stop the walk first, so this is the window between
their stop and their start rather than something they do wrong.

Callers do generally stop the running walk before starting a new one (`PlayerMovement.WalkToAsync`
does call `_walker.StopAsync()` first), so this is a race, not a missing call — which is why the
fix belongs in `Walker` regardless of caller behaviour.

## How it was diagnosed

- `sar -r` / `sar -u` history showed the step-shaped CPU curve against the linear memory curve.
- GC gen2 at 2.70 GB confirmed a managed leak rather than native growth or fragmentation.
- A `dotnet-trace` sampled-thread-time profile attributed 26% of on-CPU time to
  `Walker.StopAsync` and 34% to a saturated finalizer thread.

## Reproduction

Regression tests drive each case and count how often the walk supporter is polled during a 200 ms
window *after* every started walk should have finished. Against current `master`:

| Case | Polls in the idle window |
|---|---|
| Starting a second walk while the first still runs | 48,141 |
| Starting the same walk twice | 89,644 |
| Eight tasks restarting the same walker concurrently | 23,927 |

A healthy `Walker` reports none of these.

I have a fix and will open a PR referencing this issue.

facuc28:
Fixes #911

A `Walker` whose walk loop got orphaned would spin at full speed for the lifetime of the
process, keeping its walk supporter alive with it. On a two-core server running a bot
population this pegged both cores and grew the heap by ~640 MB/day until the box went into
swap; a profile attributed 26% of on-CPU time to `Walker.StopAsync` and 34% to the saturated
finalizer thread.

`Walker` is used by every moving object — bots only make it likelier, because the navigator
re-paths every couple of seconds across the whole population.

## The two commits

**1. `Stop orphaned walk loops from spinning a CPU core forever`**

- `StartWalkAsync` cancels and disposes the previous `CancellationTokenSource` before replacing
  it. The token derived from a source is the only thing that can end the loop it belongs to, so
  once the field pointed elsewhere that loop could never be stopped by anyone.
- `WalkLoopAsync` leaves the loop instead of continuing it after a null step. A null step means
  the walk is over for good, so there is nothing to retry — and the previous `continue` relied
  on a `StopAsync` that is a silent no-op for an orphaned loop, producing a spin with no await
  in it at all.

**2. `Only let a walk loop stop the walk it owns`**

The first commit leaves a narrower race. `StopAsync` cancels whatever `_walkCts` happens to be
when it acquires the writer lock, which is not necessarily the walk the caller started. A loop
that has run out of steps enters `StopAsync` and blocks on `WriterLockAsync()` — an acquisition
that takes no token, so cancelling that loop cannot release it. If `StartWalkAsync` wins the
lock first, it cancels the old loop, installs a new walk and releases; the old loop then
acquires the lock and cancels the walk that just started, and the player stops moving one step
after being told to walk.

The loop now gets its own `CancellationTokenSource` and routes self-stops through
`StopOwnWalkAsync`, which only stops the walk while that source is still the current one. The
public `StopAsync` keeps its old meaning — outside callers are ending "the walk of this
object", not one particular walk — and both share `StopCurrentWalkAsync`.

`InitializeWalkToAsync` ends the running walk too, now that it can do so without releasing the
writer lock. It replaces the step queue, so a loop left running there would dequeue and execute
the steps of a walk that has not been started yet — moving the supporter along the new path
early and leaving the new loop one step short. Both callers in the engine (`PlayerMovement` and
`Monster`) already stop the walk before initializing the next one, so this closes the window
between their stop and their start rather than changing what they do.

## Tests

`tests/MUnique.OpenMU.Tests/WalkerTest.cs` is new. It uses a Moq double of `ISupportWalk` (cast
`.As<IAttackable>()`) whose `IsAlive` getter counts its reads, then measures reads in a 200 ms
window after every started walk should have finished. All four tests pass with the fix and three
of them fail against `master`:

| Test | Polls in the idle window against `master` |
|---|---|
| `StartingASecondWalkDoesNotLeaveASpinningLoopAsync` | 48,141 |
| `StartingTheSameWalkTwiceDoesNotLeaveASpinningLoopAsync` | 89,644 |
| `ConcurrentWalkRestartsLeaveNoRunningLoopAsync` | 23,927 |
| `DisposingWhileWalkingEndsTheLoopAsync` | passes either way — it guards the dispose path |

The threshold is 10 reads, so the margin is four orders of magnitude. (The counts are lower than
a hand-written double would give purely because Moq's interception slows the spin down.)

The concurrent test starts each walk twice on purpose: a single start loses the race against
the other seven tasks most of the time, because `StartWalkAsync` returns early once another task
initialized its own walk in between, and then no loop is ever started to strand.

Full `MUnique.OpenMU.Tests` suite passes (791/791), three runs in a row.

## Soak time

These two commits have been running on a private server with an active bot population since
2026-08-28. The step-shaped CPU growth and the ~640 MB/day heap growth both stopped.

### DIFF
--- a/src/GameLogic/Walker.cs
+++ b/src/GameLogic/Walker.cs
@@ -107,6 +107,18 @@ public sealed class Walker : IDisposable
             return;
         }
 
+        if (this._walkCts is { } previousCts)
+        {
+            // A walk loop is still running for a previous walk. Replacing the source without cancelling it
+            // first orphans that loop: its token is the only thing which can end it, and once the field
+            // points at the new source nobody holds a reference to the old one any more. The orphan would
+            // then run forever - see the note in WalkLoopAsync - burning a core and keeping the walk
+            // supporter alive. Callers are expected to StopAsync first; this is the safety net for the
+            // window in which they did not.
+            await previousCts.CancelAsync().ConfigureAwait(false);
+            previousCts.Dispose();
+        }
+
         var cts = new CancellationTokenSource();
         this._walkCts = cts;
         _ = Task.Run(async () => await this.WalkLoopAsync(cts.Token).ConfigureAwait(false), cts.Token);
@@ -194,7 +206,16 @@ public sealed class Walker : IDisposable
                     await this.StopAsync().ConfigureAwait(false);
                 }
 
-                continue;
+                // A null step means this walk is over for good - the walker is disposed, the supporter is
+                // no longer active, the queue ran dry, or the token was cancelled. There is nothing to
+                // retry, so leave the loop instead of continuing it.
+                //
+                // Continuing here used to be load-bearing: it relied on the StopAsync above having
+                // cancelled our own token, so that the while-condition ended the loop on the next pass.
+                // That assumption breaks whenever _walkCts no longer refers to this loop's source, because
+                // StopAsync is then a no-op which leaves our token uncancelled - and since the null step
+                // is reproducible, the loop spins at full speed without ever awaiting anything.
+                break;
             }
 
             var delay = this.StepDelay(step);
--- a/src/GameLogic/Walker.cs
+++ b/src/GameLogic/Walker.cs
@@ -85,6 +85,12 @@ public sealed class Walker : IDisposable
             }
         }
 
+        // End the running walk before the new steps go into the queue: otherwise a loop of the previous
+        // walk can dequeue and execute the new walk's steps in the window between this method and
+        // StartWalkAsync, moving the supporter along the new path early and leaving the new loop one step
+        // short. We hold the writer lock here, so this is safe to do without releasing it.
+        await this.StopCurrentWalkAsync().ConfigureAwait(false);
+
         this.CurrentTarget = target;
         EnqueueSteps();
 
--- a/src/GameLogic/Walker.cs
+++ b/src/GameLogic/Walker.cs
@@ -113,15 +113,19 @@ public sealed class Walker : IDisposable
             // first orphans that loop: its token is the only thing which can end it, and once the field
             // points at the new source nobody holds a reference to the old one any more. The orphan would
             // then run forever - see the note in WalkLoopAsync - burning a core and keeping the walk
-            // supporter alive. Callers are expected to StopAsync first; this is the safety net for the
-            // window in which they did not.
+            // supporter alive.
             await previousCts.CancelAsync().ConfigureAwait(false);
             previousCts.Dispose();
         }
 
         var cts = new CancellationTokenSource();
+        var cancellationToken = cts.Token;
         this._walkCts = cts;
-        _ = Task.Run(async () => await this.WalkLoopAsync(cts.Token).ConfigureAwait(false), cts.Token);
+
+        // The loop is handed its own source as well as its token: it needs the source to recognize whether
+        // the walk it is running is still the current one before it stops anything. It cannot read that back
+        // from the field later (that is the whole point), nor from cts.Token once the source is disposed.
+        _ = Task.Run(async () => await this.WalkLoopAsync(cts, cancellationToken).ConfigureAwait(false), cancellationToken);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


sven-n:
## Review

The diagnosis is right and the two core fixes are correct. CI is green (Azure build + Codacy). One thing needs to be resolved before merge; the rest are notes.

### The fixes themselves — correct

1. **`StartWalkAsync` cancels/disposes the previous CTS.** Accurate root cause: on `master`, `_walkCts` is overwritten without cancelling, and since the loop's token is the only thing that can end it, the old loop becomes unreachable. `StopAsync` then silently no-ops for it (`_walkCts` points elsewhere), and combined with `continue` after a reproducible null step you get a loop with no await in it at all. That matches the reported profile.
2. **`break` instead of `continue`** removes the await-free path. Sound.
3. **`StopOwnWalkAsync` with `ReferenceEquals(_walkCts, ownCts)`** correctly closes the second race — a loop parked on the untokened `WriterLockAsync()` cannot be released by cancelling it, so it would otherwise cancel the walk that just started. Passing the source *and* the token into the loop is the right call, since it can't re-read either after disposal.

Disposal ordering is safe as written: `CancelAsync()` precedes `Dispose()`, so the stale token is already cancelled and later registrations from `ReaderLockAsync(token)` fire inline rather than throwing `ObjectDisposedException`. No deadlock either — `CancelAsync` under the writer lock only completes pending waiters as cancelled, while `StopOwnWalkAsync` waits for the lock to be released normally.

### Needs a change

**`InitializeWalkToAsync` was never changed.** Both the PR description and the second commit message state:

> `InitializeWalkToAsync` ends the running walk too, now that it can do so without releasing the writer lock.

`src/GameLogic/Walker.cs` on the PR head is identical to `master` in that method — there is no `StopCurrentWalkAsync()` call. So the window described in that paragraph is still open: a loop from a previous walk can dequeue and execute the *new* walk's steps between `InitializeWalkToAsync` and `StartWalkAsync`, moving the supporter along the new path early and leaving the new loop one step short.

Either add the call (one line before `EnqueueSteps()` — the writer lock is already held), or drop the paragraph from the description and amend the commit message. As it stands the code and the record disagree, which is the kind of thing that misleads whoever reads this commit next.

### Notes (non-blocking)

- **`break` on a swallowed exception is a behaviour change.** `WalkStepAsync`'s catch-all returns `null` after `Debug.Fail`, which is a no-op in Release. Previously a transient failure there was retried; now it permanently ends the walk, silently. Probably still the better default, but it may be worth distinguishing "no more steps" from "an exception happened" rather than treating both as terminal.
- **Tests are wall-clock based.** 500 ms + 200 ms per test, four tests, plus 8×50 walk restarts in the concurrent one. The 10-reads-vs-tens-of-thousands margin makes false failures unlikely, but the 500 ms drain assumption is the fragile part on a loaded CI agent: a leftover loop is still caught, a merely *slow* one could fail spuriously.
- **Minor:** `Dispose()` still reads `_walkCts.IsCancellationRequested` unsynchronized while `StartWalkAsync` may be disposing that same instance. Harmless today (`IsCancellationRequested` does not throw after disposal), but it is now a genuinely concurrent read that it wasn't before.

Nothing else stood out — the lock discipline documented on `StopCurrentWalkAsync` ("the caller must hold the writer lock") holds at both call sites, and the XML docs explain the *why* rather than restating the code.

---

PR #914, opened 2026-08-29
Fix player-slot leak when a player is disposed without a disconnect

### LINKED ISSUE #912: A player disposed without a disconnect stays in the player list forever and fills the server's max player count
## What happens

With the bot feature enabled, a game server's 500 player slots filled within minutes and real
clients were turned away. `PlayerCount` kept climbing while the actual population did not: the
list was full of players that no longer existed.

## Root cause

A player is added to the game context's player list **before** its initialization can still
fail. `AddPlayerAsync` is the first step of `OfflinePlayer.SetupCharacterAsync`, and a remote
player is added at connect time. If initialization then throws — `OnPlayerEnteredWorldAsync`
failing on unreadable character data, for instance — the failure path disposes the player
without ever removing it from that list:

- `BotManager.RemoveAndDisposeAsync` only disposes.
- `OfflinePlayerManager.RemoveAndDisposeAsync` (the `/offlevel` path) has the same shape.
  Its `StopAsync` happy path *does* call `RemovePlayerAsync` explicitly, so the cleanup is
  correct there and missing on both failure paths.

And it cannot be repaired afterwards: `Player.DisposeAsyncCore` sets `PlayerDisconnected = null`,
so the one event that would have removed the player is gone. The corpse stays in
`GameContext._playerList` — which is exactly what `PlayerCount` and the max-players gate read —
for the lifetime of the process, holding its whole object graph alive with it.

The bot feature turns a slow leak into a fast one, because `RespawnPendingAsync` and
`RotatePresenceAsync` retry a permanently failing spawn every maintenance pass, forever. But
nothing here is bot-specific: any login that fails after the player was added leaks a slot.

## Three related teardown defects found alongside it

**1. A throwing `InternalDisconnectAsync` strands the player just as permanently.**
`DisconnectAsync` lets the exception escape, so the `PlayerDisconnected` subscribers below it
never run — and the state machine has already left the state the method needs to advance from,
so no later call can raise the event again.

**2. The `PlayerDisconnected` subscribers are invoked as a multicast delegate.**
Awaiting a multicast `ValueTask` delegate awaits only the **last** subscriber's task. The
earlier ones — including the removal from the player list — run unobserved and can still be in
flight while the last one (which disposes a remote player) executes. A synchronously throwing
subscriber also aborts the rest of the chain.

**3. A `RemotePlayer` can end up disposed behind a live socket.**
Its connection cleanup sits after the base teardown, so a throw leaves a disposed player whose
socket is still open with the packet handler still subscribed, dispatching client packets into
it until the client gives up. Once the player state has advanced to `Finished`, the connection's
own `Disconnected` → `DisconnectAsync` callback can no longer run the teardown either.

**4. `GameContext.RemovePlayerAsync` fire-and-forgets the map removal.**
`player.CurrentMap?.RemoveAsync(player);` — the `ValueTask` is discarded, so the removal races
the player's own disposal on these new dispose-without-disconnect paths, with any exception lost.

## Reproduction

Three cases, all of which fail against current `master`:

1. A player disposed without ever being disconnected still counts towards `PlayerCount`.
2. A player whose teardown throws is never removed from the game context.
3. A throwing `PlayerDisconnected` subscriber prevents the later subscribers from running.

I have a fix and will open a PR referencing this issue.

facuc28:
Fixes #912

A player is added to the game context's player list before its initialization can still fail
(`AddPlayerAsync` is the first step of `OfflinePlayer.SetupCharacterAsync`, and a remote player
is added at connect time). The failure paths of `BotManager` and `OfflinePlayerManager` disposed
such a player without ever removing it from that list — and `Player.DisposeAsyncCore` drops the
`PlayerDisconnected` handler, so nothing could remove it afterwards. Each of these ghosts
permanently occupied a slot of the server's maximum player count, which reads that very list.
With the bot feature retrying failed spawns every maintenance pass, a population whose
characters could not be loaded filled a 500-slot server within minutes and locked real clients
out.

## The fix

**The invariant is enforced at the bottom.** `Player.DisposeAsyncCore` removes the player from
the game context as its first, guarded step. It is idempotent, so it is a no-op on the regular
event-driven path, and the explicit removals this makes redundant are not duplicated at the call
sites. `BotManager` and `OfflinePlayerManager` only gain doc comments recording the new
invariant.

**The disconnect path gets the same resilience**, since a teardown that throws half way used to
strand the player just as permanently — the state machine has already advanced, so no retry can
ever raise the disconnect event again:

- `DisconnectAsync` catches a throwing `InternalDisconnectAsync` and continues. It then attempts
  the progress save the throw skipped: saving is the last step of `RemoveFromGameAsync`, so
  without it the swallow would turn the failure into silent data loss — the caller sees a
  successful logout and the next session loads the character rolled back to its last periodic
  save.
- The `PlayerDisconnected` subscribers are invoked one by one. Awaiting the multicast delegate
  awaited only the last subscriber, leaving the game-context removal unobserved and racing the
  dispose; isolating each subscriber also keeps one failing handler from skipping the others,
  which matters because the handler list is already dropped and nothing could re-run them.
- `RemotePlayer` tears its connection down on disposal as well. Its only cleanup used to sit
  after the base teardown, so a swallowed throw left a disposed player behind a live socket
  whose packets kept dispatching into it, with the `Finished` state blocking the socket-drop
  retry forever.
- `GameContext.RemovePlayerAsync` awaits the map removal it used to fire-and-forget, guarded so
  the method stays exception-free for the teardown paths that call it right before the dispose.

## Tests

`tests/MUnique.OpenMU.Tests/PlayerRemovalOnDisposeTests.cs` is new and covers all three
behaviours. All three fail against `master` and pass with the fix:

- a dispose without a disconnect leaves no player behind;
- a throwing teardown still removes the player;
- a throwing disconnect subscriber does not skip the later ones.

Full `MUnique.OpenMU.Tests` suite passes (790/790).

## Soak time

Running on a private server with an active bot population since 2026-08-24. `PlayerCount` has
tracked the real population since.

### DIFF
--- a/src/GameLogic/Bots/BotManager.cs
+++ b/src/GameLogic/Bots/BotManager.cs
@@ -289,8 +289,10 @@ public sealed class BotManager
     /// Stops the bot like a regular logout (which saves its progress) and releases its resources.
     /// A stopped-but-not-disposed player keeps its persistence context - and with it the whole tracked
     /// account graph - alive; with the presence rotation stopping bots around the clock, that adds up
-    /// to a leak. Mirrors the teardown of the <see cref="Offline.OfflinePlayerManager"/>; the removal
-    /// from the game context happens through the disconnect event.
+    /// to a leak. Mirrors the teardown of the <see cref="Offline.OfflinePlayerManager"/>: the removal
+    /// from the game context normally happens through the disconnect event, and when the stop threw
+    /// before it could raise that event, the disposal removes the bot as its own safety net (see
+    /// <see cref="Player.DisposeAsyncCore"/>).
     /// </summary>
     private static async ValueTask StopAndDisposeAsync(BotPlayer bot, string key, string reason)
     {
@@ -308,6 +310,14 @@ public sealed class BotManager
         }
     }
 
+    /// <summary>
+    /// Drops a bot which never got as far as playing. It was already added to the game context if its
+    /// initialization failed after <see cref="Player.SetSelectedCharacterAsync"/> was reached, and it is
+    /// never disconnected on this path - the disposal takes it out of the player list (see
+    /// <see cref="Player.DisposeAsyncCore"/>).
+    /// </summary>
+    /// <param name="key">The key of the bot.</param>
+    /// <param name="bot">The bot to drop.</param>
     private async ValueTask RemoveAndDisposeAsync(string key, BotPlayer bot)
     {
         this._bots.TryRemove(key, out _);
--- a/src/GameLogic/GameContext.cs
+++ b/src/GameLogic/GameContext.cs
@@ -363,7 +363,21 @@ public class GameContext : AsyncDisposable, IGameContext
             this.PlayersByCharacterName.TryRemove(player.SelectedCharacter.Name, out _);
         }
 
-        player.CurrentMap?.RemoveAsync(player);
+        if (player.CurrentMap is { } currentMap)
+        {
+            // Awaited and guarded: discarding the ValueTask left the removal running in the
+            // background - racing the player's own disposal on the dispose-without-disconnect
+            // paths - with any exception lost, while this method has to stay exception-free for
+            // the teardown paths which call it right before the dispose.
+            try
+            {
+                await currentMap.RemoveAsync(player).ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                player.Logger.LogError(ex, "Error while removing player {Player} from its map.", player);
+            }
+        }
 
         player.PlayerDisconnected -= this.RemovePlayerAsync;
         player.PlayerEnteredWorld -= this.PlayerEnteredWorldAsync;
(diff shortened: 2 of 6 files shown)


sven-n:
## Code review

Reviewed the changed non-test files: `BotManager.cs` and `OfflinePlayerManager.cs` (comment-only changes), `GameContext.RemovePlayerAsync` (awaited + guarded map removal), `Player.DisconnectAsync`/`DisposeAsyncCore` (teardown continues on failure, per-subscriber invocation, removal from context on dispose), and `RemotePlayer.DisposeAsyncCore` (connection close/dispose as a last resort).

**No correctness issues found.**

The one assumption the change depends on checks out: `GameContext.RemovePlayerAsync` early-returns when `_playerList.Remove` returns `false`, so the new call from `DisposeAsyncCore` is idempotent and cannot double-decrement `PlayerCounter`. Subscriber isolation in the disconnect notification and the save-on-failure fallback also look safe as written.

---

PR #924, opened 2026-08-31
Fix bot party follow across multi-floor maps

chuanhd:
## What was wrong

Some maps, such as Dungeon and Lost Tower, contain several disconnected floors inside the same runtime map. When the party leader moved to another floor, a bot follower still considered the leader to be on the same map. It kept trying an impossible walking route and never considered using a warp gate to regroup.

There was a related problem when tracking whether the leader had settled after changing maps: map numbers are not unique enough to distinguish every runtime map instance.

## What changed

The follow logic now treats these as two separate cases:

- When the leader is on another map, the bot keeps the existing settle delay and follows through the normal legal warp flow.
- When the leader is on the same map, the bot tries to walk first. If no path exists, it looks for a legal gate whose landing area can reach the leader.

To choose that gate, the bot runs one flood fill from the leader across the walk map. This tells us which gate landing points are reachable and how far they are from the leader without running a separate A* search for every point of every gate. Among the usable gates, the bot chooses the one with the shortest worst-case landing distance.

Gate placement is random and `WarpToAsync` does not retry a bad landing point. For that reason, a gate is accepted only when every point that random placement may choose is walkable and can reach the leader. This is intentionally conservative: it avoids making follow behavior succeed or fail by chance. The gate containing the bot's current position is also skipped, so the bot cannot repeatedly warp back to the floor where it is already standing.

If no suitable gate is found, the bot waits 20 seconds before doing another gate search. This backoff is separate from the normal warp cooldown: a failed search should not prevent unrelated warp behavior. The failed follow attempt also leaves the tick available for normal local hunting instead of making the bot idle.

Leader-settle and follow-warp timing now use `TimeProvider`, which makes the cooldown behavior deterministic in server-side tests.

## Compatibility notes

- Warp lookup prefers an exact map-definition id when the warp list contains one. If it does not, it falls back to the previous same-map-number behavior so runtime or instanced map definitions still work.
- The identity tier is selected before checking level requirements. This prevents an inaccessible exact entry from falling through to an easier entry belonging to another same-number map definition.
- Map access still uses the class-adjusted move requirement and the character's plain level, just like player warps. The reset-aware combat level is not used to bypass access requirements.
- Exit-gate placement is now shared through `ExitGateExtensions`, including support for single-coordinate gates and validation of reversed bounds. For normal multi-coordinate gates it preserves the existing upper-exclusive random-placement behavior.
- That upper-exclusive rule is specific to placement. Gate containment and terrain painting elsewhere still use inclusive upper bounds; this PR documents the difference rather than changing those unrelated behaviors.

## Tests

The tests exercise the logic directly without a client or UI. They cover:

- leader transitions between runtime maps sharing the same map number;
- walking before warping, and falling back to hunting when regrouping is impossible;
- selecting a reachable gate instead of a closer gate on the wrong region;
- random landing-area safety and avoiding the bot's current gate;
- failed-search backoff and retry after the cooldown;
- exact-definition lookup, same-number fallback, and level-requirement isolation;
- exit-gate placement bounds, including single-point and invalid gates;
- actual map-change events, so a test cannot pass merely because the bot walked.

```text
dotnet test tests/MUnique.OpenMU.Tests/MUnique.OpenMU.Tests.csproj -p:ci=true --no-restore
# Passed: 857

dotnet test tests/MUnique.OpenMU.Tests/MUnique.OpenMU.Tests.csproj -p:ci=true --no-restore --filter 'FullyQualifiedName~ExitGateExtensionsTest'
# Passed: 4
```


sven-n:
## Review

I read through `src/GameLogic/Bots/BotNavigator.cs` and the new test. The direction is good — both problems described (map-number collisions in the settle tracker, and followers stuck on a disconnected region of the same map) are real, and the fix is small and localized. A few things worth addressing before merge.

### Correctness / behaviour

**1. `TryGetNearestLegalWarp` can pick the warp the bot is already standing on.** In the same-map branch the nearest legal warp is chosen purely by `GateDistance(gate, leader.Position)`. Nothing checks that the chosen gate is on a different region than the bot, or that it actually gets the bot closer. If the leader sits on an isolated pocket whose closest legal warp entry happens to be the bot's own floor gate, the bot warps to where it already is, clears `_travelPath`, and repeats every 20s — the exact "re-issuing an impossible path every tick" symptom, just at cooldown pace. Worth guarding, e.g. skip a candidate whose gate is reachable from the bot's current position (or simply closer to the bot than to the leader).

**2. `GateDistance` is Manhattan distance on the gate's centre point.** On maps whose floors are disconnected regions, coordinate proximity does not imply region membership — two gates that lead to different floors can be spatially adjacent. It's a reasonable heuristic, but please say so in the comment rather than implying it finds the leader's floor; and it makes point 1 more likely, not less.

**3. Dead-ending the tick when no legal warp exists.** Both the cross-map branch and the new same-map branch `return true` (tick consumed) even when nothing happened — no warp, no walk. The bot then neither follows nor hunts until the stuck watchdog fires. In the new branch specifically, returning `false` when `TravelTowardAsync` failed *and* no warp is available would let the bot hunt where it stands instead of idling. This is partly pre-existing, but the new path adds another way to reach it.

**4. `_hasDestination` is left set to an unreachable destination.** The branch sets `_destination`/`_hasDestination = true` before attempting travel; if travel fails and the warp is on cooldown, that stale unreachable destination survives the tick. The success path clears it, the cooldown path doesn't.

### Consistency

**5. Two clocks in one class.** `HasLeaderSettled` now uses `_timeProvider`, while `_lastWarpUtc`, `_nextShoppingCheckUtc`, `_resetDueAtUtc` and the rest still use `DateTime.UtcNow` directly. That's fine as an incremental step, but it means the new warp behaviour can't be tested deterministically (test 2 relies on `_lastWarpUtc` starting at `MinValue`). Either route the cooldown checks through `_timeProvider` too, or note in the ctor doc that it currently only covers the settle timer.

**6. `TryGetLegalWarp` semantics changed silently.** It went from `w.Gate?.Map?.Number == mapDefinition.Number` to `IsSameMapDefinition`, which — when both ids are non-empty — requires id equality. Anywhere a runtime `GameMap.Definition` is not the same persisted entity as the one referenced from `Configuration.WarpList` (e.g. instanced maps), warps that previously matched by number will now stop matching, and the bot will conclude it "cannot legally follow" and kick itself out of the party. Please confirm both sides always come from the same config graph, and consider mentioning this tightening in the PR description since it affects `TryGetLegalWarp`'s existing callers, not just the new one.

**7. `GetLegalWarps` materialises with `.ToList()`** on every call, including the per-tick `TryGetLegalWarp` path that only needs the first match. Returning the lazy `IEnumerable` keeps the old short-circuit.

### Tests

**8. `LeaderSettleTrackingDistinguishesMapsWithSameNumberAsync` doesn't test what its name says.** `_leaderMapId` is compared against `GameMap.Id`, which is a fresh `Guid.NewGuid()` per `GameMap` *instance* (`GameMap.cs:43`) — not the definition id. The assertion therefore holds for any two distinct `GameMap` instances; the differing `Number`/`Discriminator` on the definitions is incidental. It does still catch the old number-based bug, so it's a valid regression test, but the name and the doc comment overstate it. Either rename to reflect map-instance identity, or add a case with two instances of the *same* definition to pin down the intended semantics.

**9. Test 2 asserts the outcome only via position.** `Is.InRange(200, 201)` would also pass if the bot had walked there. Asserting that `_travelPath` was cleared, or that the position changed in a single tick (walking couldn't cover that distance), would make the "warped, not walked" claim explicit.

### Nits

- The `HasLeaderSettled` doc says "Tracks the leader's map instance per bot" — good, that now matches the implementation; the `internal` visibility bump on it and `TryFollowLeaderAsync` is test-only, which is fine given `InternalsVisibleTo("MUnique.OpenMU.Tests")`, but an `// internal for testing` marker helps future readers.
- `IsSameMapDefinition`'s `Guid.Empty` fallback compares `Number` + `Discriminator`; a one-line comment on why the empty-id case exists (non-persisted definitions in tests?) would help.

Nothing here is a blocker for the core idea — items 1, 3 and 6 are the ones I'd want resolved before merge.

---

sven-n:
## Re-review (head `70219d2`, 8 new commits)

Thanks for the follow-ups — most of my previous points are properly addressed:

- **Self-warp loop** — solved, and solved better than I suggested. `FindBestReachableLegalWarpAsync` requiring a walkable route from the gate's landing area to the leader implicitly excludes gates in the bot's own region (if such a path existed, `TravelTowardAsync` wouldn't have failed), and the `landingPoints.Contains(this._player.Position)` guard covers the degenerate case.
- **Tick dead-ends** — solved. `TryRegroupWithLeaderOnSameMapAsync` now returns `false` when nothing happened, and the split into `TryHandleCrossMapLeaderFollowAsync` / `TryRegroupWithLeaderOnSameMapAsync` with the integration into `EvaluateFollowerHuntingAsync` reads much better than the old single method.
- **`TryGetLegalWarp` tightening** — solved, and the tiered exact-id-then-same-number lookup with the comment about the identity tier being selected *before* the level filter is exactly the right subtlety to call out.
- **Mixed clocks** — solved; `_lastWarpUtc` and `WarpCooldown` now go through `_timeProvider`.
- **Test assertions** — much stronger. `MapChangeRecordingPlugIn` pins "warped, not walked", and the cooldown-then-after-cooldown sequence is a good addition.
- Distance-only gate ranking is gone in favour of real path lengths, so the Manhattan-heuristic objection no longer applies (`GateDistance` survives only as a tie-break).

Remaining items, roughly in order of how much they matter.

### 1. The failure path has no backoff (blocking, in my view)

`_lastWarpUtc` is only updated when a warp is actually performed. When `FindBestReachableLegalWarpAsync` returns `null` — no legal gate, or none whose landing area reaches the leader — the cooldown is never armed, so the **entire scan re-runs on every single tick** for as long as the leader stays unreachable. That's the "re-issuing an impossible path every tick" symptom the PR set out to remove, moved from one A* to N of them. Arming the cooldown on a failed search too (or a separate `_lastFollowWarpSearchUtc`) would fix it.

### 2. Cost of the search

Per invocation the search does, for every legal warp candidate, one full A* **per landing cell** of its gate, each acquiring the shared `TravelPathFinderPool` semaphore — on top of the failed leader-directed A* that got us here. For a 5×5 gate with three candidates that's 75 whole-map searches for one follower on one tick. With a party of followers on a Lost Tower-style map, and combined with item 1, this can plausibly starve the pathfinder pool for every other bot on the server.

Some options, cheapest first:
- Pre-filter candidates by `GateDistance` and only path-check the best few.
- Path once from a representative walkable landing cell, and use the walk map / a flood fill for the rest (region membership is what you actually need — the *worst-case length* is only used for ranking).
- Cache the result per (map, leader region) for the duration of the cooldown.

### 3. "Every landing cell must be walkable and reachable" may reject real gates

`GetWorstLandingPathLengthAsync` returns `null` if *any* cell in the gate rectangle is non-walkable or has no route. Given that `PlaceAtGateAsync` doesn't validate walkability, being conservative is defensible — but a single blocked tile inside an otherwise fine multi-cell gate rectangle disqualifies the gate outright, and real `ExitGate` rectangles in the config do include blocked tiles. Consider "at least one walkable cell, and every walkable cell reaches the target" instead. Either way, please state the chosen rule in the doc comment — right now "complete landing area can reach him" reads as an invariant rather than a policy.

This is also untested: every gate in `BotLeaderFollowMapIdentityTest` is built by `CreateGate(map, x, y)` with `X2 = x + 1`, which under the new exclusive-bound rule is a **single** landing cell. The worst-case-over-cells logic, the partial-blockage rejection, and the ranking between candidates of differing path length are all unexercised.

### 4. The exclusive `X2`/`Y2` contract contradicts the rest of the codebase

`ExitGateExtensionsTest.PossibleLandingPointsUseExclusiveUpperBounds` now asserts exclusive upper bounds as the intended contract, but everywhere else gates are treated as **inclusive**:

- `GameMapTerrain.cs:122` — `for (int x = gate.X1; x <= gate.X2; x++)` (safezone marking)
- `PointExtensions.cs:24` — `point.X <= rectangle.X2`
- `PlayerActions/WarpGateAction.cs:71` — `currentPosition.X <= gate.X2 + inaccuracy`

So the safezone is painted over `[X1, X2]` while placement only ever lands in `[X1, X2)`. The exclusivity is pre-existing in `Rand.NextInt(gate.X1, gate.X2)`, and I'm not asking you to change placement behaviour in this PR — but codifying it in a new named contract (`GetPossibleLandingPoints`) and a test titled "matching the random placement logic" makes the divergence permanent and harder to spot later. At minimum, please note in the XML doc that this is the placement convention and that gate containment elsewhere is inclusive; ideally raise it as a separate issue.

Relatedly, two more call sites still duplicate the raw expression and would benefit from the new helper: `Bots/BotGenerator.cs:408` and `Resets/ResetCharacterAction.cs:205`.

### 5. Smaller things

- `TryHandleCrossMapLeaderFollowAsync` no longer uses its `cancellationToken` (all travel moved to the same-map method). Drop the parameter, or keep it and add a `cancellationToken.ThrowIfCancellationRequested()` — an unused documented parameter invites a warning and confuses the next reader.
- `EvaluateFollowerHuntingAsync` discards the `bool` from `TryRegroupWithLeaderOnSameMapAsync`. If the result only exists for the tests, say so; otherwise the caller should probably act on it.
- `TryRegroupWithLeaderOnSameMapAsync` clears `_hasDestination` up front and leaves it cleared on the "walk failed, warp on cooldown" path. That's a deliberate-looking side effect on a method that reports "nothing happened" — worth a word in the comment.
- The comment removed from the pathfinder block ("the `OperationCanceledException` is expected and handled in `SafeEvaluateAsync`") was useful context; now that `FindPath` also takes the token and can throw from inside the `try`, it's more relevant than before, not less.
- `LeaderSettleTrackingDistinguishesMapsWithSameNumberAsync` still tests `GameMap` **instance** identity rather than same-number/different-definition (`_leaderMapId` compares `GameMap.Id`, a per-instance `Guid.NewGuid()`). Still a valid regression test for the old number-based bug; the name and doc just claim more than it checks.

CI: Codacy is green, the Azure `MUnique.OpenMU` build was still queued when I looked — worth confirming it goes green before merge.

Items 1 and 2 are the ones I'd want resolved; 3 and 4 are judgement calls I'd like to see stated explicitly rather than necessarily changed.

---

sven-n:
## Re-review (head `c06fb57`, 4 new commits)

### ⚠️ The PR no longer merges — please resolve the conflict

`master` has moved on (it is at `4ee8b04` now, the PR still branches off `2a2186c`), and GitHub reports the PR as conflicted. Three commits touched `BotNavigator.cs` in the meantime — `c172d98`, `3190cec`, `47bb894` (the bot party-composition / skill-learning work).

I merged it locally to see how bad it is: **one file, three hunks, all of them "keep both sides"**. Nothing conceptual collides — master added a warp-level helper, this PR rewrote the follow logic, and they happen to sit next to each other. Concretely:

```
git remote add upstream https://github.com/MUnique/OpenMU.git   # if not present
git fetch upstream master
git checkout fix/bot-follow-multifloor-upstream
git merge upstream/master
```

**Hunk 1 — the field block (~line 300).** Master added three fields next to the two this PR replaces. Drop master's `_leaderMapNumber` / `DateTime _leaderOnMapSinceUtc` (this PR replaces them), keep everything else from both sides:

```csharp
private GameConfiguration? _warpFloorConfiguration;   // from master
private CharacterClass? _warpFloorClass;              // from master
private int _minWarpLevel = FallbackMinWarpLevel;     // from master
private Guid _leaderMapId;                            // from this PR
private DateTimeOffset _leaderOnMapSinceUtc = DateTimeOffset.MinValue; // from this PR
```

**Hunk 2 — the guard in `ShouldWarpToAnotherMap` (~line 1460).** This is the one that needs actual attention: master **deleted the `MinWarpLevel` constant** (replaced by `FallbackMinWarpLevel` + `GetMinWarpLevel()`), so taking this PR's side verbatim will not compile. Take one line from each:

```csharp
if ((plainLevel < this.GetMinWarpLevel() && !mustEscape && !mapIsBarren)   // master
    || this._timeProvider.GetUtcNow() - this._lastWarpUtc < WarpCooldown)  // this PR
```

**Hunk 3 — after `TryGetLegalWarp` (~line 1867).** Looks large, but only because both sides inserted a different new method at the same insertion point. Keep both: master's `GetMinWarpLevel()` (with its closing brace — that is the one thing a naive keep-both loses), then this PR's `FindBestReachableLegalWarp` / `GetWorstLandingDistance` / `PopulateWalkableDistancesFrom` / `GetLegalWarps`.

Master's `using MUnique.OpenMU.DataModel.Configuration;` merges cleanly on its own. After resolving, please re-run the full test project — the base also gained `BotSkillHandler.TryLearnSkillsAsync` in the evaluation loop, and the Azure build needs a run against the new base either way.

(Small follow-up thought, not for this PR: master's `GetMinWarpLevel()` and this PR's `GetLegalWarps()` now both walk `Configuration.WarpList` computing `GetEffectiveMoveLevelRequirement`. Worth unifying later.)

---

### On the code itself — the two blocking items are resolved

**Backoff (my item 1) — fixed, and fixed properly.** `_nextFollowWarpSearchUtc` as a *separate* timer from `_lastWarpUtc` is the right call: a failed search now costs 20s of silence without suppressing unrelated warp decisions, and it is correctly cleared on a successful walk, on a cross-map follow warp, and in the stuck reset. Moving the cooldown check above the search so the expensive work is skipped entirely is a nice touch.

**Search cost (my item 2) — fixed, and better than what I suggested.** One reverse BFS from the leader over the walk map, replacing N×M pathfinder acquisitions, is exactly the right shape: movement is bidirectional and unweighted, so a single fill yields both reachability and a ranking distance for every landing point. The `ArrayPool` rentals with the up-front `Array.Clear`, the distance+1 encoding so `0` means unreachable, and the periodic `ThrowIfCancellationRequested` every 256 dequeues are all handled correctly. I checked the 8-neighbour expansion against `BaseGridNetwork.GetPossibleNextNodes` — neither side restricts corner cutting, so the fill and the pathfinder agree on connectivity.

**Items 3 and 4** — the fail-closed landing policy and the upper-bound convention are now explicitly documented as deliberate trade-offs in the `<remarks>` on `FindBestReachableLegalWarp` and `GetRandomPoint`/`GetPossibleLandingPoints`, which is what I asked for. The PR description covers both too. Good.

### One new issue introduced by the flood-fill rewrite

**The fill and the pathfinder disagree about what is reachable, because only one of them has a search limit.**

`PopulateWalkableDistancesFrom` explores the whole connected region — it has no node budget. `TravelTowardAsync` routes through `PathFinder` with `SearchLimit = TravelSearchLimit` (20000 closed nodes, `BotNavigator.cs:129`), and `FindPathInner` returns `null` once that is exceeded (`PathFinder.cs:154`). A 256×256 map has 65536 cells, so a long or mazy — but genuinely connected — route can be reachable per the fill and unroutable per A*.

The failure mode: the fill accepts gate B, the bot warps there, next tick the walk to the leader fails again. Gate B is now excluded (the bot stands on it), so the search can pick gate C, warp again 20s later, and the bot ping-pongs between floors instead of converging. The old `FindTravelPathAsync`-based version could not do this, because acceptance used the same pathfinder that would later do the walking.

The 20s backoff bounds how fast this can spin, so it is not a hot loop — but it is a bot that never arrives. Options:

- Reject a candidate whose `worstCaseDistance` is large enough that A* is likely to exceed its budget (crude, but a threshold well under `TravelSearchLimit` would cover it).
- Remember a gate that was accepted and then failed to produce a walk, and exclude it for a while — this also covers any other fill/pathfinder divergence that shows up later.
- Or raise `TravelSearchLimit` toward the cell count so the pathfinder can actually serve anything the fill accepts.

Worth at least a test: a map where the two floors are connected only by a corridor long enough to blow the search limit.

### Still open from last round (all minor)

- `TryHandleCrossMapLeaderFollowAsync` still takes a `cancellationToken` it never uses — the only occurrence in the method is the parameter itself.
- `EvaluateFollowerHuntingAsync:1346` still discards the `bool` from `TryRegroupWithLeaderOnSameMapAsync`. Fine if it exists only for the tests, but say so.
- The comment about `OperationCanceledException` being expected and handled in `SafeEvaluateAsync` is still gone from the pathfinder block, and is now *more* relevant — both `FindPath` and the new flood fill can throw from inside it.
- `LeaderSettleTrackingDistinguishesMapsWithSameNumberAsync` still measures `GameMap` instance identity rather than same-number/different-definition. Valid regression test, overstated name.

So: resolve the conflict, and I'd like the fill-vs-`SearchLimit` divergence addressed or consciously accepted with a comment. Everything else is polish.

---

PR #925, opened 2026-08-31
Implement Castle Siege warfare machines

### LINKED ISSUE #727: Castle Siege Warfare Machines

Zylkien:
## Summary

Implements Castle Siege warfare machines.

- Adds attacker/defender machine interaction with side and operator validation.
- Handles the Castle Siege machine packet group safely.
- Sends interface, firing, and impact-region client notifications.
- Selects configured target zones, applies delayed area damage, and prevents concurrent firing.
- Adds regression tests for authorization, targeting, cooldown, packet handling, and packet serialization.

Closes #727

## Verification

- Focused Castle Siege NPC/machine tests: 21 passed.
- GameLogic, GameServer, and test assembly compile cleanly.


sven-n:
## Review

### 1. `CastleSiegeMachine`: inject the machine type instead of comparing monster numbers

`CastleSiegeNpcController.CreateNpc` (`src/GameLogic/CastleSiege/CastleSiegeNpcController.cs:545`) already switches on the monster number to select the class, so the type can simply be passed as a constructor parameter — no second number comparison in the constructor, and no new `ArgumentOutOfRangeException` path inside spawning.

`CastleSiegeNpcDefinition.DefaultSide` works as well and is the more data-driven source: the runtime carries `CastleSiegeNpcRuntime.Definition`, and the initializer sets `Attack1` for 221 and `Defense` for 222 (`CastleSiegeInitializer.cs:192-195`). One caveat if you go that route: it must map to a *group*, not be compared for equality — an `Attack2`/`Attack3` player has to be able to operate a machine whose `DefaultSide` is `Attack1`. So `MachineType = definition.DefaultSide == Defense ? Defense : Attack`, and `CanBeUsedBy` keeps the side-group logic it has now.

### 2. `CastleSiegeMachineUseAction`: validate the opened NPC and the player's position

The action checks siege state, alive, same map, `Operator == player`, side and `IsActive`, but nothing ties the request to the open dialog or to proximity. `TalkNpcAction` sets `player.OpenedNpc` and advances to `PlayerState.NpcDialogOpened`, and the talk plug-in sets `LeavesDialogOpen = true` — so `player.OpenedNpc` is available and is the natural check. A distance check (`player.IsInRange(foundMachine.Position, …)`) should be added too: today an operator can walk to the far end of the map, or die and respawn elsewhere, and keep firing, because `Operator` is only reassigned when somebody else talks to the machine.

Related: `Operator` is never cleared on disconnect/logout — only replaced by the takeover condition in the talk plug-in, or cleared in `Dispose`. A machine whose operator quits stays "occupied".

### 3. `CastleSiegeMachineDamageHandlerPlugIn` can be removed entirely

`PacketHandlerPlugInContainer.HandlePacketAsync` (`src/GameServer/MessageHandler/PacketHandlerPlugInContainer.cs:107-113`) returns silently for an unregistered sub-code, so an empty handler adds nothing — it does not even suppress a warning. The body is a no-op by construction (`_ = new WeaponExplosionRequest(packet);`). Removing it means dropping the class, its two `PlugInResources` entries, and the three test references (`NpcRequestHandlersUseExpectedSubcodes`, `NpcRequestHandlersIgnoreTruncatedPacketsAsync`).

The open question underneath: the packet's documented purpose is "the server damages the other catapult". If machines should take splash damage, that belongs in the impact loop of `UseAsync` (machines are attackables on the map) rather than in a client-driven callback — which is the right call, but then it should actually be implemented here.

### 4. `CastleSiegeMachineUseResultPlugIn` always reports success

Both `CastleSiegeMachineUseResultPlugIn` and `CastleSiegeMachineInterfacePlugIn` hardcode `SuccessResult = 1`, and the `Result` byte in `ServerToClientPackets.xml` is a plain byte with no enum. The bigger gap is that on every rejection path `UseAsync` and the talk plug-in just `return` and send *nothing* — the client receives no response at all and its machine UI is left hanging. Suggestion: put the result into the view contracts (`ShowMachineUseResultAsync(bool success, …)`, likewise for the interface), and send failures to the requester only via `InvokeViewPlugInAsync`, not through the `ForEachWorldObserverAsync(…, true)` broadcast used for success.

### Additional findings

- **The 1.5 s impact delay blocks the packet handler.** `CastleSiegeMachineUseHandlerPlugIn` awaits `UseAsync`, which awaits `ImpactDelay` before returning. Packets of a connection are processed sequentially, so the operator's movement and attacks stall for 1.5 s per shot. The impact should be scheduled, letting the handler return promptly.
- **Impact targeting is unfiltered.** `map.GetAttackablesInRange(target, 3).Where(a => a.IsAlive)` hits everything — the operator himself, his own side, and targets inside a safe zone. Compare `AreaSkillAttackAction.GetTargetsInRange` (`:132-137`), which filters `a != player`, `!a.IsAtSafezone()` and `a.IsActive()`.
- **Damage comes from the operator's own attack stats** (`AttackByAsync(player, null, false)`), so a catapult shot equals a normal hit and scales with the player's gear. `CastleSiegeConfiguration` has the zone definitions but no machine damage value — worth deciding whether this should be configurable.
- **Magic numbers**: `ImpactDelay = 1500 ms`, notify radius `6` and damage radius `3` are inline constants without a note on where they come from.
- **Duplicate context resolver**: the handler uses `CastleSiegeHandlerContext.Get`, the talk plug-in uses `CastleSiegeTaxProvider.GetContext` — identical bodies. Pre-existing, but the PR now depends on both.
- **Test gap**: no coverage for rejection paths sending a failure response (because there is none), nor for a fire request from a player who is out of range or has not opened the machine.

---

sven-n:
## Re-review at `2a81538`

Most of the previous round is addressed. Remaining points below.

### Resolved

- Machine type is injected by `CastleSiegeNpcController.CreateNpc`, and the throwing constructor branch is gone.
- `UseAsync` now requires `player.OpenedNpc` to be the machine and checks `OperationRange`; the talk plug-in checks the range too.
- `Operator` is released on dialog close and when the player leaves the siege map (`ClearMachineOperator`).
- The impact no longer blocks packet processing.
- `CastleSiegeMachineDamageHandlerPlugIn` is removed, including its resources and test references.
- Splash targets are filtered via `IsActive()` / `IsAtSafezone()` / self-exclusion, and the ranges have named constants with comments.

### Still open

**Failure is never reported to the client.** Every rejection path in `UseAsync` and `CastleSiegeMachineTalkPlugIn` still returns without sending anything, and both view plug-ins hardcode `SuccessResult = 1` while the packet's `Result` field is a plain byte. A client whose request is rejected (wrong side, out of range, machine busy, invalid zone) gets no response at all and its machine UI hangs. Suggestion unchanged: carry the result in the view contracts (`ShowMachineUseResultAsync(bool success, …)`, likewise for the interface) and send failures to the requester only via `InvokeViewPlugInAsync`, not through the `ForEachWorldObserverAsync(…, true)` broadcast used for success.

### New findings on this commit

**1. Friendly fire on structures.** `IsValidImpactTarget` only exempts players of the operator's own side — every other attackable in the blast is fair game. Since gates, Guardian Statues and machines are attackables on the siege map, a defender's machine firing into the defense zones (e.g. `61,88`–`93,108`, which covers castle ground) damages his own gates and statues, and a machine can destroy a machine of its own side. The player filter should be extended to Castle Siege NPCs, e.g. by comparing the target NPC's side against `machine.MachineType` in the same way.

**2. `PlayerStateChangedAsync` takes the global siege lock on every dialog close, server-wide.** The new `NpcDialogOpened` → `EnteredWorld` branch calls `ClearMachineOperatorAsync`, which acquires `context.ExecutionLock` and then walks the full runtime snapshot — for every player closing any NPC dialog anywhere, not just on the siege map. `ObjectRemovedFromMapAsync` guards on the siege map before doing the same work; the state-change path should have an equivalent guard (or track the operated machine on the player) so that closing a merchant dialog on another map doesn't contend for the siege lock.

### Minor

- `_ = this.ApplyImpactAsync(…)` matches the existing detached-task style in the code base (`AttackableNpcBase.cs:371`), but that call site carries an explicit `// don't wait for completion.` — worth mirroring, since the discarded task is what resets `IsActive`.
- The test lost its assertion that `machine.IsActive` returns to `false` after the impact; it now only waits for `targetHit`. Since `IsActive` is the fire cooldown, that transition is worth keeping covered.
- Machine damage still comes from the operator's own attack stats, so a catapult shot scales with the operator's gear and there is no configurable machine damage. Fine as a follow-up, but worth an explicit decision.

---

sven-n:
## Re-review at `fcbe25e`

All findings from the previous rounds are addressed. One remaining concern and two nits below.

### Resolved

- Failure is now reported: both view contracts carry a `success` flag, the remote views map it to the packet's `Result` byte, and rejected requests are sent to the requester only via `InvokeViewPlugInAsync`. Serialization coverage for the failure packets was added.
- Friendly fire on structures is handled by `IsFriendlyTarget`, using `ICastleSiegeNpc.Runtime.Definition.DefaultSide`. This works out correctly in both directions: an attacker's machine still damages the castle gates and statues (`DefaultSide = Defense`), while machines no longer hit same-side machines or structures — which also preserves the "damages the other catapult" behavior the removed sub-handler was meant to cover.
- The dialog-close path no longer takes the siege lock for unrelated NPCs.
- The detached impact task carries the `// Don't wait for completion` comment, and the `IsActive` release is covered again via `WaitForMachineCooldownAsync`.

### Remaining concern

**The `player.OpenedNpc is CastleSiegeMachine` guard in `PlayerStateChangedAsync` makes operator release depend on subtle ordering.** It happens to hold for the normal path — `CloseNpcDialogAction` awaits `TryAdvanceToAsync(EnteredWorld)` *before* clearing `OpenedNpc`, so the handler still sees the machine. But other transitions to `EnteredWorld` null `OpenedNpc` first (`TalkNpcAction.cs:90`, `GuildMasterAnswerAction`, `EnterMarketPlaceAction`), and an operator who opens some other NPC's dialog and closes that one presents a non-machine `OpenedNpc`. In those cases the machine stays reserved until the player leaves the map or the takeover condition in the talk plug-in kicks in.

Since the map check (`CurrentMap?.Definition.Number == CastleSiegeMapDefinition?.Number`) is what actually removes the server-wide lock contention, dropping the `OpenedNpc` condition and keeping only the map check would be robust without reintroducing the cost — `ClearMachineOperator` is already a no-op for a player who operates nothing. Alternatively, track the operated machine on the player instead of scanning the snapshot.

### Nits

- `failureMachineType` does a second `GetObject(machineId)` lookup outside the lock on every request, including successful ones, purely to fill a field that is only used on the failure path. It could be resolved inside the existing lookup, or only when the request is rejected.
- `CanBeUsedBy` now doubles as the "same side" predicate inside `IsFriendlyTarget`, where the name reads a bit off for a target check. A thin `IsSameSide(side)` helper that `CanBeUsedBy` delegates to would make both call sites read naturally.

Note: I reviewed by reading only — there is no .NET SDK in this environment, so I did not compile or run the tests.

---

sven-n:
## Re-review at `32d3039`

All three points from the last round are addressed, and nothing new turned up.

- **Operator cleanup**: the `OpenedNpc` condition is dropped, so leaving *any* NPC dialog on the siege map releases the machine. That removes the dependency on whether `OpenedNpc` is cleared before or after the state advance, while the map check keeps the lock off unrelated dialogs. `LeavingAnotherNpcDialogReleasesMachineOperatorAsync` covers exactly the case that previously slipped through.
- **`machineType` for the failure response** is now resolved inside the protected lookup, so there is no second `GetObject` call. The one remaining early return (siege disabled) passes `default`, which is right — there is no context to resolve a machine from at that point.
- **`IsSameSide`** makes the friendly-target check read correctly, with `CanBeUsedBy` delegating to it.

No further findings from me. Optional nit only: `CanBeUsedBy` is now a pure alias of `IsSameSide`; keeping both is fine if the authorization-vs-comparison distinction is intentional at the call sites.

As before, this was a read-only review — no .NET SDK in my environment, so I did not compile or run the tests.

---

PR #926, opened 2026-09-01
Fix duplicate key errors when a character enters the world

sven-n:
Two `System.ArgumentException: An item with the same key has already been added` errors could keep a character - and offline/bot players - from entering the game:

```
Error while processing the message: C1 0E F3 03 45 72 69 63 44 72 61 76 65 6E
System.ArgumentException: An item with the same key has already been added. Key: Nearby party member count
   at MUnique.OpenMU.GameLogic.Attributes.ItemAwareAttributeSystem..ctor(...)
   at MUnique.OpenMU.GameLogic.Player.OnPlayerEnteredWorldAsync()

Failed to initialize offline player for "bot0040".
System.ArgumentException: An item with the same key has already been added. Key: 238
   at MUnique.OpenMU.GameLogic.SkillList..ctor(Player player)
   at MUnique.OpenMU.GameLogic.Player.OnPlayerEnteredWorldAsync()
```

## Cause

The `AttributeSystem` holds exactly one attribute per `AttributeDefinition` and the `SkillList` one entry per skill number, so a character which holds the same stat attribute (or the same learned skill) twice can't enter the game at all.

The duplicated stat attributes come from the *Regenerations Refactor* update: it adds the `Is resting` and `Nearby party member count` stat attributes to every character class without checking whether the class already has them. Since the character classes of a freshly initialized configuration contain both attributes, a configuration which was initialized after they were introduced and then received the update ends up with each of them twice - and every character of such a class gets the attribute twice as well. The same applies to the attribute definitions the update adds to the game configuration.

## Changes

**Don't create the duplicates**
* `RegenerationsRefactorPlugInBase` / `RegenerationsRefactorPlugInSeason6` only add attribute definitions and stat attributes which don't exist yet.
* `CreateCharacterAction` and `BotGenerator` take the class' stat attributes distinctly.
* `AddInitialSkillPlugInBase` doesn't add a skill the character already knows (it is also called for characters which were created outside the game).

**Repair the existing data**
* New `RemoveDuplicateStatAttributes` update (075 / 095d / Season 6) removes the duplicated stat attributes from the character classes.
* `Player.AddMissingStatAttributes` removes stat attributes which the character holds more than once, keeping the one with the highest value, so nothing that was invested into a stat is lost.
* The `SkillList` constructor removes duplicated learned skills, keeping the entry with the highest level.
* Both repairs are logged as a warning.

**Never fail on such data again**
* The attribute system of a character is built from the distinct stat attributes of the character and its account.

**Item skill removal fix**
While looking at the skill list, `RemoveItemSkillAsync` looked its entry up in the *available* skills, which holds the learned entry when the same skill is also learned. As a result a skill granted by two equipped items stayed in the list forever after both items were taken off, and the item skill list leaked entries. The entry is now looked up in the item skills, and the skill is only removed from the list when the last item granting it is gone and it isn't learned.

## Tests

* `SkillListTest.DuplicateLearnedSkillIsRemovedAsync`
* `SkillListTest.LearnedSkillKeptWhenItemSkillRemovedAsync`
* `SkillListTest.ItemSkillRemovedWithLastItemAsync`
* `DuplicateStatAttributeTests.AttributeSystemIsCreatedForDuplicatedStatAttributeAsync`
* `DuplicateStatAttributeTests.DuplicatedStatAttributesAreRemovedWhenEnteringWorldAsync`

`MUnique.OpenMU.Tests` (814 tests) and `MUnique.OpenMU.Persistence.Initialization.Tests` pass.

https://claude.ai/code/session_01LYr3urpQESKCEPXEpBV6rh

---
_Generated by [Claude Code](https://claude.ai/code/session_01LYr3urpQESKCEPXEpBV6rh)_

### DIFF
--- a/src/GameLogic/SkillList.cs
+++ b/src/GameLogic/SkillList.cs
@@ -5,6 +5,7 @@
 namespace MUnique.OpenMU.GameLogic;
 
 using System.ComponentModel;
+using Microsoft.Extensions.Logging;
 using MUnique.OpenMU.AttributeSystem;
 using MUnique.OpenMU.GameLogic.Attributes;
 using MUnique.OpenMU.GameLogic.Views.Character;
@@ -61,7 +62,7 @@ public sealed class SkillList : ISkillList, IDisposable
         this._learnedSkills = this._player.SelectedCharacter.LearnedSkills ?? new List<SkillEntry>();
         this._learnedSkills.Where(entry => entry.Skill is null).ForEach(entry => throw Error.NotInitializedProperty(entry, nameof(entry.Skill)));
 
-        this._availableSkills = this._learnedSkills.ToDictionary(skillEntry => skillEntry.Skill!.Number.ToUnsigned());
+        this._availableSkills = this.CreateAvailableSkills();
         this._itemSkills = new List<SkillEntry>();
         this._player.Inventory.EquippedItems
             .Where(item => item.HasSkill)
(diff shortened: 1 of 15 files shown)


src/GameLogic/SkillList.cs
```diff
@@ -167,7 +209,10 @@ private async ValueTask AddItemSkillAsync(Skill skill)
 
     private async ValueTask AddLearnedSkillAsync(SkillEntry skill)
     {
-        this._availableSkills.Add(skill.Skill!.Number.ToUnsigned(), skill);
+        // A learned skill replaces the entry of an equipped item which grants the same skill: an item
+        // skill is always level 0, while the learned one keeps its own level. The item skill entry stays
+        // in the item skill list, so unequipping the item doesn't take the learned skill away.
+        this._availableSkills[skill.Skill!.Number.ToUnsigned()] = skill;
```
ze-dom, line 216: Can such a thing happen? A learned skill is never the same skill as an item skill 🤔 
Nothing seems wrong with the code, just the reasoning seems faulty.

Issue #728, opened 2026-03-24
Castle Siege Life Stones


sven-n:
## Summary

Implement Life Stones — guild-summoned NPCs that heal nearby allies during Castle Siege battle. Each guild can place one Life Stone, which goes through a creation phase before becoming active.

## Prerequisites

- #724 (Guild Selection) — for join side awareness.
- #725 (NPCs) — for the `CastleSiegeLifeStone` NPC class pattern.
- #722 (State Machine) — for `Start` state and `CastleSiegeContext`.

### 1. Life Stone NPC — **`GameLogic/CastleSiege/NPC/CastleSiegeLifeStone.cs`** (created in Phase 6)

Extends `AttackableNpcBase`. Characteristics:
- Attackable (enemies can destroy it).
- No movement, no attack ability.
- Immune to pushback and Radiance effects.
- Does not drop loot on death.
- Tracks: owning guild ID, join side, creation phase progress.

### 2. Life Stone Intelligence — **`GameLogic/CastleSiege/Intelligence/CastleSiegeLifeStoneIntelligence.cs`** (created in Phase 6)

**Creation Phase (60 seconds, 5 stages):**
- On spawn, the Life Stone is in "creating" state.
- Every 12 seconds, advance to the next visual stage (0→1→2→3→4).
- Send `CastleSiegeLifeStoneState` packet to nearby players at each stage.
- After 60 seconds (stage 4 complete), transition to "active" state.

**Active Phase:**
- Every AI tick: find all players within 3 tiles.
- Filter to same-side players only.
- Heal each player: `+1% of MaxHP` for HP, MP, AG (not shield).
- Use existing `IAttackable` attribute system for healing.

**Death/Destruction:**
- Remove from `CastleSiegeContext` tracking.
- No loot drops (override drop generator behavior or check in `AttackableNpcBase`).

### 3. Summon Life Stone Action — **New file: `GameLogic/CastleSiege/Actions/CastleSiegeSummonLifeStoneAction.cs`**

**Validation:**
- Current state is `Start`.
- Player has a guild and a join side ≠ `None`.
- Player's guild does not already have an active Life Stone (check `CastleSiegeContext`).
- Player is in the Castle Siege map.

**Effect:**
- Create a `CastleSiegeLifeStone` at the player's position.
- Spawn it on the map.
- Track in `CastleSiegeContext` (guild → Life Stone mapping).

### 4. Life Stone Forced Destruction

Life Stones are forcefully killed when:
[...]

PR #929, opened 2026-09-03
Implement Castle Siege Life Stones

Zylkien:
## Summary

Implements Castle Siege Life Stones (#728).

- Adds Life Stone deployment through the existing item-consume flow.
- Builds the Life Stone through five client-visible stages over 60 seconds.
- Restores 1% HP, MP, and AG per tick to nearby allied participants once active.
- Allows only opposing siege participants to attack it; it creates no loot or experience.
- Clears Life Stones when ownership changes, the battle ends, or the object is removed.
- Synchronizes Life Stone build state through the existing B9:01 packet.
- Configures Life Stone NPC attributes for fresh Season 6 setups and upgrades existing databases safely without overwriting customized values.

## Verification

- `TestSeason6DataAsync`
- 108 Castle Siege tests
- Life Stone gameplay and packet serialization tests
- GameServer build


sven-n:
## Review

Nice, well-scoped implementation overall — the guild-keyed dictionary, the "consume only after successful placement" flow, the non-overwriting update plug-in (with a test that asserts idempotency and preserved customization), and the deliberate no-loot/no-exp death path all look right. `SpawnTrigger.ManuallyForEvent` correctly makes `AttackableNpcBase.HitAsync` remove and dispose the stone without respawn, and `ItemConstants.CastleSiegeLifeStone => new(11, 13)` matches `Misc.CreateLifeStone` (`Number = 11`, `Group = 13`), since `ItemIdentifier` is `(Number, Group)`.

A few things I'd like to see addressed, the first two being actual client-visible bugs.

### 1. Players who enter the Life Stone's scope late never see the correct build stage

`BroadcastBuildTimeAsync` only fires on a stage *change*, and `SynchronizeLifeStonesAsync` is called from `CastleSiegePlugIn.SynchronizePlayerAsync`, i.e. when the player enters the siege map — long before (or entirely without) the stone entering their client-side scope.

On the client, `ReceiveBuildTime` (`WSclient.cpp`) resolves the object via `FindCharacterIndex(Key)`, which returns `MAX_CHARACTERS_CLIENT` when the key is unknown; that's the sentinel slot of `CharacterMemoryDump`, so the packet is silently discarded (no corruption, but no effect either). The result: a player who walks into range of an already-finished Life Stone gets the spawn packet with build stage `0` and no further `B9:01` — the stone renders as permanently unbuilt for them (`GMBattleCastle.cpp` gates the completed mesh/effects on `m_byBuildTime >= 5`).

The client already expects the initial stage on the spawn packet: in the create-monster handler it reads

```cpp
BYTE bMyMob     = (Data2->TypeH) & 0x80;
BYTE byBuildTime = (Data2->TypeH & 0x70) >> 4;
auto Type = (EMonsterType)(((WORD)(Data2->TypeH & 0x03) << 8) + Data2->TypeL);
```

so bits `0x70` and `0x80` of `TypeNumber`'s high byte are free (monster 278 = `0x0116`). Encoding `BuildTime` into `0x70` in `NewNpcsInScopePlugIn` would fix this properly, and `B9:01` then only carries updates.

### 2. The owning guild never gets the "own Life Stone" flag

`m_byFriend` is set purely from `TypeH & 0x80` on spawn, and `GMBattleCastle.cpp` only records `g_fLifeStoneLocation` when `c->m_byFriend == 128`. Since the server never sets that bit, the healing feedback (the `BITMAP_PLUS` particle for allies standing in range) never appears — players get healed with no visual indication that the stone is doing anything. Same fix location as above: set `0x80` per observer for players on the owning side.

### 3. Healing rate is coupled to the periodic-task frequency

`TickLifeStonesAsync` runs from `IPeriodicTaskPlugIn`, which `GameContext` fires on a **1-second** timer. So the stone restores 1% HP **and** MP **and** AG *per second* — full heal in 100s, indefinitely, for every ally within range. That's a balance decision that currently falls out of an implementation detail; if the task interval ever changes, the heal rate changes with it. I'd suggest an explicit heal interval + percentage (ideally on `CastleSiegeConfiguration`, like the statue regen upgrades are configured) and driving it off `utcNow` rather than tick count.

Related: the heal radius `3` and the `maximum / 100` divisor are inline magic numbers — the surrounding code prefers named constants (cf. `CastleSiegeMachine.OperationRange`).

Also, is the client expected to learn about the restored HP/MP at all, or is the particle the only feedback? Potions and interval regeneration don't send anything either, so I suspect this is fine — just confirming it's deliberate.

### 4. Placement failure throws out of a packet handler

`CastleSiegeLifeStone.Initialize()` throws `InvalidOperationException` when `MaximumHealth <= 0`, and `CreateLifeStoneAsync` rethrows after cleanup. That exception propagates through `SummonAsync` → `CastleSiegeLifeStoneConsumeHandlerPlugIn.ConsumeItemAsync` → the item-consume packet handler. A misconfigured/unmigrated database then breaks the consume flow for that player rather than just refusing the placement. I'd log and return `null`/`false` instead.

Minor, same area: `DestroyAsync` doesn't call `OnRemoveFromMap()`, whereas the base class's `RemoveFromMapAndDisposeAsync` does.

### 5. `CastleSiegeLifeStoneIntelligence` is pure indirection

`LifeStone.TickAsync` → `intelligence.TickAsync` → casts `Npc` back to `CastleSiegeLifeStone` → `lifeStone.TickCoreAsync`. The round trip adds nothing except forcing `TickCoreAsync` to be `internal`. Either move the tick logic into the intelligence (like the other siege intelligences own their behavior) or drop the class and keep the tick private to the NPC.

### 6. Smaller points

- `CastleSiegeLifeStoneConsumeHandlerPlugIn` reaching for `CastleSiegeTaxProvider.GetContext` is a surprising dependency — a shared context-lookup helper would read better.
- `SummonAsync` doesn't validate the placement spot (safezone, walkability). Should a stone be placeable on top of the crown or inside the castle?
- Nothing removes a Life Stone when its owning guild disbands or leaves the map — only on win, state exit, or object removal. Intended?
- `OnDeathAsync` skipping `base` also skips `IAttackableGotKilledPlugIn`. Presumably intentional, but worth a comment in the code so it doesn't read as an oversight.
- `internal TickCoreAsync` is declared after `protected` members (SA1202) and has no XML doc; `UpdateVersion.cs:567` has trailing whitespace (SA1028).

### 7. Test coverage

The new test is a good end-to-end pass, but a few behaviors that would regress silently aren't covered: `CanBeAttackedBy` (opposing side only — allies and non-participants must not damage it), `KillAllLifeStonesAsync` on the `Start` → next-state transition, the consume handler consuming the item *only* on successful placement, and build-time delivery to a player who arrives after construction finished (which is issue 1 above).

---

sven-n:
## Re-review (`041b907`)

Thanks — I re-read the delta since `eb578c3` against the client code. All seven points from the previous round are addressed, and the two client-visible bugs are genuinely fixed:

- **Build stage on spawn**: `GetNpcTypeNumber` writes `(BuildTime & 0x07) << 12` / `0x8000` into `TypeNumber`, which `AddNpcsToScopeRef` serializes big-endian, so those land exactly on the `TypeH & 0x70` / `& 0x80` bits `WSclient.cpp` reads. The stage now travels with the spawn, and `SynchronizeLifeStonesAsync` / `SynchronizeBuildTimeAsync` are correctly gone — the remaining `B9:01` broadcasts go through `ForEachWorldObserverAsync`, which is scope-correct. `& 0x03FF` for the type stays clear of the flag bits (client masks `& 0x03`).
- **Placement failure** no longer escapes the consume handler, with a regression test that asserts the item survives a misconfigured definition.
- Explicit `HealingInterval`, named `HealingRange`/`HealingPercentage`, `OnRemoveFromMap()` in `DestroyAsync`, the intelligence indirection dropped, `CastleSiegeContextResolver` extracted, terrain/safezone/occupancy validation, and the comment explaining the deliberate `base.OnDeathAsync` skip — all good. The new tests (opposing-side authorization, consume-on-success, heal throttling, late-scope encoding, state-exit cleanup) cover what I was worried about. CI is green.

Two things left, one of which is a real behavioral question:

### 1. The ownership flag is guild-scoped but the healing is side-scoped

`GetNpcTypeNumber` sets `0x8000` only when `lifeStone.OwnerGuildId == observerGuildId`, while `TickAsync` heals every player whose `GetPlayerJoinSide(player) == this.JoinSide` — i.e. all guilds on that side. Members of an *allied* guild therefore get healed while their client never records `g_fLifeStoneLocation` (`GMBattleCastle.cpp` gates that on `m_byFriend == 128`), so they get no `BITMAP_PLUS` feedback and no indication the stone is theirs to stand near.

Worth deciding deliberately rather than leaving the two scopes to disagree. Note the client's particle code additionally gates on `Hero->GuildMarkIndex == c->GuildMarkIndex`, so the aura is per-guild client-side no matter what the server sends — which hints the original heal may have been guild-scoped too. Either narrow the heal to `OwnerGuildId`, or set the flag for the whole side and accept that the aura only renders for guild mates (a code comment would help the next reader).

### 2. `NewNpcsInScopePlugIn` now depends on `GameLogic.CastleSiege.NPC`

The generic spawn path type-checking for one event's NPC will attract more branches over time. A small abstraction on the NPC side — e.g. an interface exposing "extra type-number flags for this observer" that `GetNpcTypeNumber` consults — would keep the view plug-in event-agnostic. Related: `NewNpcsInScopePlugIn075/095` don't carry the encoding, which is fine while Life Stones are Season 6-only, but that's implicit right now.

### Minor

- `catch (Exception ex)` in `CreateLifeStoneAsync` also swallows `OperationCanceledException` during shutdown; an exception filter would keep cancellation propagating.
- Healing is still a hardcoded 1% of HP **and** MP **and** AG per second (full heal in ~100s) — now explicit and testable, which was the important part, but a `CastleSiegeConfiguration` entry would let servers tune it. Fine to leave if it matches original behavior.
- The safezone check could reuse the existing `LocateableExtensions.IsAtSafezone` helper.
- The occupancy guard (`GetNpcsInRange(position, 0)`) only rejects an exact tile collision, so a stone can still be dropped adjacent to the Crown or a statue. Probably fine — just flagging that it's a narrower guard than it reads.

No dotnet toolchain in my environment, so I did not build or run the tests locally; the above is from reading the diff against `master` and the MuMain client handlers.

---

sven-n:
## Re-review (`95f02a4`)

All three points from the last round are handled, and I have nothing blocking left.

- **Scope mismatch resolved**: `GetTypeNumberFlags` now sets the friendly bit for every observer whose `GetTrackedPlayerJoinSide` matches the stone's `JoinSide`, so the flag scope matches the side-scoped healing, and the comment correctly notes MuMain narrows the particle to the observer's own guild mark on top of that. Using the *tracked* side rather than `GetPlayerJoinSide` is the right call here — the map check would be redundant (the packet only goes to observers on the siege map) and racy during map entry.
- **Layering resolved**: `IObserverDependentNpcTypeNumber` keeps `NewNpcsInScopePlugIn` free of any Castle Siege reference, and the remark makes the 075/095 behavior explicit instead of implicit.
- `OperationCanceledException` is now rethrown, and `IsAtSafezone()` is reused.
- The new `LifeStoneSpawnMarksSameSideObserversAsFriendlyAsync` test pins both the allied and the opposing case, which is exactly the regression that would otherwise slip.

Merged cleanly with master, CI green, mergeable.

One optional nit, no need to hold the PR for it: `IObserverDependentNpcTypeNumber` doesn't say which bits an implementation may set. Since the flags are OR-ed straight onto the type number, a future implementer has no way to know that only `0x7000`/`0x8000` are free (and that the type itself needs the low bits). One sentence on the interface would prevent that.

The heal rate staying a hardcoded 1%/s of HP+MP+AG is fine by me if it matches original behavior — it's explicit and tested now, which was the part that mattered.

LGTM.

---

didiconcs:
I checked out `Zylkien:castle-siege/728-life-stones` (95f02a47e) in an isolated worktree and went through build, tests, and the core logic. Summary below.

## Build

Clean. `Persistence.SourceGenerator` first, then `Startup.csproj` → **0 errors**. The 194 warnings that come out of the full build are all pre-existing (StyleCop/analyzer noise in unrelated files), with one exception noted below.

One real (minor) regression: this PR introduces **two new SA1202 warnings** ("public members should come before internal members") in `CastleSiegeContext.cs` that don't exist on `master` (verified by building the file before/after the PR):
- `internal LifeStones` (line 119) sits between two blocks of `public` properties (`SwitchUsers` above, `CrownAccumulatedTime`/`IsCrownAvailable` below).
- `KillAllLifeStonesAsync` / `TickLifeStonesAsync` / `RemoveLifeStone` (lines 342, 376) sit directly before `public async ValueTask LoadAsync()` (line 381).

`CreateLifeStoneAsync` is placed correctly (after `SetPlayerJoinSideAsync`, a public method), so it's just these two spots. Doesn't affect behavior, just style, easy fix by moving those members down next to the other internal members further in the file (e.g. near `TrackPlayer`/`UntrackPlayer`).

## Tests

All green, no failures:
- `MUnique.OpenMU.Tests` filtered on `CastleSiege`: **113 passed**, 0 failed (includes 6 new Life Stone tests: `LifeStoneBuildsHealsAndEncodesLateScopeStateAsync`, `LifeStoneSpawnMarksSameSideObserversAsFriendlyAsync`, `LifeStoneAttackRequiresOpposingSiegeSideAsync`, `LifeStoneItemIsConsumedOnlyAfterSuccessfulPlacementAsync`, `LifeStoneMisconfigurationDoesNotEscapeConsumeHandlerAsync`, `LeavingBattleStateDestroysLifeStonesAsync`, plus `SerializeLifeStoneBuildTimeAsync` in `CastleSiegeCrownRemoteViewTests.cs`).
- `MUnique.OpenMU.Network.Packets.Tests` (Castle Siege packet layout/round-trip tests): **140 passed**.
- `TestSeason6DataAsync`: **passed**.

Note: the PR body says "108 Castle Siege tests", the actual count I get filtering on the same name is 113 (in `MUnique.OpenMU.Tests` alone), which is just a documentation nit, not a discrepancy that matters since everything is passing either way.

The DB-upgrade test (`TestInitializationWithEfCore`) is a nice touch: it sets a customized `MaximumHealth` and removes `DefenseBase`, runs `ConfigureCastleSiegeLifeStoneUpdatePlugIn` twice, and asserts the customized value survives while the missing attribute gets backfilled, good idempotency coverage for the "upgrades safely without overwriting customized values" claim.

## Logic review

Went through `CastleSiegeLifeStone`, `CastleSiegeContext`, `CastleSiegeSummonLifeStoneAction`, `CastleSiegeLifeStoneConsumeHandlerPlugIn`, `CastleSiegePlugIn`, `CastleSiegeCrownMechanics`, `NewNpcsInScopePlugIn`, and the migration plug-in against the claims in the description:

- **Build timing**: `CompletedBuildTime = 5`, `BuildStageInterval = 12s` → 60s total, matches the description and is covered by `LifeStoneBuildsHealsAndEncodesLateScopeStateAsync` (asserts stage 1 at t=12s, stage 5/active at t=60s).
- **Healing**: 1%/tick of HP/MP/AG to same-side players in range once active, verified by the same test (heals at t=60s and t=61s, not at t=60.5s, correct 1s tick gating).
- **Attack restriction**: `CanBeAttackedBy` requires the attacker to be a player (or player surrogate) currently on the siege map with an assigned side different from the stone's `JoinSide`. Owner and non-participants get rejected, opposing side gets through, exactly matches `LifeStoneAttackRequiresOpposingSiegeSideAsync`.
- **No loot/exp**: `OnDeathAsync` is overridden to skip `AttackableNpcBase`'s base implementation entirely (no `DropItemDelayedAsync`, no `AddExpAfterKillAsync`), only sends the kill notification to observers. Map removal still happens correctly afterward because `SpawnTrigger.ManuallyForEvent` doesn't match any of the `ShouldRespawn` triggers, so `AttackableNpcBase.HitAsync`'s generic post-death path calls `RemoveFromMapAndDisposeAsync` for you.
- **Cleanup on ownership change / battle end / object removal**: `KillAllLifeStonesAsync` is called from `ChangeWinnerGuildAsync` (Crown capture) and from `ChangeStateAsync` whenever leaving `CastleSiegeState.Start` (covers both natural and forced state transitions, since both paths route through `ChangeStateAsync`). Individual removal (death, explicit despawn) is handled via `Dispose()` and the `IObjectRemovedFromMapPlugIn` hook, both funneling into the same `RemoveLifeStone`. The three code paths are redundant with each other (a stone removed by combat gets `RemoveLifeStone` called 2, 3 times), but that's harmless: the dictionary is a `ConcurrentDictionary<uint, CastleSiegeLifeStone>`, and `RemoveLifeStone` checks `ReferenceEquals` before removing, so repeats are no-ops.
- **Concurrency**: Life Stone creation (`CreateLifeStoneAsync` via `SummonAsync`) and the periodic tick's `TickLifeStonesAsync`/`KillAllLifeStonesAsync` (via `ChangeWinnerGuildAsync`/`ChangeStateAsync`) all go through `context.ExecutionLock`, consistent with how the rest of `CastleSiegeContext` guards shared mutable state. Combat-triggered removal happens outside that lock (on the attacker's own processing path, not the periodic task), but that's the same pattern every other Castle Siege NPC in this codebase already follows, and it's safe here specifically because `_lifeStones` is a `ConcurrentDictionary` and `TickAsync` bails out early once `IsAlive` is false. I didn't find a race that could double-tick, double-heal, or use a disposed stone.
- **B9:01 packet reuse**: confirmed `docs/Packets/C1-B9-01-CastleSiegeLifeStoneBuildTime_by-server.md` predates this PR (introduced in "Define and validate Castle Siege network packets") and is untouched by this diff. The description's claim that this reuses an existing packet checks out.
- **DB migration**: `ConfigureCastleSiegeLifeStoneUpdatePlugIn` only backfills `MaximumHealth`/`DefenseBase` when missing (`AddAttributes(missingAttributes, ...)` guarded by `Any(...)` checks per attribute), so it won't clobber a server operator's custom tuning.

Nothing else jumped out. The per-guild "one Life Stone at a time" guard (`_lifeStones` keyed by `guildId`, `TryAdd` failing closed) and the safezone/occupied-tile checks in `SummonAsync` look correct and are exercised by tests.

## Verdict

Solid, well-tested phase of the Castle Siege epic. Build is clean, all claimed test suites pass (and then some), and the logic matches the PR description point-for-point, including the trickier bits (no loot/exp on a custom `OnDeathAsync`, side-restricted attackability, cleanup on all three described triggers). Only concrete issue is the two new SA1202 ordering warnings in `CastleSiegeContext.cs`, cosmetic, not a blocker, but worth a quick fixup before merge to keep the warning count at zero.

sven-n:
## Re-review (`da30cc1`)

The only change since my last pass is the `IObserverDependentNpcTypeNumber` remark, which now names `0x7000`/`0x8000` as the reserved masks and states that the lower bits hold the type number — exactly the nit I raised, and it matches what `NewNpcsInScopePlugIn.GetNpcTypeNumber` actually does with the returned flags.

Nothing else changed, CI is green, and the PR is still mergeable. Still LGTM from my side.

---

sven-n:
I'll merge when the conflicts are resolved 👍

Zylkien:
> I'll merge when the conflicts are resolved 👍

yea sorry.. conflicts are resolved

PR #930, opened 2026-09-03
Implement Castle Siege administration panel

### LINKED ISSUE #733: Castle Siege Admin Panel Integration
## Summary

Add Castle Siege management capabilities to the Blazor admin panel, allowing administrators to view event status, force state transitions, manage castle ownership, and edit configuration.

## Prerequisites

- #722 (State Machine) — for `CastleSiegeContext` and state management.
- #721 (Data Model) — for `CastleSiegeConfiguration` type.

## Background

The admin panel is a Blazor Server application in `Web/AdminPanel/`. It uses:
- Generic configuration editors via `AutoFields.cs` in `Web/Shared/Components/Form/`.
- Navigation via `NavMenu.razor` in `Web/AdminPanel/Components/Layout/`.
- Server API controllers in `Web/AdminPanel/API/ServerController.cs`.
- The `IConfigurationChangeMediator` for live configuration updates.

### 1. Castle Siege Management Page — **New file: `Web/AdminPanel/Components/CastleSiegeManagement.razor`**

A Blazor component providing:

#### Status Display
- Current Castle Siege state (name, start time, remaining duration).
- Castle owner guild name (or "Unoccupied").
- Number of registered guilds.
- Tax rates and tribute money balance.

#### Admin Actions
- **Force State Transition**: Dropdown to select a target state, button to force-transition. Calls `CastleSiegePlugIn.ForceChangeState(state)` or equivalent.
- **Change Owner**: Text field for guild name, button to set as castle owner.
- **Reset Siege**: Button to clear all registration data and reset to `Idle1`.
- **Edit Tax Rates**: Inputs for chaos, store, and hunt tax rates with save button.
- **Clear Tribute Money**: Button to reset tribute money to 0.

#### Registered Guilds View
- Table showing all registered guilds: name, marks, registration order.
- Button to remove individual registrations.

#### NPC Status View
- Table showing castle NPCs: type, instance ID, upgrade levels, current HP, alive status.

### 2. Navigation Entry — **Modify: `Web/AdminPanel/Components/Layout/NavMenu.razor`**

Add a "Castle Siege" navigation entry under the Events section:
```razor
<NavLink class="nav-link" href="castle-siege">
    <span class="oi oi-shield" aria-hidden="true"></span> Castle Siege
</NavLink>
```

### 3. API Endpoints — **Modify: `Web/AdminPanel/API/ServerController.cs`** (if needed)

If the Blazor component needs server-side API calls (for force-start, etc.), add endpoints:
- `POST /api/castle-siege/force-state` — force state transition.
- `POST /api/castle-siege/set-owner` — change castle owner.
- `GET /api/castle-siege/status` — get current status.

Alternatively, if the Blazor server component has direct access to the `IGameServerContext` (which it does through DI), API endpoints may not be necessary.

### 4. Configuration Editor

The existing `AutoFields.cs` system should automatically generate form fields for `CastleSiegeConfiguration` properties when editing the `GameConfiguration`. Verify:
- All simple properties (bool, int, byte) render correctly.
- Navigation properties (to `GameMapDefinition`, `ItemDefinition`) render as dropdowns.
- Collections (`StateSchedule`, `NpcDefinitions`, `UpgradeDefinitions`) render as editable lists.
- `CastleSiegeZoneDefinition` renders with X1/Y1/X2/Y2 fields.

If custom rendering is needed, create component builders similar to `LocalizedStringFieldBuilder.cs`.

### 5. Chat Commands for Admin

Optionally add chat commands for in-game admin control:

#### `CastleSiegeForceStateChatCommand` — **Optional new file**
```
/csforcestate <state_name>
```
Forces the Castle Siege to transition to the specified state.

#### `CastleSiegeInfoChatCommand` — **Optional new file**
```
/csinfo
```
Shows current Castle Siege state, owner, and remaining time.

## Files to Create

| File | Description |
|------|-------------|
| `Web/AdminPanel/Components/CastleSiegeManagement.razor` | Admin management page |
| `GameLogic/PlugIns/ChatCommands/CastleSiegeForceStateChatCommand.cs` | Optional admin chat command |
| `GameLogic/PlugIns/ChatCommands/CastleSiegeInfoChatCommand.cs` | Optional info chat command |

## Files to Modify

| File | Changes |
|------|---------|
| `Web/AdminPanel/Components/Layout/NavMenu.razor` | Add Castle Siege nav entry |
| `Web/AdminPanel/API/ServerController.cs` | Add CS endpoints (if needed) |

## Acceptance Criteria

- [ ] Castle Siege management page is accessible from the admin panel navigation.
- [ ] Current state, owner, tax rates, and tribute money are displayed correctly.
- [ ] Admin can force state transitions via the UI.
- [ ] Admin can change castle ownership via the UI.
- [ ] Registered guilds are listed with their mark counts.
- [ ] NPC status is visible (upgrade levels, HP, alive/dead).
- [ ] Configuration editor renders `CastleSiegeConfiguration` properties correctly.
- [ ] Chat commands (if implemented) work for in-game admins.

Zylkien:
Closes #733

## Summary

Adds an administrator-only Castle Siege management page to the Blazor admin panel.

## Included

- Live event status: state, timing, owner, registrations, taxes, and tribute.
- Safe controls to force a state transition, change owner, reset registrations to Idle1, update taxes, clear tribute, and remove registrations.
- Castle NPC status table with instance, upgrades, health, and alive state.
- Context-scoped administrative state requests, so selecting one game server does not affect other in-process servers.
- Offline guild-name resolution for owner changes, including persistence and Dapr wiring.
- Navigation entry under Events.

The page intentionally uses direct in-process game-server access and reports when that is unavailable in a distributed deployment; no broad public admin API endpoints were added.

## Verification

- `dotnet build` for GameLogic and AdminPanel succeeded.
- Focused Castle Siege and Guild Server tests passed: 34/34.


sven-n:
## Review

Overall this is a well-structured change. I like that the runtime state is exposed as immutable snapshots through a game-logic service instead of handing mutable `CastleSiegeContext` state to the panel, that every mutation goes through `context.ExecutionLock`, that guild-name resolution happens *before* the lock is taken, and that the per-context `RequestState` replaces the process-wide `ForceState` for admin-triggered transitions (with a test covering the scoping). The `ApplyOwner`/`BroadcastOwnershipChangeAsync` extraction is a clean refactor rather than a duplication.

A few things I'd like to see addressed before merge.

### 1. `ConfigureAwait(false)` in the Blazor component (blocking)

`src/Web/AdminPanel/Pages/CastleSiegeManagement.razor.cs` uses `ConfigureAwait(false)` on every await, including in `OnInitializedAsync`, `OnServerChangedAsync` and `RunOperationAsync`. Every other page code-behind in the AdminPanel uses `ConfigureAwait(true)` — see `AccountSecurity.razor.cs`, `AdminUsers.razor.cs`, which have zero `ConfigureAwait(false)` calls between them.

This isn't only a convention break: dropping the renderer's `SynchronizationContext` means the continuation that mutates `_snapshot`/`_statusMessage`/`_isWorking` and the implicit re-render after the event handler can run off the `Dispatcher`, which is exactly the "The current thread is not associated with the Dispatcher" class of failure. Please switch these to `ConfigureAwait(true)`. (The awaits inside `CastleSiegeAdministration` and `CastleSiegeManagementService` are fine as `false`.)

### 2. No confirmation on destructive actions (blocking)

"Clear registrations and reset to Idle1", "Clear tribute" and "Remove" registration all fire immediately on click. The project already has `IModalService.ShowQuestionAsync` (`src/Web/Shared/ModalExtensions.cs`), used e.g. by `CreationPanelService` for discard confirmation. These three should go through it.

Related: changing the owner silently wipes the taxes and the tribute, because `ApplyOwner` resets `TaxChaos`/`TaxStore`/`TaxHunt`/`TributeMoney`/`IsHuntZoneEnabled`. That's correct for a real capture, but from the panel it's an invisible side effect of typing a guild name — it should at minimum be stated in the confirmation text or the form hint.

### 3. Hardcoded UI strings

The page adds one resource (`Resources.CastleSiege`) and then hardcodes roughly 35 English strings ("Game server", "Status", "Force state", "Change owner", "Chaos tax (%)", "Registered guilds", "NPC status", "Unoccupied", "Elapsed", …), plus all of the user-facing error and success messages in `CastleSiegeAdministration`/`CastleSiegeManagementService`/`CastleSiegeManagement.razor.cs`. Every other page in `src/Web/AdminPanel/Pages` resolves its labels through `Resources`. These should move into `Resources.resx`.

While you're there: the `.resx` entry was appended at the end of the file but the `Resources.Designer.cs` property was inserted after `AccessDeniedDescription`. Designer files are generated — please regenerate rather than hand-place, so the ordering stays consistent.

### 4. Magic numbers duplicated in the markup

```razor
<input id="castle-siege-chaos-tax" type="number" min="0" max="3" ... />
<input id="castle-siege-hunt-tax"  type="number" min="0" max="300000" ... />
```

The server-side validation uses `CastleSiegeTaxProvider.MaximumPercentageTax` and `CastleSiegeTaxProvider.MaximumHuntTax`. Bind the `max` attributes to those constants so the two can't drift.

### 5. Two force-state paths can both fire in one tick

In `CastleSiegePlugIn.ExecuteTaskAsync` the pre-existing `_forceRequestVersion` block and the new `TryTakeRequestedState` block are independent `if`s. If a global `ForceState(state)` and a per-context request are both pending on the same tick, `ChangeStateAsync` runs twice and the per-context one silently wins, with the first transition's enter-notifications already sent. An `else if`, or folding both into a single "pending transition" concept on the context, would make the ordering explicit.

### 6. `CreateNpcSnapshots` can throw on duplicate keys

```csharp
var runtimes = context.NpcController.GetRuntimeSnapshot()
    .ToDictionary(runtime => (runtime.Definition.MonsterDefinition?.Number, runtime.Definition.InstanceId), ...);
```

The key is nullable on the first component, so any two runtimes whose definitions have no `MonsterDefinition` and share an `InstanceId` make `ToDictionary` throw `ArgumentException` — which surfaces as a dead admin page for a mere status read. A `ToLookup` + `FirstOrDefault`, or the `ToDictionary` overload with a duplicate-tolerant grouping, would keep a read-only view read-only.

### 7. Owner name still can't be resolved for an offline guild

The PR description lists "offline guild-name resolution" as a feature, but that only applies to the *write* path (`GetPersistentGuildIdByNameAsync`). For display, `GetGuildNameAsync` goes `GetGuildIdAsync(Guid)` → `GetGuildAsync(uint)`. That path works, but via `GetOrCreateGuildContainerAsync`, so every page refresh materialises the owner guild's runtime container as a side effect of a read — and the page falls back to printing a raw GUID whenever it can't. Since you're already adding a persistence-level lookup, a symmetric `GetGuildNameByIdAsync` (or reusing the same `IGuildServerContext`) would be both cheaper and more reliable here.

### 8. Dapr client conflates "not found" with "call failed"

`Dapr/ServerClients/GuildServer.GetPersistentGuildIdByNameAsync` catches every exception and returns `null`, which `SetOwnerAsync` reports to the admin as "The guild was not found." An admin debugging a broken guild-server connection will be told their guild doesn't exist. Worth distinguishing, or at least wording the message so it doesn't assert something the code doesn't know.

### Smaller notes

- `CastleSiegeManagement.RefreshAsync` swallows exceptions into `_statusMessage` with no `ILogger` anywhere in the page or the service. Please log them.
- `SetOwnerAsync` broadcasts an ownership change even when `ApplyOwner` returned `false` (same guild re-assigned), and it broadcasts the admin's typed string rather than the guild's canonical stored name — casing differences will show up in the client message.
- `RemoveRegistrationAsync` is restricted to `CastleSiegeState.RegisterGuild`, but marks are submitted during `RegisterMark`. Is excluding that state deliberate?
- `CastleSiegeAdministration.ForceState` re-checks `GetContext(...)` and then returns `Succeeded` unconditionally, even though `_plugIn.ForceState` no-ops if the context vanished in between. Minor, but the result would be lying.
- The "Remaining" column is computed once per refresh, so it goes stale immediately. A periodic refresh (or at least dropping the field in favour of the absolute end time already shown) would avoid confusion.
- Nice touch fixing the missing trailing newlines in `GuildServerContext.cs`, `IGuildServerContext.cs`, `GuildServerInMemoryContext.cs` and `Startup.cs`.

### CI

The Azure build is green; **Codacy Static Code Analysis** is `action_required` — worth a look before merge.

---

sven-n:
## Review

Nice work — the shape is right: a game-logic administration service that owns the locking and validation, a thin admin-panel service, and no new public admin API surface for runtime state. The `IGuildServer.GetPersistentGuildIdByNameAsync` addition is implemented consistently across EF, in-memory, the Dapr client and the controller, and the offline lookup is deliberately done *before* taking `ExecutionLock`, which is the right call.

### Findings

**1. `_isWorking` never renders (`CastleSiegeManagement.razor.cs`)**

`RunOperationAsync` sets `_isWorking = true` and then awaits without a `StateHasChanged()`. Blazor only re-renders after the handler completes, by which point `finally` has already set it back to `false` — so the buttons are never actually disabled and the operations can be double-submitted. Same for `RefreshAsync`. Needs `this.StateHasChanged()` after setting the flag (and the component won't re-render from the background completion otherwise either — consider `InvokeAsync(StateHasChanged)`).

**2. Two state transitions can run in one tick (`CastleSiegePlugIn.ExecuteTaskAsync`)**

The new `TryTakeRequestedState` block sits right after the existing `_forceRequestVersion` block, and both call `ChangeStateAsync` unconditionally. If a global force and a per-context request are pending on the same tick, two transitions are applied back to back — the first one's state-enter notifications fire for a state that is immediately abandoned. Suggest making it a single "pending transition" resolution: take the per-context request first and skip the global force when one was taken (or vice versa), so at most one `ChangeStateAsync` runs per tick.

**3. `SetOwnerAsync` broadcasts the typed name, not the canonical one**

```csharp
await CastleSiegeCrownMechanics.BroadcastOwnershipChangeAsync(context, guildName.Trim())
```

The lookup resolved a guild id, but the broadcast uses the administrator's input verbatim. On a case-insensitive or otherwise non-exact match the clients get a name that doesn't match the stored guild name. Better to resolve the guild's actual name from the id and broadcast that.

**4. `ForceState` reports success it can't guarantee**

`CastleSiegeAdministration.ForceState` null-checks the context, then calls `_plugIn.ForceState(...)`, which internally does `GetContext(...)?.RequestState(state)` — a second lookup that can return null in between, after which `Succeeded` is still returned. Minor, but the whole method could just do one `GetContext` and act on that instance.

**5. Stale registrations become user-visible**

When a guild is disbanded outside the siege flow, its `CastleSiegeGuildRegistration` row is removed by the database, but `CastleSiegeContext.RegisteredGuilds` still holds the in-memory registration. The new snapshot reads runtime state, so the panel will display a registration for a guild that no longer exists, and `RemoveRegistrationAsync` will then no-op on the database (`GetByIdAsync` returns null) while still reporting success. Not introduced by this PR, but the panel turns it from invisible into user-facing — worth either having the guild-deletion path notify the siege context, or having the snapshot drop registrations whose guild no longer resolves.

**6. `ResetCycleAsync` does a lot of I/O under `ExecutionLock`**

Two persistence round-trips plus `NpcController.DespawnAllAsync()` are all inside the lock. Given the state guard already rejects `Ready`/`Start`/`End`, this is probably acceptable, but it's the longest lock hold in the file — worth a comment saying so, or moving the despawn out.

**7. Row-by-row deletes in `ClearRegistrationsAsync`**

The reset button makes this path admin-triggerable: it loads every registration and deletes them one at a time through the ORM. A single bulk delete (`ExecuteDeleteAsync`) would be cheaper. Pre-existing, so out of scope here, but now easier to hit.

**8. Localization inconsistency**

The nav entry and page title go through `Resources`, but every label, alert and error string on the page — and all the `CastleSiegeAdministrationResult.Failed(...)` messages in GameLogic — are hardcoded English. Either is defensible, but picking one would be nicer. Note that `Resources.Designer.cs` is generated; hand-editing it works until someone regenerates.

**9. No auto-refresh**

`FormatRemainingTime` computes against `DateTime.UtcNow` at render time, so "Remaining" freezes until the user clicks Refresh or triggers an operation. A timer-driven refresh (or at least a note in the UI) would help, since the whole point of the page is watching a state machine tick.

### Verification

I reviewed the diff only; I did not build or run the tests locally.


---

sven-n:
## Re-review (`0be30b3`)

Thanks — all nine points from the first round are addressed, and several of the fixes are better than what I suggested. Notes on what I checked, then a few things the new code introduced.

### Confirmed resolved

| # | Finding | Resolution |
|---|---|---|
| 1 | `_isWorking` never rendered | `InvokeAsync(StateHasChanged)` around both `RefreshAsync` and `RunOperationAsync`, plus a reentrancy guard. Correct now. |
| 2 | Two transitions per tick | `if (TryTakeRequestedState) … else if (forceRequestVersion …)`, with the context request consuming the global version. Covered by `AdministrationStateRequestPreventsSecondTransitionInSameTickAsync`, and the `RegisteredGuilds` assertion is a good proxy for "EndCycle did not also run". |
| 3 | Broadcast used the typed name | Now resolves the canonical name via `GetPersistentGuildNameAsync` and broadcasts only when ownership actually changed. |
| 4 | `ForceState` could report false success | Single `GetContext` + `context.RequestState`; the redundant `CastleSiegePlugIn.ForceState(IGameContext, …)` overload is gone. |
| 5 | Stale registrations | Filtered in `ResolveRegistrationsAsync` — see N2 below, I think the filter is the wrong direction. |
| 6 | Lock hold in `ResetCycleAsync` | Comment added explaining it is deliberate. Good enough; the reasoning is sound. |
| 8 | Localization | Thorough — error enum + resource lookup, no hardcoded English left on the page. I diffed every `Resources.X` reference against `Resources.resx` and `Resources.Designer.cs`: no missing or orphaned keys. |
| 9 | No auto-refresh | Countdown row dropped and a manual-refresh hint added. Reasonable call. |

Splitting UI text out into `CastleSiegeAdministrationError` was the right move — GameLogic no longer carries user-facing strings at all.

Also confirming the conditional `SaveOwnerAsync` in `SetOwnerAsync` is correct: `ApplyOwner` returns `false` only when `IsOccupied && OwnerGuildId == ownerGuildId`, i.e. nothing was mutated, so skipping the save is right.

### New findings

**N1. `ResolveRegistrationsAsync` is an N+1 on every snapshot**

One `GetPersistentGuildNameAsync` per registration, and each call in `GuildServer` opens its own persistence context (over Dapr, its own HTTP round trip). The page re-snapshots after *every* administrative operation, so a cycle with many registered guilds pays that fan-out repeatedly. A batch lookup — `Where(g => ids.Contains(g.Id)).Select(g => new { g.Id, g.Name })` behind something like `GetPersistentGuildNamesAsync(IEnumerable<Guid>)` — makes it one query.

**N2. The stale filter now hides registrations instead of surfacing them**

`ResolveRegistrationsAsync` drops any registration whose name doesn't resolve. That removes the phantom rows, but it also means an administrator can never *see* a stale registration, and therefore can't remove it — and the "Registrations" count in the Status card silently disagrees with the runtime `RegisteredGuilds` dictionary the state machine actually uses. Since removal works by guild id and would still succeed, I'd keep the row, fall back to the denormalized `registration.GuildName`, and mark it (e.g. a "guild deleted" badge). That turns the panel into the tool that fixes the inconsistency rather than one that conceals it.

Related: with the Dapr client now rethrowing, a single failed lookup aborts the whole snapshot and the page shows the generic unexpected-error message. Worth deciding explicitly whether one unresolvable guild should cost the entire status view.

**N3. `GuildServer.GetPersistentGuildNameAsync` materializes the aggregate for one string**

`GetByIdAsync<Guild>(guildId)?.Name` loads the whole guild entity, while its sibling `GetPersistentGuildIdByNameAsync` projects in the database. A projecting counterpart on `IGuildServerContext` would be symmetric and would compound nicely with N1.

**N4. The `const` → `static { get; }` change in `CastleSiegeTaxProvider` is a step backwards**

This looks like it was made to satisfy Codacy, but it costs real things: the values are no longer compile-time constants, which is exactly why `huntTax is < 0 or > MaximumHuntTax` had to be rewritten as two comparisons, and it rules out use in patterns, `switch` arms and attributes later. `public const int` is entirely idiomatic here — if the original complaint was about `internal const` visibility, widening to `public const` addresses it without the downgrade.

Also inconsistent: `MaximumPercentageTax` and `MaximumHuntTax` became `public`, `HuntTaxStep` stayed `internal` (and is still only used by `CastleSiegeGuardsmanTalkPlugIn` — the hunt-fee input has no `step` attribute, so either wire it up or leave it alone).

**N5. Empty `else` branch in `ExecuteTaskAsync`**

```csharp
else
{
    // No requested transition is pending; continue with the regular schedule below.
}
```

The `if / else if` chain already falls through to `AdvanceExpiredStatesAsync`. Deleting the branch and keeping the comment above the chain says the same thing without a dead block.

**N6. `AvailableGameServers` as a property**

It runs LINQ over the server list and allocates a new `List<>` on every access. Framework design guidelines put that on the "should be a method" side of the line, which is what it was. Minor, and it's only read once today — but the property form invites callers to touch it in a loop.

**N7. Nit: `Assert.DoesNotThrowAsync` in an async test**

In `AdministrationSnapshotToleratesDuplicateNpcRuntimeKeysAsync`, just `await`ing `GetSnapshotAsync` fails the test on throw anyway, and avoids the sync-over-async assertion inside an `async ValueTask` test.

**N8. Nit: the `RefreshAsync` reentrancy guard silently discards**

`OnServerChangedAsync` returns without doing anything if a refresh is in flight, which would leave the dropdown on server B showing server A's data. The select is `disabled` while working so this is hard to reach in practice — noting it rather than asking for a change.

### Verification

No .NET SDK in this environment, so I could not build or run the tests myself. CI is green on `0be30b3` (Azure Pipelines `MUnique.OpenMU` succeeded, Codacy succeeded), and I reviewed the diff and the surrounding code directly.

---

sven-n:
## Re-review 2 (`ef2be90`)

Most of the second round is addressed, and the batch-lookup work is nicely done — the new `GetPersistentGuildNamesAsync` is plumbed consistently through `IGuildServerContext`, EF (projected), in-memory, `IGuildServer` and the Dapr controller/client, with the singular overload now delegating to it. `GetPersistentGuildNamesAsyncFindsOfflineGuilds` covering both the hit and the miss is exactly the right test.

### Confirmed resolved

* **N1** — one query per snapshot instead of one per registration; owner and registrations resolve through the same batch.
* **N2** — `IsGuildDeleted` on the snapshot, a "Guild deleted" badge in the table, and the row stays removable. `AdministrationSnapshotKeepsDeletedGuildRegistrationAsync` pins the behaviour including the fallback to the denormalized name.
* **N3** — the EF path now projects `new { guild.Id, guild.Name }`; no aggregate materialization.
* **N5** — empty `else` gone, and the `contextStateRequested` flag keeps the precedence semantics identical.
* **N7** — the test awaits the snapshot and asserts on it.

### Still open

**N4 — `CastleSiegeTaxProvider` `const` → `static { get; }`**

Unchanged. Restating briefly since it's the one substantive point not picked up: these are compile-time constants that no longer are, which is what forced `huntTax is < 0 or > MaximumHuntTax` to be split into two comparisons, and `HuntTaxStep` is still `internal` while its two neighbours are `public`. `public const int` gets the accessibility the razor needs without any of that.

**N6 — I think this one over-corrected, and that's on my phrasing**

`AvailableGameServers` moved from a re-evaluating property to a list built once in the constructor. My point was only that a property shouldn't do LINQ work per access; freezing the result changes behaviour. `CastleSiegeManagementService` is registered `AddScoped`, and a Blazor Server circuit is long-lived, so a game server that starts (or stops) after the page's scope was created never appears in — or disappears from — the dropdown, and the admin has to reload the page to notice. Reverting to the original `GetAvailableGameServers()` method is the version I'd take: no per-access surprise, no stale snapshot.

### New

**Empty batch still costs a round trip**

`ResolveGuildNamesAsync` calls `GetPersistentGuildNamesAsync(guildIds)` unconditionally. With an unoccupied castle and no registrations — the common Idle state — that's a query (an HTTP call under Dapr) that can only return an empty dictionary, on every refresh and after every operation. A `guildIds.Count == 0` short-circuit returning `snapshot` covers it.

**Leftover: `SetOwnerAsync` still makes two guild-server round trips**

`GetPersistentGuildIdByNameAsync` followed by `GetPersistentGuildNameAsync` for the id just resolved. Now that the batch plumbing exists, a single "get id and canonical name by name" lookup would fold these into one. Minor, and only on an explicit admin action.

### Merge conflict

The PR is currently un-mergeable against `master`: `src/Web/AdminPanel/Properties/Resources.resx` conflicts. It's purely additive on both sides — `master` gained the network-analyzer keys (`AnalyzeTraffic`, `ArchivedSessions`, `ObserveAccount`, …) in the same region where this branch appends the Castle Siege keys — so merging `master` in and keeping both blocks should resolve it. `Resources.Designer.cs` merges cleanly. Worth doing before the next CI run, since the current head can't be validated against the merge result.

### Verification

Still no .NET SDK available here, so this is diff and code reading only, plus a `git merge-tree` against `origin/master` for the conflict check. Note that CI's last green run was `0be30b3`; I have not seen a completed run for `ef2be90`.

---

didiconcs:
I built this branch in an isolated worktree (`git worktree add` off `zylkien/castle-siege/733-admin-panel`, HEAD `bf216250f`) and reviewed the diff. Summary below.

**Build**

- Source generator (`Persistence.SourceGenerator`) then `src/Startup/MUnique.OpenMU.Startup.csproj`, Debug: clean, 0 errors, 193 warnings, all pre-existing StyleCop/analyzer noise unrelated to this change (verified with a `--no-incremental` rebuild filtered for the new Castle Siege file names, zero warnings attributable to this PR's files).
- Also built `Dapr/GuildServer.Host` and `Dapr/AdminPanel.Host` (the distributed-deployment targets) since the PR touches Dapr wiring: both clean, 0 errors.

**Tests**

Ran the two test classes this PR actually touches (`CastleSiegeStateMachineTests`, `GuildServerTest`) with `dotnet test --filter`: **40/40 passed**, 0 failures. That's a few more than the "34/34" in the description, likely just a difference in filter scope (e.g. a broader `~CastleSiege|~GuildServer` name match pulls in 121 tests including unrelated `BotProgressionTests` cases like `IsBotLearnableSkill_CastleSiegeRoleSkill_ReturnsFalse` that happen to share the substring) rather than any discrepancy in coverage. Not a concern, just noting the number didn't reproduce exactly as stated.

**Authorization**

`CastleSiegeManagement.razor` carries `@attribute [Authorize(Policy = AdminPolicies.Administrator)]`, identical to every other Administrator-tier page in the panel (`AdminUsers`, `ApiKeys`, `LogFiles`, `Plugins`, `Setup`, `Updates`). `CastleSiegeManagementService` is only reachable through that page (DI-injected, scoped, never exposed via a controller/HTTP endpoint), so all six mutating actions (force state, set owner, reset cycle, set taxes, clear tribute, remove registration) sit behind the same auth gate. I didn't find a path that bypasses it. The new Dapr-internal `GuildServer` RPCs (`GetPersistentGuildId/NameAsync`) carry no `[Authorize]` on `GuildServerController`, but that's consistent with every pre-existing method on that controller (server-to-server, protected at the network/sidecar layer, not app-level), not a new hole introduced by this change.

**Context scoping (the "one game server shouldn't affect another" claim)**

This is the part I dug into hardest, and it holds up. `CastleSiegePlugIn` maps `IGameContext -> CastleSiegeContext` via a `ConditionalWeakTable`, and `CastleSiegeContext` now carries an *instance* field `_requestedState` (`Interlocked.Exchange`/`TryTakeRequestedState`) rather than reusing the pre-existing `_forcedState`/`_forceRequestVersion` fields on the plug-in itself, which are legacy globals shared across every context that plug-in instance serves. `CastleSiegeAdministration.ForceState` calls `context.RequestState(...)`, the new per-context path, never the old global one, so admin "force state" from the panel is properly isolated per game server regardless of whether one `CastleSiegePlugIn` instance ends up serving multiple game contexts in-process. `CastleSiegeManagementService.Resolve(gameServerId)` also independently re-resolves the specific `IGameServerContext` and its `CastleSiegePlugIn` per call from `IServerProvider`, so there's no shared mutable selection state across requests either. `ExecuteTaskAsync` was updated so a pending context-specific request consumes the tick and takes precedence over a simultaneous global request without double-applying (`AdministrationStateRequestPreventsSecondTransitionInSameTickAsync` covers exactly this). And there's a dedicated test, `AdministrationStateRequestIsScopedToOneGameContextAsync`, that spins up two fixtures and asserts forcing state on the first leaves the second's `CurrentState` untouched, i.e. the exact property called out in the PR description is asserted directly, not just implied. I didn't find a static/global leak here.

**Offline guild-name resolution + persistence + Dapr wiring**

`GetPersistentGuildIdByNameAsync` / `GetPersistentGuildNameAsync` / `GetPersistentGuildNamesAsync` are threaded consistently through `IGuildServerContext` (both EF Core and in-memory implementations), `IGuildServer`, the in-process `GuildServer`, the Dapr `GuildServerController`, and the Dapr `ServerClients.GuildServer` proxy, all four layers match up and both hosts build clean. `SetOwnerAsync` resolves the guild id/canonical name *before* taking `context.ExecutionLock`, with a comment explaining why (avoid holding the siege state-machine lock during what may be a network round trip to the guild server in a distributed deployment); `CastleSiegeAdministration.GetSnapshotAsync` does the same: the guild-name batch resolution for the registration list happens after the lock is released. That's a real subtlety and it's handled correctly. Registrations for guilds that get deleted out from under the siege aren't dropped: they're flagged `IsGuildDeleted` and remain individually removable (covered by `AdministrationSnapshotKeepsDeletedGuildRegistrationAsync`).

**Reuse over duplication**

`CastleSiegeCrownMechanics.ApplyOwner` was refactored from taking a `CastleSiegeGuildParticipant` to taking a `Guid ownerGuildId` directly, so the new admin `SetOwnerAsync` reuses the exact same ownership-transfer logic (tax/tribute/hunt-zone reset) as the existing capture-flow instead of re-implementing it in parallel, which avoids the two paths silently diverging over time.

**Minor/nit**

- `CastleSiegeAdministration.ForceState` validates `Enum.IsDefined(state)` even though the only current caller (the admin page's `<select>`) can't produce an invalid value, reasonable defense-in-depth for a service method, not an issue.
- I did not find any endpoint or code path where selecting/mutating one game server's Castle Siege state could affect another server's runtime state, and did not find any control reachable without the Administrator policy.

Overall this is a well-scoped, carefully guarded addition, nice awareness of lock/network-call ordering and of reusing the existing ownership-transfer logic instead of forking it. No blocking issues found. Nice work.

sven-n:
## Re-review 3 (`bf21625`) — merging

The only new commit is the `master` merge, which resolves the `Resources.resx` conflict. I checked the resolution rather than trusting it: no resource key was lost from either side (all of master's network-analyzer keys and all 74 Castle Siege keys are present), and the diff against `master` touches only the 29 files this PR owns — nothing from master was reverted in the resolution. CI is green on `bf21625` (the Azure run started ~25 seconds after the merge commit, so it covers this head), and the branch is mergeable.

Two points from the previous round remain open, and I'm merging with them rather than blocking:

* **N4** — `CastleSiegeTaxProvider` limits are still `static { get; }` rather than `const`. Style and future flexibility, no functional defect.
* **N6** — `AvailableGameServers` is still computed once in the constructor, so a game server started or stopped after the page's scope was created won't show up until reload. Minor, and partly my fault for how I worded the original note.

Both are small enough not to hold up the feature; worth a follow-up rather than another round trip here.

Thanks for the thorough iteration on this — three rounds, and every substantive finding came back with a fix plus a test that actually pins the behaviour.

---

PR #934, opened 2026-09-04
Delete the members of a member of an aggregate, too

### LINKED ISSUE #933: Deleting an entity orphans the members of its aggregate which it references itself (item storages, appearance data, ...)
**Describe the bug**

Deleting a character leaves its inventory — and every item lying in it — in the database as unreachable rows. The same happens for an account's vault, and for 26 further one-to-one members of an aggregate.

The cause is the direction of the foreign key. `EfCoreModelGenerator` emits `ON DELETE CASCADE` for every `[MemberOfAggregate]` member:

```csharp
// EfCoreModelGenerator.cs, ~line 317
if (isCollection)
{
    modelBuilder.Entity<T>().HasMany(entity => entity.RawX).WithOne().OnDelete(DeleteBehavior.Cascade);
}
else
{
    modelBuilder.Entity<T>().HasOne(entity => entity.RawX).WithOne().OnDelete(DeleteBehavior.Cascade);
}
```

For a **collection** that is correct: the foreign key is held by the child (`Character."AccountId"`, `Item."ItemStorageId"`), and a cascade runs from the referenced row to the referencing one — deleting the account deletes its characters.

For a **reference** (one-to-one) the foreign key is held by the *owner*:

```
Character."InventoryId" → ItemStorage."Id"   (ON DELETE CASCADE)
Account."VaultId"       → ItemStorage."Id"   (ON DELETE CASCADE)
```

So the cascade we generate today means *"deleting the item storage deletes the character"* — the opposite of what's meant. Deleting the character does nothing to the storage, and no database can express the other direction with a plain foreign key. This affects **28 relationships**, among them `LetterBody.SenderAppearance`, `BattleZoneDefinition.Ground/LeftGoal/RightGoal`, `CharacterClass.ComboDefinition`, `MonsterDefinition.MerchantStore`, `Skill.AreaSkillSettings`, `QuestReward.ItemReward` and the castle siege respawn areas.

**To Reproduce**

1. Create a character, put items into its inventory.
2. Delete the character in the game (`DeleteCharacterAction`, which just calls `DeleteAsync(character)`).
3. `select count(*) from data."ItemStorage"` — the inventory row is still there, and so are its items.

`BotGenerator.DeleteAllBotsAsync` already works around this by loading the whole account graph and deleting the storages itself, but that only covers the bot purge; the normal game path leaks.

**Expected behavior**

Deleting the owner of an aggregate removes the whole aggregate, whichever side holds the foreign key.

**Possible solutions**

1. **Invert the foreign keys** so the child carries the owner id (`ItemStorage."CharacterId"`), and let the generator emit `HasOne(...).WithOne().HasForeignKey<Child>(...)`. This is the modelling-correct fix, but `ItemStorage` serves three owners (`Character.Inventory`, `Account.Vault`, `MonsterDefinition.MerchantStore`), so it needs three nullable owner columns or a split of the table, plus a data migration. It also needs a change in `JsonQueryBuilder.AddNavigation`, which currently bails out on inverse navigations (`"inverse property, no data required"`) and would silently drop inventories and vaults from the loaded account JSON.
2. **Delete triggers, generated from the same annotation.** A generic `plpgsql` trigger function plus one `after delete` trigger per one-to-one member of an aggregate, emitted by `EfCoreModelGenerator` next to the cascades it already generates. Much smaller than (1), and it also covers the *shallow* delete: PostgreSQL fires row triggers for rows deleted by a foreign key cascade, so `delete from data."Account" where "Id" = ...` alone cleans up the characters and each of their inventories, without loading anything.
3. **Handle it in the application layer** (in the persistence layer's delete, not in `BotGenerator`). Works for every provider, but can never cover a shallow or bulk delete without loading the whole graph first.

(2) is implemented in #934. Whatever we pick, a one-time cleanup of the rows which are already orphaned is needed as well — no future cascade or trigger reaches them.

**Additional context — the annotations need a review**

`[MemberOfAggregate]` is not a clean "owned exclusively" list today, and that only stays harmless as long as the delete behaviour is broken in the direction described above. Two members reference a type which is shared:

- `Buff.MagicEffectDefinition` — `MagicEffectDefinition` lives in `GameConfiguration.MagicEffects` and is referenced (unmarked) by `Skill.MagicEffectDef` and `ItemDefinition.ConsumeEffect`. Deleting one buff entry off an NPC would take the effect definition away from a skill and an item.
- `Skill.MasterDefinition` — `MasterSkillDefinition` is in the generator's own `StandaloneTypes` list.

Both are excluded automatically by the implementation in #934 (the target type is in a `GameConfiguration` collection, or in `StandaloneTypes`), so they keep today's behaviour until it's decided whether the annotation or the sharing is wrong.

Two more where I could not tell from the code whether the referenced row is exclusively owned, and which would be good to confirm before the triggers go in:

- `MiniGameChangeEvent.SpawnArea` → `config.MonsterSpawnArea`, which also appears in `GameMapDefinition.MonsterSpawns`.
- `QuestReward.ItemReward` → `data.Item`.

sven-n:
Fixes #933.

`EntityFrameworkContextBase.DeleteAsync` removes the members of an aggregate through `ForEachAggregate`, which was not recursive. So deleting an **account** removed its characters, but not the characters' inventories — and an inventory is referenced BY its character, so its foreign key sits at the character and no delete cascade of the database ever reaches it. The storage row, and every item lying in it, stayed behind as unreachable rows. `BotGenerator.DeleteAllBotsAsync` worked around exactly this by deleting the storages itself.

`ForEachAggregate` now walks the whole aggregate instead of one level.

## What's in here

- **`EntityFrameworkContextBase.ForEachAggregate`** recurses into the members it hands to the action. Members are tracked by *reference* while it walks: the entities compare equal by their id, so a `HashSet<object>` on the default comparer would confuse two members sharing the default id, and a graph which references itself would recurse forever. `DetachInternal` no longer recurses itself — it passed an action which called back into `ForEachAggregate` — so the graph is walked once.
- **The recursion stops at a shared configuration object.** `[MemberOfAggregate]` is not a clean "owned exclusively" list: `Buff.MagicEffectDefinition` marks a row which lives in `GameConfiguration.MagicEffects` and is referenced by `Skill.MagicEffectDef` and `ItemDefinition.ConsumeEffect` too. Walking one level deeper would have taken it away from a skill and an item when a monster definition is deleted. So when deleting, the traversal stops when it reaches an object which a collection of the `GameConfiguration` holds — those are roots of their own — *through the aggregate of a member*. Direct members are untouched by this rule, so nothing which was deleted before this PR is kept now: only the new recursion is constrained. Detaching still walks everything, because a stale configuration has to be detached whole.
- **A migration** removes the rows which are already orphaned, which neither a cascade nor the traversal can ever reach: item storages (inventories and vaults) and appearance data.
- **`BotGenerator`** no longer deletes the storages by hand; the recursion covers them. It keeps the full-graph reload, since `ForEachAggregate` can only walk what is loaded.

## What changed since the first version

The first version added `AFTER DELETE` triggers. The review on this PR showed that they conflict with `ForEachAggregate`, which *already* removes single-object `[MemberOfAggregate]` members: the trigger removed the row first, then EF's own `DELETE` hit 0 rows and threw a `DbUpdateConcurrencyException` — breaking the in-game character deletion, the bot purge and admin-panel deletes of config objects. The cross-schema triggers would also have been denied, since the config role holds only `GRANT SELECT` on `data."ItemStorage"` and `data."Item"`. All of that is reverted; the source generator is untouched again.

That review also corrected the premise I opened this with: `DeleteCharacterAction` did **not** leak the inventory, because `ForEachAggregate` removed it. Only the account → characters → inventory path leaked. #933 is updated accordingly.

## Limitations

- **The cleanup covers the account-data side only.** Because the direct members were already removed, a one-to-one member only leaked at depth 2 or deeper: item storages via account → characters, and appearance data as a safety sweep (`LetterBody` is the only table referencing it). Config-side depth-2 orphans are left — a map's `BattleZone` → `Ground`/`LeftGoal`/`RightGoal` rectangles, a monster's `ItemCraftings` → `SimpleCraftingSettings`, and so on. Enumerating the referencing columns for those by hand and getting one wrong would delete live configuration, which is worse than a few unreachable rows. Easy to add as a follow-up with a second pair of eyes on the list.
- **The annotations still deserve a review**, independently of this PR: `Buff.MagicEffectDefinition` and `Skill.MasterDefinition` mark shared rows. The guard above keeps the recursion away from them, but dropping the annotations is the real fix — it changes the model (the generated cascade goes away and `JsonQueryBuilder` would emit a `$ref` instead of inlining the object), so it wants its own migration. Noted in #933.

## Testing

Not run here: this environment has no .NET SDK and no PostgreSQL, so nothing was built or executed locally, and there is no new automated test — an integration test for this needs a database, so it could only be added in the `[Ignore]`d style of `BackupServiceEfCoreTests` and would give CI no signal. CI is green on the branch (Azure, GitHub Actions and Codacy), which covers compilation and the existing suite.

What would actually settle it, against a real database: delete a character, and an account with a vault and characters, and confirm `data."ItemStorage"` is empty afterwards and no `DbUpdateConcurrencyException` is thrown. Then delete a `MonsterDefinition` which has buffs and confirm its `MagicEffectDefinition` rows survive. Happy to write that up as an `[Ignore]`d test if you want it in the repo.

https://claude.ai/code/session_01Q9QAbEmp9rhnSWqZvkVaeN


src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
```diff
@@ -0,0 +1,186 @@
...
+create trigger ""trg_Skill_DeleteAreaSkillSettings"" after delete on config.""Skill""
+    for each row execute function config.delete_owned_row('AreaSkillSettingsId', 'config', 'AreaSkillSettings');
+
+drop trigger if exists ""trg_Account_DeleteVault"" on data.""Account"";
+create trigger ""trg_Account_DeleteVault"" after delete on data.""Account""
+    for each row execute function data.delete_owned_row('VaultId', 'data', 'ItemStorage');
+
+drop trigger if exists ""trg_Character_DeleteInventory"" on data.""Character"";
+create trigger ""trg_Character_DeleteInventory"" after delete on data.""Character""
+    for each row execute function data.delete_owned_row('InventoryId', 'data', 'ItemStorage');
```
sven-n: **This trigger conflicts with `EntityFrameworkContextBase.DeleteAsync`, so deleting a character in-game will fail.**

`DeleteAsync` → `ForEachAggregate(obj, a => this.Context.Remove(a))` (`EntityFrameworkContextBase.cs:158`) removes single-object `[MemberOfAggregate]` members too, not only collections. So `DeleteCharacterAction` (`DeleteCharacterAction.cs:67`) marks both the `Character` and its `Inventory` as Deleted.

`Character` holds the FK, so EF deletes it first. This trigger then removes the `ItemStorage` (and its items through the cascade). EF's own `DELETE FROM data."ItemStorage" WHERE "Id" = …` then affects 0 rows and throws `DbUpdateConcurrencyException`. The graph stays marked Deleted, so every later save of that player's context fails as well.

This also means the inventory was *not* leaked by `DeleteCharacterAction` before this PR: `ForEachAggregate` already deleted it. The leak was only the non-recursive account → characters → inventory path.

I derived this from the code and EF's delete ordering; I haven't run it against Postgres.

---

src/GameLogic/Bots/BotGenerator.cs
```diff
@@ -406,30 +406,18 @@ private async ValueTask<BotAccountDeleteOutcome> TryDeleteBotAccountAsync(string
...
-                    await context.DeleteAsync(inventory).ConfigureAwait(false);
-                }
-            }
-
-            if (account.Vault is { } vault)
-            {
-                await context.DeleteAsync(vault).ConfigureAwait(false);
-            }
-
             var deleteQueued = await context.DeleteAsync(account).ConfigureAwait(false);
```
sven-n, line 505: **The failure the comment above warns about still happens here.**

`DeleteAsync(account)` goes through `ForEachAggregate`, which calls `Context.Remove(account.Vault)`, because `Vault` is a `[MemberOfAggregate]` single-object member. Every bot account has a vault (created at line 167). `trg_Account_DeleteVault` removes it right after the `Account` row, then EF's DELETE of the tracked vault (plus its tracked items) affects 0 rows → `DbUpdateConcurrencyException`. `TryDeleteBotAccountAsync` returns `Failed` for every account, and the purge never completes.

Separately, the characters' inventories and items that the trigger removes stay tracked as `Unchanged` in this long-lived context. It is reused for the whole purge, so the change tracker (and `DetectChanges`) keeps growing with every purged account. In a player context, any later change to one of those entities would produce an UPDATE on a row that no longer exists.

---

src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
```diff
@@ -0,0 +1,186 @@
...
+create trigger ""trg_PowerUpDefinition_DeleteBoost"" after delete on config.""PowerUpDefinition""
+    for each row execute function config.delete_owned_row('BoostId', 'config', 'PowerUpDefinitionValue');
+
+drop trigger if exists ""trg_QuestReward_DeleteItemReward"" on config.""QuestReward"";
+create trigger ""trg_QuestReward_DeleteItemReward"" after delete on config.""QuestReward""
+    for each row execute function config.delete_owned_row('ItemRewardId', 'data', 'Item');
+
+drop trigger if exists ""trg_Skill_DeleteAreaSkillSettings"" on config.""Skill"";
+create trigger ""trg_Skill_DeleteAreaSkillSettings"" after delete on config.""Skill""
+    for each row execute function config.delete_owned_row('AreaSkillSettingsId', 'config', 'AreaSkillSettings');
```
sven-n: **The same conflict breaks admin-panel deletes of config objects that own a one-to-one member.**

`EditConfigGrid.razor.cs:212` → `DeleteAsync` removes the member explicitly, and the trigger has already deleted it. Examples: a `Skill` with `AreaSkillSettings` (this trigger), a `MonsterDefinition` with a `MerchantStore`, a `MagicEffectDefinition` with `Duration`/`Chance`, a `CharacterClass` with a `ComboDefinition`.

`SaveChangesAsync` throws `DbUpdateConcurrencyException`, and the admin sees "Couldn't delete …, probably because it's referenced", even though nothing references it.

---

src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
```diff
@@ -0,0 +1,186 @@
...
+create trigger ""trg_MagicEffectDefinition_DeleteDurationPvp"" after delete on config.""MagicEffectDefinition""
+    for each row execute function config.delete_owned_row('DurationPvpId', 'config', 'PowerUpDefinitionValue');
+
+drop trigger if exists ""trg_MiniGameChangeEvent_DeleteSpawnArea"" on config.""MiniGameChangeEvent"";
+create trigger ""trg_MiniGameChangeEvent_DeleteSpawnArea"" after delete on config.""MiniGameChangeEvent""
+    for each row execute function config.delete_owned_row('SpawnAreaId', 'config', 'MonsterSpawnArea');
+
+drop trigger if exists ""trg_MonsterDefinition_DeleteMerchantStore"" on config.""MonsterDefinition"";
+create trigger ""trg_MonsterDefinition_DeleteMerchantStore"" after delete on config.""MonsterDefinition""
+    for each row execute function config.delete_owned_row('MerchantStoreId', 'data', 'ItemStorage');
```
sven-n: **Permission problem: this config trigger deletes from the data schema.**

The trigger (and `trg_QuestReward_DeleteItemReward` at line 150, which deletes from `data."Item"`) runs with the caller's privileges. `ConfigurationContext` connects as `config`, and that role only has SELECT on `data."ItemStorage"` / `data."Item"` (see `MyNpgsqlMigrationsSqlGenerator.DataTablesRequiredByConfigRole`).

So deleting a merchant `MonsterDefinition` or an item `QuestReward` through the config context fails with `permission denied for table ItemStorage` / `Item`, and the whole delete rolls back. Before this PR it worked. Options: grant DELETE to `config` on those tables, or make the function `security definer`. The second needs care with `search_path`.

---

src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
```diff
@@ -0,0 +1,186 @@
...
+    for each row execute function data.delete_owned_row('SenderAppearanceId', 'data', 'AppearanceData');
+");
+
+            // Item storages which were orphaned before the triggers existed - their items are removed
+            // by the cascade of the storage. These are the only three references to an item storage.
+            migrationBuilder.Sql(
+                @"delete from data.""ItemStorage"" s
+                where not exists (select 1 from data.""Character"" c where c.""InventoryId"" = s.""Id"")
+                  and not exists (select 1 from data.""Account"" a where a.""VaultId"" = s.""Id"")
+                  and not exists (select 1 from config.""MonsterDefinition"" m where m.""MerchantStoreId"" = s.""Id"")");
```
sven-n: **The orphan cleanup only covers `ItemStorage`.** The trigger list above shows 23 other one-to-one relationships that leaked the same way before the triggers existed, e.g. `LetterBody.SenderAppearance` → `AppearanceData` for every deleted letter, plus `PowerUpDefinition`, `PowerUpDefinitionValue`, `Rectangle`, etc. Existing databases keep those orphans (and their cascaded children) forever. Either extend the cleanup to all of them, or note in the description that only item storages are cleaned up.

---

src/Persistence/SourceGenerator/EfCoreModelGenerator.cs
```diff
@@ -181,6 +186,35 @@ private static bool IsStandaloneType(string typeName, Type referencingType)
...
+            "as $$",
+            "declare",
+            "    owned_id uuid;",
+            "begin",
+            "    -- The foreign key of a one-to-one member of an aggregate is held by the owner, so the",
+            "    -- database can't cascade into it. tg_argv holds the key column of the deleted owner row",
+            "    -- and the schema and table of the owned row.",
+            "    owned_id := (to_jsonb(old) ->> tg_argv[0])::uuid;",
+            "    if owned_id is not null then",
+            "        execute format('delete from %I.%I where \"Id\" = $1', tg_argv[1], tg_argv[2]) using owned_id;",
```
sven-n: **The generic function is slow and fails silently.**

- `to_jsonb(old)` serializes the whole deleted row, and `execute format(...)` prepares dynamic SQL, once per trigger per row. Deleting one `MagicEffectDefinition` fires 4 triggers, and bulk deletes (a `GameConfiguration`, mass character purges) pay this for every row.
- If a FK column is ever renamed, `->> tg_argv[0]` just returns NULL and the trigger does nothing, so the orphans come back without any error.

Generating one static function per relationship (`delete from data."ItemStorage" where "Id" = old."InventoryId"`) is cheaper, and it fails at migration time if the column doesn't exist.

---

src/Persistence/SourceGenerator/EfCoreModelGenerator.cs
```diff
@@ -367,6 +401,116 @@ protected void AddJoinDefinitions(Microsoft.EntityFrameworkCore.ModelBuilder mod
...
+
+        var parts = new List<string>
+        {
+            "-- Deletes the members of an aggregate which are referenced by their owner (one-to-one).",
+            "-- Collections of an aggregate don't need this, because their foreign key is held by the",
+            "-- child, so 'on delete cascade' already covers them.",
+            "-- This script is generated (see EfCoreModelGenerator) and idempotent.",
+            string.Empty,
+            CreateTriggerFunction(ConfigurationSchemaName),
+            CreateTriggerFunction(AccountDataSchemaName),
```
sven-n: Nit: `config.delete_owned_row` and `data.delete_owned_row` are identical, and the target schema is already a trigger argument, so one function would do. As it is, the body is duplicated in the generator output, the generated const and the migration copy, and a fix has to be made twice. (Moot if you switch to static per-relationship functions.)

---

sven-n (commented):
Thanks for tackling the orphaned item storages. I think the triggers conflict with the existing aggregate-delete mechanism, though, and would break ordinary deletes. Details are in the inline comments; here is the root cause.

`EntityFrameworkContextBase.DeleteAsync` → `ForEachAggregate` already calls `Context.Remove` on single-object `[MemberOfAggregate]` members (Inventory, Vault, MerchantStore, …). With the new AFTER DELETE triggers, the member row is gone by the time EF issues its own DELETE for it. That DELETE hits 0 rows and throws `DbUpdateConcurrencyException`. Affected paths: deleting a character in-game, the bot purge, and admin-panel deletes of config objects with one-to-one members.

The actual leak was narrower: `ForEachAggregate` isn't recursive, so deleting an account removes its characters but not the characters' inventories. Making `ForEachAggregate` recurse into the members it removes would fix that in one place, without triggers. If the triggers stay, `ForEachAggregate` has to stop removing non-collection members, or every `DeleteAsync` caller breaks.

This analysis comes from reading the code and EF's delete ordering. I haven't run it against Postgres. An integration test that deletes a character and an account with a vault would confirm it either way.

---

src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
sven-n: You're right, and I verified it in the code before changing anything: `ForEachAggregate` handles the single-object case as well (`else if (propertyValue is { }) action(propertyValue)`), and `Character.Inventory` / `Account.Vault` both carry `[MemberOfAggregate]`. So EF marks the storage Deleted itself, the trigger removes the row first, and EF's own DELETE hits 0 rows.

The triggers are gone. `ForEachAggregate` now walks the whole aggregate instead, which fixes the leak you identified (account → characters → inventories) in one place.

And thanks for the correction on `DeleteCharacterAction` — my claim that it leaked was wrong, and it was the headline argument in the issue and the PR description. I've corrected both.


---

src/GameLogic/Bots/BotGenerator.cs
sven-n: Confirmed — every bot account has a vault, so this would have failed for every single account and the purge would never have completed.

Both problems go away with the triggers: nothing deletes rows behind EF's back any more, so the change tracker stays consistent with the database and there are no stale `Unchanged` entries for rows that no longer exist.

The method keeps the full-graph reload, since `ForEachAggregate` can only walk members that are actually loaded. Only the manual storage deletes are gone — the recursion covers them now.


---

src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
sven-n: Same root cause, same resolution — the triggers are gone, so `EditConfigGrid` deletes behave exactly as before this PR.

Worth noting for the recursion, though: it now reaches one level deeper on the config side too. `MonsterDefinition` → `Buffs` → `Buff.MagicEffectDefinition` is the one place where that touches a *shared* row, because that annotation is wrong (the effect definition lives in `GameConfiguration.MagicEffects` and is also referenced by `Skill.MagicEffectDef` and `ItemDefinition.ConsumeEffect`). Deleting a merchant monster would now take the shared effect definition with it. Dropping `[MemberOfAggregate]` from `Buff.MagicEffectDefinition` is the right fix, but it changes the model (the generated cascade disappears, and `JsonQueryBuilder` would emit a `$ref` instead of inlining it), so it needs its own migration. I've noted it in the PR description and in #933 — say the word if you'd rather I guard the recursion here instead.


---

src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
sven-n: Verified — `MyNpgsqlMigrationsSqlGenerator` grants the config role only `GRANT SELECT ON TABLE data."<name>"` for `Item`, `ItemOptionLink`, `ItemItemOfItemSet` and `ItemStorage`, so both of those triggers would have been denied. I'd missed that the schema split is a privilege boundary, not just a namespace.

Moot now that the triggers are gone.


---

src/Persistence/EntityFramework/Migrations/20260904064500_AddAggregateDeleteTriggers.cs
sven-n: Partly extended, and the set turns out to be smaller than the trigger list suggested.

Because `ForEachAggregate` already removed the *direct* members, a one-to-one member only leaked when it sat at depth 2 or deeper from whatever was passed to `DeleteAsync`. `LetterBody.SenderAppearance` is a direct member of the letter body, so `DeleteAsync(letterBody)` already took it. The collections at depth 2+ didn't leak either — their foreign key is at the child, so the database cascade reaches them.

The cleanup now covers both data-schema cases: item storages (account → characters → inventory) and appearance data, as a safety sweep — `LetterBody` is the only table referencing it, so a `not exists` sweep can only hit rows nothing points at.

Config-side depth-2 orphans are left (a map's `BattleZone` → `Ground`/`LeftGoal`/`RightGoal` rectangles, a monster's `ItemCraftings` → `SimpleCraftingSettings`, …), and the description now says so. I didn't want to sweep config tables on a `not exists` over references I'd have to enumerate by hand: get one referencing column wrong there and the migration deletes live configuration, which is a worse outcome than a few unreachable rows. Happy to add them if you'd rather have them gone — the enumeration is mechanical from the model snapshot, it just wants a second pair of eyes.


---

src/Persistence/SourceGenerator/EfCoreModelGenerator.cs
sven-n: Both points stand, and the silent-NULL one is the worse of the two — a trigger that quietly stops working after a rename is exactly the kind of thing nobody notices until the orphans are back. Moot now: the generator is untouched again and there are no functions or triggers at all.


---

src/Persistence/SourceGenerator/EfCoreModelGenerator.cs
sven-n: Agreed, and moot — the generator is back to its state on master, so there's nothing duplicated any more.


---

sven-n:
Closed the two points I'd left hanging on a question rather than waiting for an answer.

**The shared configuration objects are now guarded** (`8720603d`). I'd flagged that the recursion reaches `Buff.MagicEffectDefinition` — a row that lives in `GameConfiguration.MagicEffects` and is used by a skill and an item definition too — and asked whether to guard it here. It is a regression this PR would have introduced, so it shouldn't wait: when deleting, the traversal now stops when it reaches an object which a collection of the `GameConfiguration` holds, if it got there *through the aggregate of a member*. Deleting a monster definition removes its buffs, and leaves their effect definitions alone.

The "through a member" part matters: direct members are treated exactly as before, so nothing that was deleted before this PR is kept now. Deleting a `Buff` on its own still removes its effect definition, as it always did — that's the annotation being wrong, which is a separate fix (it changes the model and needs its own migration). Detaching still walks everything, since a stale configuration has to be detached whole.

**The config-side orphan cleanup stays out**, deliberately rather than pending a decision. The migration sweeps `data."ItemStorage"` and `data."AppearanceData"`, where I could verify the complete set of referencing columns from the model snapshot. Doing the same for the config tables means enumerating referencing columns by hand across a dozen types, and one mistake there deletes live configuration — a worse outcome than leaving a few unreachable rows. It's a good follow-up with someone checking the list.

CI was green on the previous head; this push re-runs it.


---

sven-n:
Both remaining review points are done rather than deferred (`5dfba147`).

**The cleanup now covers the whole aggregate, not just item storages.** My reason for skipping the rest was that hand-enumerating referencing columns is how you delete live configuration by accident — so the migration doesn't enumerate them. It reads them from `pg_constraint` at run time, which means a foreign key I don't know about still protects a row, and one added later is covered without touching the migration.

What did need a decision is *which* tables can be swept by reachability at all. That turns out to be a property of the model rather than a judgement: a table whose rows can also be owned through a foreign key of their own can't be swept this way, because "nothing points at it" is then true for rows that are perfectly alive — an item in a storage holds that reference itself, so the naive sweep would delete almost every item of every player. Those tables are exactly the element types of `[MemberOfAggregate]` collections, which gives five exclusions (`Item`, `PowerUpDefinition`, `MonsterSpawnArea`, `CastleSiegeZoneDefinition`, `MagicEffectDefinition`) and eleven tables swept.

Two guards keep it honest: the tables of the target's own aggregate are ignored when looking for references (they point at their owner, and an owner only its own members still reference is precisely what is unreachable), and a target that nothing points at at all is skipped entirely — so a missing constraint can never degrade the statement into a full delete.

**The integration test is in**, deleting an account with a vault and a character with an inventory and asserting no item storage survives. It needs a database, so it's `[Ignore]`d like `BackupServiceEfCoreTests`, which also means I've never run it — worth a manual run, along with the migration against a copy of a real database.


---

PR #936, opened 2026-09-04
Tighten bot party formation: hunt within 6, regroup within 2 of leader

eduardosmaniotto:
Tightens bot party-member formation in `BotNavigator`:

- `FollowDistance`: 10 → 2 tiles (regroup radius).
- New `EvaluateFollowerHuntingAsync`: close-range followers hunt only
  monsters within `HuntingRange` (6) of the leader via
  `HuntingOrigin = leader.Position`, and walk back to within 2 tiles
  when there is nothing to fight.
- Followers no longer fall through to independent monster-homing /
  random ground picking or solo warps; the leader decides where the
  party hunts. In town they stay stacked on the leader.

## Why
Previously followers roamed breaking party cohesion, EXP sharing and heal/buff coverage.

### DIFF
--- a/src/GameLogic/Bots/BotBuffHandler.cs
+++ b/src/GameLogic/Bots/BotBuffHandler.cs
@@ -97,7 +97,7 @@ internal static class BotBuffHandler
         {
             return effects.Values.ToArray().Any(e => e?.Definition == effectDef);
         }
-        catch (InvalidOperationException ex)
+        catch (Exception ex) when (ex is InvalidOperationException || ex is ArgumentException)
         {
             player.Logger.LogDebug(ex, "Bot '{Name}' encountered a concurrent modification while inspecting active effects.", player.Name);
             return true;
--- a/src/GameLogic/Bots/BotNavigator.cs
+++ b/src/GameLogic/Bots/BotNavigator.cs
@@ -134,7 +134,7 @@ internal sealed class BotNavigator : AsyncDisposable
     private const int MaxPointPickAttempts = 25;
 
     /// <summary>A party member keeps within this distance (tiles) of its leader; beyond it, it walks back.</summary>
-    private const int FollowDistance = 10;
+    private const int FollowDistance = 2;
 
     /// <summary>
     /// Hunting grounds closer than this (tiles) to a recent death site are avoided after the same
@@ -644,8 +644,9 @@ internal sealed class BotNavigator : AsyncDisposable
         }
 
         // Party members follow their leader instead of roaming on their own: to the leader's map when
-        // it warped away, and towards the leader when it moved off. Near the leader they hunt normally
-        // (the party heal/buff handlers and the party experience bonus need the group to stay together).
+        // it warped away, and towards the leader when it moved off. Near the leader they hunt around
+        // him (within hunting range) and regroup to within the follow distance when idle - they never
+        // roam to independent hunting grounds.
         // Followers never warp on their own - the leader (the lowest-level member, so it only ever picks
         // maps the whole group can hunt on) decides where the party goes.
         var leaderToFollow = this.GetPartyLeaderToFollow();
@@ -655,6 +656,9 @@ internal sealed class BotNavigator : AsyncDisposable
             {
                 return;
             }
+
+            await this.EvaluateFollowerHuntingAsync(map, leaderToFollow, inSafezone, cancellationToken).ConfigureAwait(false);
+            return;
         }
 
         // A bot whose map is no longer the best it could safely hunt warps away with priority - even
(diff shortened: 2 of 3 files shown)


sven-n:
## Review

The intent (leader decides where the party hunts, followers tether their combat to him) is right and fits the existing design. But as written, the two new rules fight each other, and one of the new branches can't be reached. Details below.

### 1. The regroup branch in `EvaluateFollowerHuntingAsync` is dead code

`EvaluateFollowerHuntingAsync` is only called when `TryFollowLeaderAsync` returned `false`, and on the same map that happens exactly when `GetDistanceTo(leader.Position) <= FollowDistance` (`BotNavigator.cs:1166`). So this can never be true inside the new method:

```csharp
if (this._player.GetDistanceTo(leader.Position) > FollowDistance)
```

The walk-back it describes is already performed one level up, by `TryFollowLeaderAsync`.

### 2. `FollowDistance` (2) < `HuntingRange` (6) makes followers oscillate instead of fighting

This is the substantive one. The tether says "fight monsters within 6 of the leader"; `MovementHandler.MoveCloserToTargetAsync` will therefore walk the bot toward a monster up to ~6 tiles from the leader. On the next idle tick, `TryFollowLeaderAsync` sees distance > 2 and consumes the tick walking straight back to the leader — before the hunting evaluation ever runs. A monster at range 3–6 from the leader is a valid target the follower can never stay next to: it ping-pongs and lands no hits.

The unconditional walk-back has to be gated on "nothing to fight around the leader", which is precisely the state the new method computes. Suggestion: drop the distance check from `TryFollowLeaderAsync`'s same-map path (leave the map-following there), and let `EvaluateFollowerHuntingAsync` own both the tether and the regroup — then its `> FollowDistance` branch stops being dead and becomes the only walk-back. Alternatively, keep the regroup radius >= `HuntingRange` so combat drift never trips it.

### 3. Lowering `FollowDistance` also changes mini-game behavior

`FollowDistance` is used in `EvaluateInsideMiniGameAsync` (`BotNavigator.cs:1021`) too. At 2 tiles, event bots re-path to the leader on almost every idle tick, and that branch `return`s before the "close in on the nearest event target" logic — so Blood Castle / Chaos Castle followers get noticeably worse. The PR description only claims party-formation scope. Please split the constants: e.g. keep `FollowDistance` as the follow/mini-game trigger and add a separate `PartyRegroupDistance = 2` for the new formation radius, each with its own doc comment.

### 4. The tether is only applied on ticks that reach the new branch

`EvaluateAsync` unconditionally sets `HuntingOrigin = ... : this._player.Position` at line 564, and several early returns sit between that and the new call — most importantly `if (this._player.IsWalking) return;`. So on every tick where the follower is walking, the origin is the bot's own position again and the combat handler is free to pick a target away from the leader; the tether only holds on idle ticks. If the tether is meant to hold continuously, set it where the origin is decided (line 564: use the leader's position when `GetPartyLeaderToFollow()` is non-null), not after the early returns.

### Minor

- The tether check uses the local `HuntingRange` constant, while the combat handler's effective range is `Math.Max(DefaultRange, config.HuntingRange)` and can be larger. With a configured range above 6 the navigator would consider the ground empty while the combat handler still has targets. Pre-existing elsewhere in the file, but it matters more here since this value now gates whether the bot moves at all.
- `map.GetAttackablesInRange(...).OfType<Monster>().Any(...)` runs per follower per tick against the leader's position; for a large party on a busy map that is the same scan repeated N times. Not blocking, but worth a thought if party sizes grow.
- Behavioral edge: followers no longer reach `TryWarpToBetterMapAsync`, the barren-map step-down, or ground picking at all. If the leader idles in town (or is wedged), the whole party now idles with him indefinitely. That may be the intended trade-off — if so, please say so in the comment block, since the previous fall-through was the safety net.
- No test covers the new method. `tests/Offline/` already exercises `HuntingOrigin`-driven behavior; a test asserting "monster near leader → no regroup walk" and "nothing near leader → walks back" would pin down exactly the interaction that is currently broken by (2).

Happy to look again once the ordering between following and hunting is settled.

---

sven-n:
## Re-review of `1c1931d`

All four points from the previous round are addressed, and correctly:

1. **Dead regroup branch** — fixed by moving same-map spacing out of `TryFollowLeaderAsync` into `EvaluateFollowerHuntingAsync`. That method now owns the walk-back and it is reachable.
2. **Oscillation** — fixed at the root: the regroup walk is gated on "no fightable monster near the leader", so combat drift out to hunting range no longer yanks the follower back mid-fight. This is the right ordering — `TryFollowLeaderAsync` is now purely cross-map, and its doc comment and return contract say so.
3. **Mini-game regression** — fixed by splitting `FollowDistance` (10, mini-game follow trigger, unchanged) from the new `PartyRegroupDistance` (2, open-world formation). `EvaluateInsideMiniGameAsync` keeps its old behavior.
4. **Intermittent tether** — fixed by resolving the leader before the origin assignment at the top of `EvaluateAsync`, so `HuntingOrigin` holds on walking and errand ticks too.

The extra touch in the stuck-watchdog branch (a wedged follower regroups on its leader instead of picking a random ground) is a good catch that follows from the same principle, and it happens to be what rescues the edge case below.

### Two remaining nits, neither blocking

- **Follower in the safezone while the leader is outside** (typical after a death respawn): `isTetheredToLeader` deliberately excludes `inSafezone` at the origin assignment, but `EvaluateFollowerHuntingAsync` sets `HuntingOrigin = leader.Position` unconditionally before its `inSafezone` early return — so for that one case the origin does point out of town, which is exactly what the original comment at the assignment avoids. The follower also clears its destination and stands in town, since the regroup branch sits behind the early return. The stuck watchdog does pull it out to the leader after the timeout, so it recovers rather than deadlocks; letting the safezone branch walk to the leader when the leader is outside the safezone would just make it immediate.
- **No test coverage** for the new method. The interaction worth pinning is precisely the one that was broken in the first commit: monster near leader → no regroup walk; nothing near leader and beyond `PartyRegroupDistance` → walks back.

### Out of scope

The two `catch (InvalidOperationException)` → `catch (Exception ex) when (ex is InvalidOperationException || ex is ArgumentException)` hunks in `BotBuffHandler` and `BuffHandler` are unrelated to party formation and unmentioned in the PR description. The change itself is sound (`ToArray()` over a concurrently mutated dictionary can throw `ArgumentException` for a too-short destination array, and the handling is the same debug-log-and-bail), so I am not holding the PR for it — but it belongs in its own commit or PR with its own description.

CI is green and the branch merges cleanly. Merging.

---

Issue #730, opened 2026-03-24
Castle Siege Mini-Map & Guild Commands


sven-n:
## Summary

Implement the Castle Siege mini-map feature (showing ally and NPC positions during battle) and the guild command system (alliance masters issuing directional orders to their side).

## Prerequisites

- #724 (Guild Selection) — for join side awareness and final guild list.
- #725 (NPCs) — for NPC position tracking.
- #722 (State Machine) — for `Start` state tick handlers.

### 1. Mini-Map Feature

**Access Control:**
- Only **alliance masters** of participating guilds can view the mini-map.
- Only available during `Start` state.
- Server tracks which players have requested the mini-map in `CastleSiegeContext`.

**Update Cycle (every 3 seconds during `Start`):**

1. For each participating guild with an online guild master who requested the mini-map:
   a. Collect positions of all same-side players in the CS map (up to 1000 points per guild).
   b. Identify the guild master position separately.
   c. Collect positions of NPC objects: gates and guardian statues (alive ones).
2. Send two packets to each requesting guild master:
   - `CastleSiegeMiniMapPlayerPositions` — member coordinates.
   - `CastleSiegeMiniMapNpcPositions` — gate/statue coordinates.

### 2. Mini-Map Request Handler

When a player sends a mini-map request:
- Validate: player is an alliance master of a participating guild, state is `Start`.
- Add to the mini-map request tracking set.
- Send initial acknowledgment.

### 3. Guild Commands

Alliance masters can issue commands to all same-side members on the CS map.

**Packet structure:**
```
team : byte     — Team identifier
x : byte        — Target X coordinate
y : byte        — Target Y coordinate
command : byte  — Command type
```

**Processing:**
1. Validate: player is an alliance master of a participating guild, state is `Start`.
2. Rebroadcast the command to all players on the same side in the CS map.

### 4. Mini-Map Data Gathering — **New file: `GameLogic/CastleSiege/CastleSiegeMiniMap.cs`**

```csharp
public class CastleSiegeMiniMap
{
    public async ValueTask UpdateAsync(CastleSiegeContext context, GameMap csMap)
    {
        // 1. Iterate all players in csMap
        // 2. Group by guild, filter participating
        // 3. Build position data per guild
        // 4. Build NPC position data
        // 5. Send to each requesting alliance master
    }
}
```

### 5. View Interfaces — **New files in `GameLogic/Views/CastleSiege/`**

[...]

PR #941, opened 2026-09-05
Implement Castle Siege Mini-Map & Guild Commands (#730)

didiconcs:
## Summary

Implements Castle Siege Phase 11 (#730): the mini-map broadcast and guild-command relay.

- **`CastleSiegeMiniMap`** gathers own-side player positions and alive gate/statue positions, and pushes them every 3 seconds to every online alliance master of a participating guild, while the siege is in the `Start` state.
- **`CastleSiegeGuildCommandAction`** lets an alliance master issue a directional command (type + coordinates) to every player on their own side.
- New `ICastleSiegeMiniMapPlugIn` / `ICastleSiegeCommandPlugIn` view interfaces and matching `GameServer/RemoteView` + `GameServer/MessageHandler` plug-ins, following the existing Castle Siege conventions.

## Design decisions (spec doesn't fully cover the current codebase shape)

The original implementation plan for this phase predates phases 4-10, so a few things needed a concrete call rather than a literal reading, flagging these for review:

- **Push, not subscribe.** There's no existing "open mini-map" request packet, so every online alliance master of a participating guild is treated as an implicit recipient of the 3-second push, rather than gating on a per-player "wants updates" request. Happy to add a request packet + flag if a stricter model is preferred.
- **`CastleGuildCommand.Team` is never trusted for authorization or audience selection.** The issuer's actual side is always re-derived server-side from `CastleSiegeContext`, so a spoofed `Team` byte can't mis-target orders or leak to the wrong side.
- **Alliance master** = the issuing guild's `GuildMaster`, where that guild's `CastleSiegeGuildParticipant.IsAllianceMaster` is `true` in `context.FinalGuildList`.
- Both mini-map and guild commands are restricted to `CastleSiegeState.Start`. The spec says so explicitly for the mini-map; for guild commands this is inferred (commands are meaningless outside battle) rather than spec-mandated, worth confirming that's the intended behavior.
- Only alive gates/statues appear on the NPC layer (`CastleSiegeMiniMapNpcType` only defines `Gate`/`GuardianStatue`, matching the wire format).

## Verification

- `dotnet build src/Startup/MUnique.OpenMU.Startup.csproj -c Debug`: 0 errors.
- `dotnet test tests/MUnique.OpenMU.Tests --filter FullyQualifiedName~CastleSiege`: 106/106 passing (existing suite, confirms no regression, this PR doesn't add new automated tests yet, since the mini-map/command flow is push-based and harder to unit-test without a fuller integration harness; happy to add coverage if there's a preferred pattern from the recent Castle Siege PRs).

## Known minor items

A couple of small things worth a look, not blocking:
- Client-supplied command coordinates are relayed without validation against actual map bounds.
- The `Start`-state check is read once before acquiring the context lock and isn't re-checked once held, so a command could in theory land in the narrow window right as the state transitions away from `Start`.


sven-n:
## Review

Overall this is well-executed and fits the existing Castle Siege conventions closely: the handler mirrors `CastleSiegeGateOperateHandlerPlugIn`, the two remote views use the exact `GetRequiredSize` / `GetSpan(size)[..size]` / `SendAsync(Write)` idiom of `CastleSiegeGuildListPlugIn`, and splitting a lock-free `GetDefenseStructures` out of `GetDefenseStructureSnapshotAsync` (with the remark explaining why) is the right call — `GetRuntimeSnapshot` copies under `_runtimeLock` and releases before returning, so there is no lock-ordering hazard with `ExecutionLock`.

I checked the wire format against the client (`sven-n/MuMain`). The offsets line up exactly: `PWHEADER_DEFAULT_WORD2` = 4-byte C2 header + `INT Value` → count at index 4, entries at 8 (`CastleSiegeMiniMapPlayerPositions`), and `PWHEADER_DEFAULT_WORD` = header + `BYTE Value` → count at 4, entries at 5 (`CastleSiegeMiniMapNpcPositions`). `NpcType` 0/1 maps to the client's `m_byType + 1` (1 = gate, 2 = statue). Good.

Two things came out of that client comparison that I think need to change, plus a few smaller items.

---

### 1. `Team` is a command-group index, not a side — overwriting it breaks the feature

This is the main finding, and it inverts the design decision in the PR description.

In the client, `Team` is **not** an audience or authorization field. It is the number of the command group (squad) the commander picked in the UI:

- `NewUISiegeWarCommander.cpp:82-88` — `SelectCmd.byTeam = m_iCurSelectBtnGroup;` then `SendCastleGuildCommand(...)`. `m_iCurSelectBtnGroup` is the index of one of `MAX_COMMANDGROUP` (= 7, `NewUISiegeWarBase.h:10`) group buttons.
- On receive, `NewUISiegeWarBase.cpp:539` — `SetMapInfo` stores into `m_CmdBuffer[data.byTeam]`, one slot per group.
- `NewUISiegeWarBase.cpp:417-460` — `RenderCmdIconInMiniMap` renders **all 7 slots simultaneously** and draws the label `byTeam + 1` ("1".."7") next to each icon.

So the audience is already fully determined server-side by *which players receive the packet* — `Team` only selects the display slot and the number shown on the mini map. Replacing it with `(byte)CastleSiegeJoinSide` (1..4) means:

- the commander's group selection is silently discarded — an order for group 5 renders as "2".."5" depending on the issuer's side;
- all 7 independent markers collapse into a single slot per side, so a new order always overwrites the previous one instead of standing alongside it;
- issue #730's packet spec ("`team : byte` — Team identifier") and the packet definition's own "Team number from 0 to 7" are not honoured.

Recommendation: **relay `request.Team` unchanged**, but validate it — and this part matters, because `SetMapInfo` indexes `m_CmdBuffer[data.byTeam]` with **no bounds check** (`GuildCommander m_CmdBuffer[7]`, `NewUISiegeWarBase.h:95`). A client sending `Team = 200` would currently be an out-of-bounds write in every same-side player's client. So the handler should drop the packet (or clamp) when `Team > 6`. The rendering path does guard with `byTeam >= 0 && byTeam <= 6`, but `SetMapInfo` writes before that check runs.

That also flips the two items in "Known minor items":

- **Coordinate validation is a non-issue** — `PositionX`/`PositionY` are bytes and the siege map is 256×256, so every possible value is in bounds. Nothing to add there.
- **The `Team` byte you deliberately dropped is the one that actually needs validation.** Worth swapping the emphasis in the code comment too.

Sanity check on the enum mapping while relaying: `_ => Wait` turns any out-of-range `Command` byte into a valid "wait" marker. The client's `RenderCmdIconInMiniMap` treats `byCmd == 3` as "empty slot" and leaves `iWidth`/`iHeight` uninitialized for anything above 2, so rejecting values > 2 outright is safer than mapping them to `Wait`.

### 2. Re-check the `Start` state inside the lock

You flagged this yourself, and it's a two-line fix with a direct precedent: `CastleSiegeGateOperateAction.IsAuthorizedAsync` reads `context.CurrentState` while holding `ExecutionLock`. Moving the `CurrentState: Start` check from the pre-lock pattern match into the `try` block closes the window and matches the neighbouring action. Keep the cheap `Configuration.Enabled` check outside if you like.

While you're in there: the issuer's side comes from `participant.Side` but the recipient filter uses `context.GetPlayerJoinSide(candidate)`. Those are two different sources of truth — `GetTrackedPlayerJoinSide` prefers the per-character `PlayerJoinSides` entry and only falls back to the guild list. They can diverge briefly after `CastleSiegeCrownMechanics` swaps `guild.Side` on a crown capture (the resync via `SetPlayerJoinSideAsync` happens right after, but not atomically). Using `context.GetPlayerJoinSide(player)` for the issuer as well makes the fan-out self-consistent — the `GetSiegePlayers().Contains(player)` check already guarantees it won't return `None` for map reasons. Same applies to `entry.Participant!.Side` in `CastleSiegeMiniMap.BroadcastAsync`.

### 3. Heads-up: the current client discards the mini-map packets entirely

Not caused by this PR, but it means the mini-map half can't be verified end-to-end today, so it's worth knowing before this is called done.

`WSclient.cpp:12028` and `12048` both guard with:

```cpp
if (g_pSiegeWarfare->GetCurSiegeWarType() != TYPE_GUILD_COMMANDER)
    return;
```

`GetCurSiegeWarType()` returns `m_iCurSiegeWarType`, which is assigned from `CNewUISiegeWarfare::SIEGEWAR_TYPE_*` (`NewUISeigeWarfare.h:20-26`: `NONE = -1, OBSERVER = 0, COMMANDER = 1, SOLDIER = 2`). But `TYPE_GUILD_COMMANDER` is from the unrelated `csmapinterface` enum (`_enum.h:3601-3607`: `TYPE_OBSERVER = 0, TYPE_GUILD_SOLDIER = 1, TYPE_GUILD_COMMANDER = 2`). So the guard admits only **soldiers** (2) and rejects commanders (1) — and `CNewUISiegeWarfare::SetGuildMemberLocation` then no-ops for anyone who isn't `SIEGEWAR_TYPE_COMMANDER` (1). Net effect: both `B6` and `BB` are always dropped. The two constants need to be unified on the client side (a `sven-n/MuMain` change) before any of this is observable.

### 4. Mini-map request handling vs. the spec

The push model is a defensible call and there is genuinely no client→server mini-map request packet in `ClientToServerPackets.xml` today. Two things to square up though:

- #730's acceptance criterion is "Only alliance masters of participating guilds can **request** the mini-map", with `CastleSiegeMiniMapRequestHandlerPlugIn` and a tracking set in the context. If push is the accepted model, that criterion should be struck from the issue rather than left open.
- The XML descriptions ("sends ... to a Castle Siege mini-map **requester**") and the new doc comments ("broadcasts ... to **requesting** alliance masters", `CastleSiegeMiniMap.cs`) still describe a request model that doesn't exist. Worth rewording to match the implementation.

One ordering detail is load-bearing and undocumented: `ReceiveGuildMemberLocation` calls `ClearGuildMemberLocation()` for the *whole* buffer, and `ReceiveGuildNpcLocation` only appends. So the player packet must be sent before the NPC packet in every cycle, and the player packet must be sent even when the list is empty — which `SendAsync` does correctly today. A short comment there would stop a future refactor from swapping or short-circuiting them.

Also missing from the spec: the "up to 1000 points per guild" cap (the client reserves 1600 in `m_vGuildMemberLocationBuffer`). It's a `std::vector`, so there's no overflow risk, and the C2 length field won't wrap until ~32k players — but a `.Take(1000)` is free and keeps the packet honest.

### 5. Tests

`tests/MUnique.OpenMU.Tests` already has the pattern this needs: `CastleSiegeNpcRemoteViewTests`, `CastleSiegeCrownRemoteViewTests`, `CastleSiegeMachineRemoteViewTests` etc., all built on `CastleSiegeRemoteViewTestHelper.CreatePlayer()` which hands you a player plus the raw output. Serializing `ShowPlayerPositionsAsync` / `ShowNpcPositionsAsync` / `ShowGuildCommandAsync` and asserting the bytes is a dozen lines each and requires none of the integration harness you mentioned — it would also have caught the `Team` semantics above. That's the gap I'd most like closed before merge; the push/broadcast loop itself is fine to leave uncovered for now.

### Smaller items

- **No rate limit on the command handler.** Each accepted packet fans out to every same-side player on the map, with no cooldown, and takes `ExecutionLock` on the way. The client's `byLifeTime = 100` means rapid re-issues are pointless for gameplay, so a ~1 s per-player cooldown in the action would cost nothing and removes the amplification vector.
- **Tick placement.** `Task.WhenAll` in `BroadcastAsync` surfaces exceptions, and the mini-map block sits *before* the `NextNpcSaveUtc` block in `OnTickAsync`. A single failing send would skip the periodic NPC state save for that tick. Moving the block after the save (or catching around the fan-out) makes that independent.
- **`PlugInResources.Designer.cs`** — the diff also strips trailing whitespace from two unrelated lines (617, 1031). Harmless, but it makes the file diverge from what the ResX generator emits; worth reverting to keep the generated file byte-clean.
- The spec listed `ICastleSiegeMiniMapNpcPlugIn` / `CastleSiegeMiniMapNpcPlugIn` as separate files. Merging both sends into one interface is cleaner and I'd keep it — just noting it as a deliberate deviation from the file list in #730.
- `GetSiegePlayers()` is re-enumerated a few times per broadcast (once for recipients, once per distinct side). Irrelevant at real player counts; only worth folding into a single grouped pass if you're touching the method anyway.

CI is green and the packet definitions were already in `master`, so nothing else needs regenerating. With the `Team` relay fixed and the state check moved under the lock, this looks good to me.

---

didiconcs:
Thanks for the thorough review, especially for checking the wire format against the client, that caught something real.

Addressed everything:

1. **Team semantics fixed.** `Team` is now relayed unchanged from the request instead of being overwritten with `(byte)side` - the handler drops the packet if `Team > 6` (bounds check for the client's `GuildCommander[7]`) or if `Command` doesn't map to a known value (previously defaulted to `Wait`, now rejected instead).
2. **Start-state check moved inside `ExecutionLock`** in `CastleSiegeGuildCommandAction`, matching `CastleSiegeGateOperateAction`'s pattern.
3. **Issuer side now resolved via `context.GetPlayerJoinSide(player)`** in both the guild-command action and `CastleSiegeMiniMap.BroadcastAsync`, same source as the recipient filter.
4. **Tests added** (`CastleSiegeMiniMapRemoteViewTests`): mini-map player/NPC packet serialization, the empty-list-still-sent case, and specifically a regression test asserting `ShowGuildCommandAsync` relays `Team` unchanged - this one would have caught the original bug.
5. Reworded the mini-map XML/doc comments away from "requester" language to match the push model that's actually implemented.
6. Added the 1s per-player cooldown and the `.Take(1000)` cap you flagged under "smaller items", and moved the mini-map broadcast to run after the periodic NPC save in `OnTickAsync` so a failing send can't skip it.
7. Reverted the incidental whitespace diff in `PlugInResources.Designer.cs`.

Left `#730`'s "only alliance masters can **request** the mini-map" acceptance criterion as-is since I can't edit that issue myself, flagging here that it should be struck or reworded given the push model - happy to leave that to a maintainer.

On your heads-up about the client dropping these packets today (`TYPE_GUILD_COMMANDER`/`SIEGEWAR_TYPE_COMMANDER` mismatch): noted, that's tracked as a `sven-n/MuMain` issue, not something for this PR.

123/123 tests passing, clean build.

sven-n:
## Review of 65546e19

Everything from the last round is addressed, and addressed properly rather than literally — the `Team > 6` rejection, the unmapped-`Command` rejection, the state check under the lock, and the comment explaining *why* the player packet must precede the NPC packet are all exactly right. The regression test asserting `ShowGuildCommandAsync` relays `Team` unchanged is the right test to have written.

Five things from the new revision, one of which is a regression the fix introduced.

### 1. `BroadcastAsync` can throw `KeyNotFoundException`

`CastleSiegeMiniMap.cs:36-53` now evaluates `context.GetPlayerJoinSide(player)` **twice** per recipient — once to build the dictionary keys, once to index it:

```csharp
var playersBySide = recipients
    .Select(player => context.GetPlayerJoinSide(player))
    .Distinct()
    .ToDictionary(side => side, side => GatherPlayerPositions(context, side));

await Task.WhenAll(recipients.Select(player =>
        SendAsync(player, playersBySide[context.GetPlayerJoinSide(player)], npcs).AsTask()))
```

The old `entry.Participant!.Side` was a value read once, so this is new. `GetPlayerJoinSide` starts with `Configuration.CastleSiegeMapDefinition?.Number != player.CurrentMap?.Definition.Number → None`, and `player.CurrentMap` is not covered by `ExecutionLock` — a recipient who warps out or disconnects between the two evaluations yields `None` on the second call, which is not a key in the dictionary. `playersBySide[None]` then throws.

It lands in `ExecuteTaskAsync`'s catch-all and, since the broadcast is now last in the tick, nothing downstream is skipped — so it's an error log and a lost mini-map cycle, not a stuck siege. Still trivially avoidable by resolving the side once, which is roughly the tuple shape you had before:

```csharp
var recipients = context.GetSiegePlayers()
    .Where(player => CastleSiegeGuildResolver.ResolveParticipatingAllianceMaster(player, context) is not null)
    .Select(player => (Player: player, Side: context.GetPlayerJoinSide(player)))
    .ToList();
```

then key and index off `entry.Side`. That also drops one `GetPlayerJoinSide` call per recipient.

### 2. The cooldown doesn't close the vector its comment describes

`CastleSiegeGuildCommandAction.cs` — the comment says the cooldown removes an amplification vector because each command "takes `ExecutionLock` along the way", but the check runs *after* the lock has already been taken and released. A client spamming subcode `1D` still forces one `ExecutionLock.WaitAsync()` per packet, serialized against the siege tick; only the fan-out is suppressed.

Worse for the case you'd most want to stop: validation failures `return` before the cooldown code, so an **unauthorized** spammer (any player on the map, no guild-master check passed) never stamps the timestamp and is never rate-limited at all — they take the lock on every single packet.

Moving the check-and-stamp above `await context.ExecutionLock.WaitAsync()` fixes both: it becomes a pure pre-lock gate that applies to authorized and unauthorized senders alike, which is what the comment claims.

### 3. Cooldown uses `DateTime.UtcNow` instead of the injected clock

Castle Siege is otherwise consistent about `TimeProvider` — `CastleSiegePlugIn._timeProvider`, `CastleSiegeContext(..., TimeProvider? timeProvider = null)`, and `CastleSiegeStateMachineTests` drives the whole state machine off a fake one. `DateTime.UtcNow` here is the only wall-clock read in the feature, and it makes the cooldown untestable: there's no way to assert "second command within 1 s is dropped, after 1 s is delivered" without a real `Task.Delay`.

`CastleSiegeContext._timeProvider` is private with no accessor, so this needs a one-line `internal TimeProvider TimeProvider => this._timeProvider;` on the context, then `context.TimeProvider.GetUtcNow().UtcDateTime` here. Cheap, and it makes the branch testable.

### 4. The Designer.cs revert dropped the file's BOM

`src/GameServer/Properties/PlugInResources.Designer.cs` line 1 lost its UTF-8 BOM in this commit:

```
master: ef bb bf 2f 2f 2d ...
pr941:  2f 2f 2d ...
```

Every other generated designer in the repo still has it (`src/GameLogic/Properties/PlugInResources.Designer.cs`, `src/DataModel/Properties/Resources.Designer.cs`, `src/ClientLauncher/...`, `src/Network/Analyzer.WinForms/...`). The whitespace revert itself is correct — this is just an editor side effect on the same file. Re-adding the BOM makes the diff purely additive again.

### 5. Test coverage: the two branches that protect the client aren't covered

The new tests cover the serialization side well. What's untested is the part of this round that actually matters for safety — `CastleSiegeCommandHandlerPlugIn`'s `Team > 6` and unmapped-`Command` rejections. Those exist specifically to keep an out-of-bounds write off every recipient's client, and nothing would notice if a later refactor turned either back into a clamp or a `_ => Wait`.

There's direct precedent for testing handlers in this repo — `CastleSiegeNpcTests.NpcRequestHandlersUseExpectedSubcodes` (asserts each handler's `Key`) and `NpcRequestHandlersIgnoreTruncatedPacketsAsync` (feeds `Memory<byte>.Empty`), plus `CastleSiegeRegistrationTests` calling `HandlePacketAsync` with a crafted packet. `GameServer` has `InternalsVisibleTo("MUnique.OpenMU.Tests")`, so the internal handler is directly constructible. Three short cases would do it: `Key == 0x1D`, a truncated packet is ignored, and `Team = 7` / `Command = 9` produce no view call.

### Nits

- `CastleSiegeMiniMap.cs` — `MaximumPlayersPerGuild`'s comment says it matches "the client's `m_vGuildMemberLocationBuffer` reservation", but the client reserves **1600** (`NewUISiegeWarCommander.cpp:22`). 1000 is #730's number. Worth citing the issue instead, so nobody later "corrects" the constant to 1600 against a vector that grows anyway.
- `CastleSiegeMiniMapRemoteViewTests.SerializeGuildCommandMapsCommandTypesAsync` — the doc comment promises "including the fallback for the 'wait' value", but the body only asserts `Defend`. Either add the `Wait` case or drop the clause.
- `ResolveParticipatingAllianceMaster(player, context) is not { }` reads as a double negative; `is null` says the same thing more plainly now that the binding isn't used.

None of these block on their own. #1 is the one I'd fix before merge; #2 is a small change that makes the cooldown do what its comment says.

---

src/GameLogic/CastleSiege/CastleSiegeMiniMap.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        var npcs = GatherNpcPositions(context);
+        var playersBySide = recipients
+            .Select(player => context.GetPlayerJoinSide(player))
+            .Distinct()
+            .ToDictionary(side => side, side => GatherPlayerPositions(context, side));
+
+        await Task.WhenAll(recipients.Select(player =>
+                SendAsync(player, playersBySide[context.GetPlayerJoinSide(player)], npcs).AsTask()))
```
sven-n: `context.GetPlayerJoinSide(player)` is evaluated twice per recipient — once here to build the keys, once on line 52 to index the dictionary. The previous `entry.Participant!.Side` was read once, so this is new in 65546e19.

`GetPlayerJoinSide` starts with `Configuration.CastleSiegeMapDefinition?.Number != player.CurrentMap?.Definition.Number → None`, and `player.CurrentMap` is not covered by `ExecutionLock`. A recipient who warps out or disconnects between the two evaluations yields `None` on the second call, which is not a key in `playersBySide` → `KeyNotFoundException`.

It's caught by `ExecuteTaskAsync`'s catch-all and, now that the broadcast runs last in the tick, nothing downstream is skipped — so it's an error log and a lost mini-map cycle rather than a stuck siege. Still worth resolving the side once, which is roughly the tuple shape you had before:

```csharp
var recipients = context.GetSiegePlayers()
    .Where(player => CastleSiegeGuildResolver.ResolveParticipatingAllianceMaster(player, context) is not null)
    .Select(player => (Player: player, Side: context.GetPlayerJoinSide(player)))
    .ToList();
```

then key and index off `entry.Side`. That also drops one `GetPlayerJoinSide` call per recipient.

---

src/GameLogic/CastleSiege/Actions/CastleSiegeGuildCommandAction.cs
```diff
@@ -0,0 +1,93 @@
...
+            // diverge from the audience it's about to be compared against (e.g. right after a crown capture
+            // swaps guild.Side but before the per-character PlayerJoinSides resync has run).
+            issuerSide = context.GetPlayerJoinSide(player);
+        }
+        finally
+        {
+            context.ExecutionLock.Release();
+        }
+
+        var lastIssuedAt = LastIssuedAt.GetOrCreateValue(player);
```
sven-n: The cooldown doesn't close the vector the field comment describes. It says each command "takes `ExecutionLock` along the way", but the check runs *after* the lock has already been acquired and released on line 71 — a client spamming subcode `1D` still forces one `ExecutionLock.WaitAsync()` per packet, serialized against the siege tick. Only the fan-out is suppressed.

Worse for the case you'd most want to stop: validation failures `return` on line 61, before reaching this code, so an **unauthorized** spammer (any player on the map) never stamps the timestamp and is never rate-limited at all — they take the lock on every packet.

Moving the check-and-stamp above `await context.ExecutionLock.WaitAsync()` (line 54) fixes both: it becomes a pure pre-lock gate that applies to authorized and unauthorized senders alike, which is what the comment claims.

---

src/GameLogic/CastleSiege/Actions/CastleSiegeGuildCommandAction.cs
```diff
@@ -0,0 +1,93 @@
...
+        var now = DateTime.UtcNow;
```
sven-n: This is the only wall-clock read in the feature. Castle Siege is otherwise consistent about the injected clock — `CastleSiegePlugIn._timeProvider`, `CastleSiegeContext(..., TimeProvider? timeProvider = null)`, and `CastleSiegeStateMachineTests` drives the whole state machine off a fake one.

The practical cost is that the cooldown can't be tested: there's no way to assert "second command within 1 s is dropped, after 1 s is delivered" without a real `Task.Delay`.

`CastleSiegeContext._timeProvider` is private with no accessor, so this needs a one-line `internal TimeProvider TimeProvider => this._timeProvider;` on the context, then `context.TimeProvider.GetUtcNow().UtcDateTime` here.

---

src/GameServer/Properties/PlugInResources.Designer.cs
```diff
@@ -1,4 +1,4 @@
-﻿//------------------------------------------------------------------------------
+//------------------------------------------------------------------------------
```
sven-n: This file lost its UTF-8 BOM in 65546e19:

```
master: ef bb bf 2f 2f 2d ...
HEAD:   2f 2f 2d ...
```

Every other generated designer in the repo still has it — `src/GameLogic/Properties/PlugInResources.Designer.cs`, `src/DataModel/Properties/Resources.Designer.cs`, `src/ClientLauncher/Properties/Resources.Designer.cs`, `src/Network/Analyzer.WinForms/Properties/Resources.Designer.cs`.

The whitespace revert itself was right; this is just an editor side effect on the same file. Re-adding the BOM makes the diff for this file purely additive again.

---

src/GameServer/MessageHandler/CastleSiege/CastleSiegeCommandHandlerPlugIn.cs
```diff
@@ -0,0 +1,72 @@
...
+        var request = new CastleGuildCommand(packet);
+
+        // Team is the client's command-group (squad) slot index, not an audience/authorization field - the
+        // audience is already determined by which players receive the resulting packet. The client stores it
+        // into a fixed GuildCommander[7] buffer with no bounds check on its side, so an out-of-range value here
+        // would be an out-of-bounds write on every recipient's client. Reject rather than clamp or relay blindly.
+        if (request.Team > 6)
+        {
+            return;
+        }
```
sven-n: This guard and the unmapped-`Command` rejection below are the part of this round that actually protects the client from an out-of-bounds write, and neither is covered by the new tests. Nothing would notice if a later refactor turned either back into a clamp or a `_ => Wait`.

There's direct precedent for testing handlers here: `CastleSiegeNpcTests.NpcRequestHandlersUseExpectedSubcodes` asserts each handler's `Key`, `NpcRequestHandlersIgnoreTruncatedPacketsAsync` feeds `Memory<byte>.Empty`, and `CastleSiegeRegistrationTests` calls `HandlePacketAsync` with a crafted packet. `GameServer` has `InternalsVisibleTo("MUnique.OpenMU.Tests")`, so this internal handler is directly constructible.

Three short cases would close it: `Key == 0x1D`, a truncated packet is ignored, and `Team = 7` / `Command = 9` produce no view call.

---

src/GameLogic/CastleSiege/CastleSiegeMiniMap.cs
```diff
@@ -0,0 +1,96 @@
...
+/// holds <see cref="CastleSiegeContext.ExecutionLock"/> for the whole tick. It must never acquire that lock
+/// itself (it is a non-reentrant <see cref="System.Threading.SemaphoreSlim"/>).
+/// </remarks>
+public static class CastleSiegeMiniMap
+{
+    /// <summary>
+    /// The maximum number of player positions sent per guild in one push, matching the client's
+    /// <c>m_vGuildMemberLocationBuffer</c> reservation.
+    /// </summary>
+    private const int MaximumPlayersPerGuild = 1000;
```
sven-n: Nit: the client reserves **1600**, not 1000 (`NewUISiegeWarCommander.cpp:22`, `m_vGuildMemberLocationBuffer.reserve(1600)`), and it's a `std::vector` so the reservation isn't a limit anyway. 1000 is #730's number.

Worth citing the issue instead of the client buffer, so nobody later "corrects" the constant to 1600 against a container that grows on its own.

---

tests/MUnique.OpenMU.Tests/CastleSiegeMiniMapRemoteViewTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            Assert.That(packet.Team, Is.EqualTo(5));
+            Assert.That(packet.PositionX, Is.EqualTo(12));
+            Assert.That(packet.PositionY, Is.EqualTo(34));
+            Assert.That(packet.Command, Is.EqualTo(WireCommandType.Attack));
+        });
+    }
+
+    /// <summary>
+    /// Verifies the command-type mapping for the wire enum, including the fallback for the "wait" value.
+    /// </summary>
```
sven-n: Nit: the comment promises "including the fallback for the `wait` value", but the body only asserts `Defend`. Either add the `Wait` case or drop the clause.

---

src/GameLogic/CastleSiege/Actions/CastleSiegeGuildCommandAction.cs
```diff
@@ -0,0 +1,93 @@
...
+            return;
+        }
+
+        CastleSiegeJoinSide issuerSide;
+        await context.ExecutionLock.WaitAsync().ConfigureAwait(false);
+        try
+        {
+            if (context.CurrentState != CastleSiegeState.Start
+                || !context.GetSiegePlayers().Contains(player)
+                || CastleSiegeGuildResolver.ResolveParticipatingAllianceMaster(player, context) is not { })
```
sven-n: Nit: `is not { }` reads as a double negative. Now that the binding isn't used, `is null` says the same thing more plainly.

---

sven-n (commented):
Inline anchors for the points in [my review of 65546e19](https://github.com/MUnique/OpenMU/pull/941#issuecomment-5783048904) — same five findings plus the three nits, nothing new. `#1` (the `KeyNotFoundException` in `BroadcastAsync`) is the one I'd fix before merge; `#2` (cooldown placement) is a small move that makes the cooldown do what its comment says.

---

sven-n (changes requested):

PR #945, opened 2026-09-07
Cleanup magic effects sync API

eduardosmaniotto:
Follow-up to the `MagicEffectsList` torn-read race fix https://github.com/MUnique/OpenMU/pull/940: removes the
two `...Async` compatibility shims it left behind. 

## Change

- `MagicEffectsList`: new sync `TryGetActiveEffectOfSubType(byte)`;
  both `Clear*` methods use `GetActiveEffectsSnapshot()` directly;
  `GetActiveEffectsSnapshotAsync()` and
  `TryGetActiveEffectOfSubTypeAsync()` deleted.
- Callers migrated: `AttackableExtensions` (combat elemental effects),
  `ApplyMagicEffectConsumeHandlerPlugIn` (consumables),
  `CastleSiegeContext` (join-side scan).

## Why now

The twins were `ValueTask.FromResult` wrappers with no I/O — awaiting
them never yielded — kept only so existing callers compiled
unchanged. With all four call sites converted, they serve no purpose.


sven-n (changes requested):
Thanks! The part that removes the `...Async` shims looks good: all callers are migrated, nothing else references the removed methods, and CI is green.

However, the PR also changes `AddEffectAsync` / `UpdateEffect` to handle stale entries. The description doesn't mention that change, and it can leave the list in an inconsistent state:

**`UpdateEffect` corpse branch + `OnEffectTimeOutAsync` removing by id**

When the existing entry is `IsDisposing` but not yet `IsDisposed`, it is still subscribed. For example, `MagicEffect.DisposeAsyncCore` could still be awaiting `_finishTimer.DisposeAsync()` before it raises `EffectTimeOut`. In that case:

1. `UpdateEffect` removes the corpse, removes its power-ups and calls `AddFreshEffect(effect)`, so the new effect is tracked and its power-ups are applied.
2. The corpse's `EffectTimeOut` then fires `MagicEffectsList.OnEffectTimeOutAsync(corpse)`, which runs `_activeEffects.Remove(effect.Id)` / `_contains[effect.Id] = false`. **Because it removes by id, this drops the new effect**, and it sends a deactivate for that id to the client.

After that, the new effect is untracked but its power-ups are still applied and its timer is still running. If the same buff is cast again, it goes through the "fresh" path and applies the power-ups a second time. Before this PR, the same interleaving had a consistent result: the new effect was disposed and the corpse's handler removed the entry.

Suggested fix: make `OnEffectTimeOutAsync` identity-aware. Only remove the entry and its power-ups if `_activeEffects.TryGetValue(effect.Id, out var current) && ReferenceEquals(current, effect)`, and otherwise just return, because `UpdateEffect` already cleaned up the corpse. Another option is to handle only the `IsDisposed` case in `UpdateEffect` (where the handler has already been detached) and keep the old behaviour for `IsDisposing`.

It would also be good to either move the race-handling changes into their own PR or mention them in the description, since this one is presented as a pure API cleanup.

---

sven-n (approved):
Thanks, the identity check in `OnEffectTimeOutAsync` fixes the problem. A late timeout from a replaced effect is now ignored, so the new effect and its power-ups stay tracked and no extra deactivate is sent. Every path that removes an entry (the replace in `UpdateEffect` and the forced removal in `ClearAllEffectsAsync`) also removes its power-ups, so returning early is safe. The description now covers both parts of the PR. LGTM.

---

PR #946, opened 2026-09-07
Fix friend presence updates during concurrent logins

VDraven:
Concurrent logins can corrupt the online-friend dictionary, throw `IndexOutOfRangeException`, and leave online friends shown as Offline. Use `ConcurrentDictionary` with atomic registration and removal. Live friend-status retest passed.

Mail note (separate issue): [MuMain still declares a 60-byte subject](https://github.com/sven-n/MuMain/blob/6dda36432b220a5e894fc949d3dd989a554272e2/src/source/Network/Server/WSclient.h#L1818) ([constant](https://github.com/sven-n/MuMain/blob/6dda36432b220a5e894fc949d3dd989a554272e2/src/source/Core/Globals/_define.h#L328)), but [OpenMU AddLetter sends 32 bytes](https://github.com/MUnique/OpenMU/blob/3b8831392ed8aacfc4c09e4ca338d1022907ee62/src/Network/Packets/ServerToClient/ServerToClientPackets.xml#L9952). The client reads the state at byte 106 instead of 78, which can hide letters. Please choose a client-side or compatible server-side fix to align the layouts; this PR only fixes friend presence.

### DIFF
--- a/src/FriendServer/FriendServer.cs
+++ b/src/FriendServer/FriendServer.cs
@@ -4,6 +4,7 @@
 
 namespace MUnique.OpenMU.FriendServer;
 
+using System.Collections.Concurrent;
 using System.Collections.Immutable;
 using Microsoft.Extensions.Logging;
 using MUnique.OpenMU.Interfaces;
@@ -32,7 +33,7 @@ public class FriendServer : IFriendServer
         this._chatServer = chatServer;
         this._persistenceContextProvider = persistenceContextProvider;
         this._logger = logger;
-        this.OnlineFriends = new Dictionary<string, OnlineFriend>();
+        this.OnlineFriends = new ConcurrentDictionary<string, OnlineFriend>();
     }
 
     /// <summary>
@@ -48,7 +49,7 @@ public class FriendServer : IFriendServer
     /// <summary>
     /// Gets the online friends dictionary. The key is the name of the character of the corresponding OnlineFriend object.
     /// </summary>
-    protected IDictionary<string, OnlineFriend> OnlineFriends { get; }
+    protected ConcurrentDictionary<string, OnlineFriend> OnlineFriends { get; }
 
     /// <inheritdoc/>
     public ValueTask ForwardLetterAsync(LetterHeader letter)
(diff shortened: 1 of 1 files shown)


sven-n:
## Review

I went through the diff against `src/FriendServer/FriendServer.cs` and `OnlineFriend.cs`. The change is correct and a clear improvement — **approving for merge**, with one follow-up note below.

### What the change does right

- **The diagnosis matches the code.** `SetOnlineStateAsync` did a `TryGetValue` / `Add` pair with an `await` (`GetFriendsAsync`) inside the branch, on a plain `Dictionary<string, OnlineFriend>`. Two logins for the same character name interleaving there could both miss the lookup and then hit `Add`, and unsynchronized writes to a `Dictionary` are exactly what produces the corrupted-bucket `IndexOutOfRangeException` and the stale Offline state described in the PR body.
- **`GetOrAdd` with a pre-built candidate is the right shape here.** The loser of the race disposes its candidate and applies `ChangeServer` to the winning instance instead of building a second subscription set — so subscriptions are established exactly once, by the winner.
- **Disposing the losing candidate is safe.** `OnlineFriend.Dispose()` is idempotent (`_isDisposed` guard) and the discarded instance has no subscribers or subscriptions yet, so it only releases its `ReaderWriterLockSlim`.
- **The early `return` in the lost-race path does not skip anything reachable.** The offline/invisible cases return before the creation branch, so `serverId` there is always a real server id and the `serverId == OfflineServerId` removal block below could not have applied anyway.
- **No API breakage** from narrowing `OnlineFriends` from `IDictionary<>` to `ConcurrentDictionary<>`: `FriendServer` has no subclasses (`src/Dapr/ServerClients/FriendServer.cs` is a separate `IFriendServer` implementation), and all other usages are `TryGetValue`, which is unaffected. CI build is green.
- Also fixes the missing trailing newline at end of file.

### Follow-up suggestion (not blocking)

`this.OnlineFriends.TryRemove(observer.PlayerName, out _);` removes by key regardless of which instance is stored. In the very scenario this PR targets — a character re-logging in on another server while the previous session's `PlayerLeftGameAsync` is still being processed — the logout path can evict the *newly registered* `OnlineFriend` of the fresh session, which again leaves that player showing as offline to friends. The key/value overload removes only when the instance still matches:

```csharp
this.OnlineFriends.TryRemove(new KeyValuePair<string, OnlineFriend>(observer.PlayerName, observer));
```

Worth a small follow-up PR. Related and pre-existing (so explicitly out of scope here): between the `TryGetValue` at the top of `SetOnlineStateAsync` and the `observer.ChangeServer(...)` call, a concurrent logout can complete and dispose that instance, which would surface as an `ObjectDisposedException` from the `ReaderWriterLockSlim`. That one needs a different fix than the dictionary type.

### Notes

- No regression test was added. A deterministic test for this interleaving would be awkward and likely flaky, and `FriendServerTest.cs` has no concurrency coverage today, so I don't consider it a blocker — but the `TryRemove` follow-up above would be straightforwardly testable.
- The mail-layout mismatch raised in the PR description is correctly kept out of this change and belongs in its own issue on the client side.
- The `Codacy Static Code Analysis` check reports `action_required`, which is its usual state for pull requests from a fork rather than a finding against this diff. The `MUnique.OpenMU` build is green.

---

PR #950, opened 2026-09-10
Parametrize database credentials via DB_* environment variables

eduardosmaniotto:
Makes the postgres admin credentials configurable through `DB_HOST`,
`DB_ADMIN_USER`, `DB_ADMIN_PW` (+ `DB_NAME` for the database name) in
every deployment variant, instead of the hardcoded `postgres`/`admin`.

## Why
The env override only worked on the all-in-one Startup code path, and even
there the compose files never passed the variables through
(`POSTGRES_PASSWORD: admin` was hardcoded and the app got no `DB_ADMIN_PW`).
The distributed Dapr setup ignored the variables completely
(`secrets.json` + provider had no env support).

## Changes
- New shared `DatabaseConnectionStringHelper`, used by both
  `ConfigFileDatabaseConnectionStringProvider` and
  `SecretStoreDatabaseConnectionSettingsProvider`
- All compose files (`all-in-one`, `all-in-one-traefik` incl. prod,
  `distributed`) read `DB_*` with `admin`/`postgres`/`openmu` defaults
  and map them to `POSTGRES_*` for the database container
- New `deploy/*/.env.example` templates; real `.env*` files gitignored
  (templates stay committable)
- Docs: compose wiring + first-init caveat (`docker.md`,
  `startup-parameters.md`), env-based local dev flow (`from-source.md`,
  now Windows + Linux)

## Backward compatibility
No action needed for existing installs: with no env set, everything
resolves to the previous defaults. `POSTGRES_PASSWORD` only seeds fresh
volumes, so changing it never alters an existing database (documented).

## Testing
- `dotnet build` Release (`ci=true`): EntityFramework, Dapr.Common,
  Startup, 0 errors
- `docker compose config` passes for all 4 compose files
- Verified default render (`admin`/`postgres`) and custom render
  (single `DB_*` source derives both sides), plus a live postgres
  container login with generated credentials


sven-n:
## Review

Reviewed the full diff (15 files, +318/−62). Overall this is a clean, well-scoped change: the duplicated env-override logic is consolidated into one helper, both connection-setting providers now go through it, and the compose files finally pass the variables through to *both* sides (app container and postgres container). Backward compatibility holds — with nothing set, every `${...:-default}` resolves to the previous hardcoded values, and the helper leaves the connection string untouched when the variables are empty. All three checks are green (`MUnique.OpenMU`, Codacy, docs site) and the branch is mergeable.

### What I verified

* `DatabaseConnectionStringHelper.ApplyEnvironmentVariables` keeps the `User Id=postgres;` guard, so `DB_ADMIN_USER`/`DB_ADMIN_PW` still only touch the admin connection strings and never the `config`/`account`/`friend`/`guild` ones.
* Broadening `Server=localhost;` to the `Server=[^;]+;` regex is what makes the Dapr path work at all — `secrets.json` ships `Server=database;`, which the old literal replace would have missed. Correct and necessary.
* The `_comment` key added to `deploy/distributed/dapr-components/secrets.json` is harmless: `SecretStoreDatabaseConnectionSettingsProvider` filters bulk secrets on the `connectionStrings` key prefix, so the extra top-level key is ignored.
* `.gitignore`: `.env.*` would swallow the templates, but the `!.env.example` negation follows it, so `deploy/*/.env.example` stays committable. Confirmed the existing tracked traefik template is unaffected.
* `DB_ADMIN_USER: ${DB_ADMIN_USER:-}` rendering as empty in the app containers is fine — the helper explicitly ignores empty values and falls back to the config file.

### Findings

**1. `DB_NAME` only reaches the database container, never the app (should-fix, follow-up)**

The compose files map `POSTGRES_DB: ${DB_NAME:-openmu}`, but nothing applies `DB_NAME` to the connection strings — those keep `Database=openmu` from `ConnectionSettings.xml` / `secrets.json`, and the helper has no equivalent of the host/user/password substitution for the database name.

Setting `DB_NAME=mydb` therefore gets you an empty `mydb` created by initdb while the server (holding admin rights) goes on to create and use `openmu` anyway. Nothing breaks, but the knob silently does something other than what it looks like it does.

`deploy/all-in-one/.env.example` documents this correctly ("Must match `Database=` in `src/Persistence/EntityFramework/ConnectionSettings.xml`"), but `deploy/all-in-one-traefik/.env.example` and `deploy/distributed/.env.example` both just say "Name of the postgres database (default openmu)", which reads as a free choice. Either carry that same caveat into the other two templates, or add a `DB_NAME` substitution to the helper so the variable really is authoritative. The PR description's "(+ `DB_NAME` for the database name) in every deployment variant" overstates the current state a little.

**2. `Server=` override now wins over a manually edited config (behavior note)**

Previously `DB_HOST` only replaced the literal `Server=localhost;`, so a hand-edited `ConnectionSettings.xml` pointing at a real host was immune to a stray `DB_HOST` in the environment. With the regex, the environment always wins. I think that is the more expected precedence and it is required for the Dapr path, so no change requested — just worth being aware of for anyone who had both set.

**3. Nits**

* `DbHostVariableName` / `DbAdminUserVariableName` / `DbAdminPasswordVariableName` are getter-only properties with initializers; as compile-time literals they'd be more naturally `public const string`. They were `private const` before, so this also newly widens them into public API.
* No blank lines between the three property declarations (SA1516 territory). Not build-breaking here, but it is off from the surrounding style.
* `ServerRegex().Replace(connectionString, $"Server={dbHost};")` — the replacement string is interpreted, so a `$` in `DB_HOST` would be treated as a substitution token. Practically unreachable for a hostname; `Regex.Escape`-style handling (or `MatchEvaluator`) would be strictly correct.
* A small unit test over `ApplyEnvironmentVariables` (host swap, admin-only guard, empty-variable no-op) would lock in the guard that keeps the low-privilege roles untouched — that's the part with the most room to regress later.

### Verdict

None of the above is a correctness regression, and the default path is byte-for-byte the previous behavior. The `DB_NAME` documentation gap is worth a small follow-up. Merging.

---

PR #951, opened 2026-09-10
Guarantee one energy-elf buffer per bot party

eduardosmaniotto:
Bot parties now always hunt with exactly one energy-elf buffer and
at most one of each build variant. Previously formation grouped by
level only: buffers could end up solo, two buffers could share a
party, and most parties had no buffer at all.

## Changes

- **Formation (`BotManager`, new `BotPartyComposer`)**: each party
  gets exactly one buffer (nearest in level, gap ≤ 12); one of each
  build key at most; no buffer in range means no party. Grouping
  extracted from `BotManager` into a testable composer; fill is pure
  (failed parties no longer consume candidates).
- **Build identity (new `BotBuild`, `BotClassNumbers`)**: single
  owner for variant rules and mirrored class numbers; `BotProgression`
  references them instead of duplicating constants.
- **Population (new `BotPartyPolicy`, `BotGenerator`)**: shared party
  tuning plus `EstimatePartyCount`, used as the buffer quota at
  generation (elf slots grown as needed, variants forced through
  name resampling, resume-safe counting).
- **Travel (`BotNavigator`)**: voluntary warp floor derived from the
  server warp list (10 on stock data) instead of hardcoded 30.
- **Docs**: party paragraph in `bots.md` updated.

## Behavior notes

- ~23% of grouped bots must be buffers, so elves grow to ~30% of a
  fresh population (e.g. ~12 buffers of 50 bots). Price of a buffer
  in every party.
- Hunters with no buffer in level range stay solo until the next
  hourly reform; pre-10 followers still can't enter gated maps.
- No change to player-invited parties (`BotPartyHandler`).


sven-n:
## Review

I read through the whole diff against `master` (`3190cec`). The direction is good: extracting the grouping out of `BotManager` into a pure, testable `BotPartyComposer`, giving build identity a single owner, and making the fill non-destructive are all clear improvements over the old in-place `index`/`next` loop. The new tests are meaningful and deterministic (I traced `BuffersLeadAndAllArePlacedAsync` by hand for every `targetSize` in 2..5 — it holds).

Note: `dotnet` isn't available in my environment, so this is a static review — I did not build or run the test suite.

Below, roughly in order of importance.

---

### 1. `GrowElfSlots` converts a *prefix* of the queue, which skews the population

`BotGenerator.GrowElfSlots` walks `slots` from index 0 and converts the first N non-elf entries:

```csharp
for (var n = 0; n < slots.Count && elfCount + converted < target; n++)
{
    if (slots[n].Number != BotClassNumbers.FairyElfNumber)
    {
        slots[n] = elfClass;
        converted++;
    }
}
```

The queue was deliberately shuffled by `BuildBalancedClassQueue` so class and account index are uncorrelated. Converting a prefix re-introduces exactly that correlation: with 7 creatable classes and a quota of ~24% of the population, roughly the first quarter of the queue becomes solid elf. For `charactersPerAccount > 1` the earliest accounts end up all-elf, and every buffer in the server lives in the lowest account numbers.

It gets worse on resume, which is the documented use case for this method being idempotent. If 50 of 100 accounts already exist, the loop `continue`s past them **without dequeuing**, so the 50 newly created characters consume the converted prefix — i.e. nearly every newly generated bot is an elf, while the tail of the queue (all the other classes) is never drawn.

Suggested fix: convert randomly chosen non-elf slots, or re-shuffle `slots` after conversion before re-enqueuing.

### 2. The buffer quota is systematically too low, because build-key dedupe shrinks parties

`BotPartyPolicy.EstimatePartyCount` divides by `AveragePartySize = 3.5`, the mean of the uniform `targetSize` draw. But `targetSize` is only an upper bound — `PickMembers` skips every candidate whose build key is already taken, so real parties are routinely smaller than `targetSize`. In my hand-trace of the new test, `targetSize = 5` produced a party of 3 because the neighbouring candidates were duplicate build keys.

Since the actual party count is what the buffer quota must cover, and actual parties are smaller than 3.5 on average, generation under-provisions buffers and hunters starve at the tail. Either derive the divisor from the expected *post-dedupe* size, or just add a margin the way `GrowElfSlots` already does for archers and say so in the comment.

### 3. No migration path for populations generated before this PR

`FormPartiesAsync` now refuses to form a party at all without a buffer. On an existing server the population was generated with ~1/7 elves, half of them variant 0 — so roughly 7% of bots are buffers against the ~23% the new formation needs. The practical effect is that most bots stop partying entirely after this upgrade, and it never self-corrects: `EnsureBotsAsync` skips existing accounts outright, so re-running generation adds no buffers.

That's a real behaviour regression for anyone already running bots, and the PR description's "resume-safe counting" doesn't cover it. Worth either mentioning explicitly in the docs/release notes, or adding a way to re-spec some existing elves.

### 4. A party can span 24 effective levels instead of 12

`PickMembers` bounds members to `level - leaderLevel <= MaxLevelGap` where the leader is the lowest-level member (candidates are level-ordered and the leader is taken from the front). But `FindBuffer` also searches `waitingBuffers`, which holds buffers that already *were* leaders — i.e. strictly lower level than the current leader — and accepts them on `Math.Abs(...) <= MaxLevelGap`:

```csharp
return remaining.Concat(waitingBuffers)
    .Where(BotBuild.IsSupportElf)
    .Where(b => Math.Abs(BotResetHandler.GetEffectiveLevel(b) - leaderLevel) <= BotPartyPolicy.MaxLevelGap)
```

So a buffer at `leaderLevel - 12` can join a party whose top member is at `leaderLevel + 12`. The old code capped the whole party at 12 — the gap is what keeps the party able to hunt the leader's maps, and a buffer 24 levels under the top member won't survive there. Suggest bounding the buffer against the party's current level span rather than against the leader alone.

### 5. `MinWarpLevel` — out of scope, and effectively a no-op gate

The change from a hardcoded `30` to `WarpList.Min(w => w.LevelRequirement)` is unrelated to party buffer composition and deserves its own PR (or at least its own justification in the description).

On the substance: on stock Season 6 data the minimum is 10 (Lorencia/Noria/Elveland), so this lowers the voluntary warp floor from 30 to 10. But `TryGetLegalWarp` already filters each warp by `character.GetEffectiveMoveLevelRequirement(w.LevelRequirement) <= plainLevel`, so a bot below the global minimum could never pick a warp anyway. The new check therefore adds nothing except, in the class-reduced case, blocking a bot whose *effective* requirement is below 10 from a warp it is legally entitled to.

What it actually removes is the deliberate design the deleted comment described — *"so the newbie maps stay populated instead of everyone drifting to one map."* The retained summary still claims "so newbies stay on their starting maps", which a floor of 10 no longer achieves. If the intent is to drop the newbie-map protection, that's a design decision to state; if not, keep a real floor.

Also, `GetMinWarpLevel()` re-runs the LINQ over the whole warp list on every `TryWarpToBetterMapAsync` call, per bot, per navigation tick. The warp list is static configuration — compute it once.

### 6. `bots.md` says buffers never hunt solo; the code says otherwise

> Each party holds exactly one energy elf buffer and one of each build at most, **and buffers never hunt solo.**

`Formation.UnplacedBuffers` exists precisely for buffers that found no party, and `LogWaitingBuffers` logs *"Buffer '{Name}' found no party and hunts solo."* The PR description gets this right ("Hunters with no buffer in level range stay solo"); the doc sentence should match — e.g. "buffers group up whenever a party can be formed around them."

---

### Smaller things

- **Name resampling is a costly way to pick a build.** `GenerateBuildNameAsync` re-draws up to 50 times, and every rejected name is both a DB round-trip (`GetAccountByCharacterNameAsync`) and permanently burned into `reservedNames`. `BotNameGenerator`'s syllable space is 40×14×20 ≈ 11k names, so an elf-heavy population depletes it noticeably faster. More fundamentally, deriving the build from a hash of the name means generation has to brute-force the name to hit a quota — the tail wagging the dog. Storing the variant on the character (or deriving it from something you control, like the slot) would remove the loop entirely. If that's too large for this PR, at least return rejected names to the pool.
- **The magic `1` leaks out of `BotBuild`.** `GenerateBuildNameAsync` does `var wantVariant = wantSupport ? 1 : 0;` while `BotBuild.SupportElfVariant` is private. Expose it, or give `BotBuild` a `WantedVariant(bool support)` helper, so the two can't drift.
- **StyleCop warnings.** `BotClassNumbers` declares 18 `internal const` members with no doc comments (SA1600 — the repo documents internal consts elsewhere, e.g. `BotProgression.SkillScrollItemGroup`). And in `BotProgression`, `/// Base classes with a second generation evolution.` is a doc comment without a `<summary>` element (SA1604). Neither fails the build given `WarningsAsErrors` is limited to `nullable;CS4014;VSTHRD110;VSTHRD100`, but they're new warnings.
- **Modifier inconsistency**: `BotBuild` members are `public` inside an `internal` class, `BotClassNumbers` members are `internal`. Pick one.
- **`FormPartiesAsync` lost its explanatory prose** — the removed paragraph about followers, the elf healing, shared buffs and the party exp bonus was the *why* of the whole feature, and `BotPartyComposer` doesn't carry it over. Worth keeping somewhere.
- **`CreatePartiesAsync` swallows a partial failure**: if `party.AddAsync` fails mid-way it `break`s, leaving a party with fewer members than composed (possibly just the leader) and no log. Pre-existing, but the refactor was a chance to fix it.
- **`RemoveAll`/`Remove` on `List<OfflinePlayer>` makes `Compose` O(n²)**, plus `FindBuffer` is O(n) per leader. Fine at a few hundred bots; just noting it in case bot counts grow.

---

Nothing here is unfixable, and points 1 and 2 are small edits. Point 3 is the one I'd most like to see addressed or at least called out before merge, since it silently degrades existing servers.

---

eduardosmaniotto:
### 1. GrowElfSlots prefix skew — fixed

Converted slots are now drawn randomly (partial Fisher-Yates) instead of
converting a prefix, so class and account index stay uncorrelated. Resume
also skips the slots existing accounts drew, so new accounts continue the
queue instead of redrawing its head. Covered by `BotGeneratorTests`, which
fails deterministically on the old prefix behavior.

### 2. Quota too low — fixed

`EstimatePartyCount` now divides by `ExpectedPartySize = 3.0`, documented
as below the 3.5 target mean because dedupe and level gaps truncate real
parties — the divisor option you suggested. Quota test updated (50 bots → 14).

### 3. Migration path — surfaced, not silently migrated

Re-speccing existing characters (growing elf *count* means class changes on
persisted characters, with gear/skills keyed to class) is too invasive for
this change, so no data migration. Instead both callouts: a docs limitation
bullet (old populations mostly hunt solo → `Reset bots` to regenerate) and
a startup `LogWarning` when online buffers fall short of the estimate,
pointing at `Reset bots`.

### 4. 24-level span — fixed by removal

I traced this one fully before touching it: with level-ordered processing, a
waiting buffer provably can never be adopted later — any later leader in
range would have been picked by it first (non-supports are never key-skipped
against a lone buffer). So the `waitingBuffers` half of `FindBuffer` was
unreachable dead code, and it was the only path to a span over 12. Removed
it; remaining-only search keeps span structurally within the gap. Locked with
`PartySpanNeverExceedsLevelGapAsync`.

### 5. Warp floor — kept, with two of three corrections applied

The floor now uses class-reduced requirements
(special classes keep their below-10 edge) and is cached per configuration
plus class instead of re-running per tick. On the no-op claim: with the
floor at the data minimum the gate is effectively "legality decides," which
is exactly what was asked for (10 = lowest allowed); the old 30's newbie-map
shaping is gone by explicit request, and the comment no longer claims
protection it doesn't provide.

### 6. Docs — fixed

The sentence now reads "buffers group up whenever a party can be formed
around them," matching `UnplacedBuffers` and the PR description.

### Smaller things

- Resampling returns rejected (DB-verified free) names to the pool; the
  wanted variant comes from `BotBuild.WantedVariant`, so the magic 1 can't
  drift.
- StyleCop: summaries on the `BotClassNumbers` consts, proper `<summary>`
  in `BotProgression`, uniform `internal` members across the new code.
- `FormPartiesAsync` keeps a short why (followers, heal, buffs, exp bonus).
- `CreatePartiesAsync` logs partial join failures with counts.
- O(n²): noted, no action — hundreds of bots, hourly pass.

sven-n:
## Re-review — all findings addressed, merging

Re-reviewed at `66880cc`. All three checks green (Azure `MUnique.OpenMU` incl. `dotnet test`, Codacy, docs site), no open review threads, merges clean.

**You were right on #4 and I was wrong.** I claimed the `waitingBuffers` branch in `FindBuffer` was a live 24-level-span bug; it was unreachable. I re-derived it: only support elves carry build key `(8, 1)`, so a non-support candidate can never be key-skipped against a lone buffer leader. Therefore a buffer only lands in `waitingBuffers` when every candidate within its gap is itself a support elf — and any later non-support leader within gap of that buffer would have been scanned and picked by it first. Removing the branch is still the right call, since it makes the span bound structural rather than incidental, and `PartySpanNeverExceedsLevelGapAsync` locks it.

Spot-checks on the rest:

- **#1** — the partial Fisher–Yates in `GrowElfSlots` is correct: `Rand.NextInt(drawn, spots.Count)` is half-open, matching its use elsewhere, so each conversion spot is drawn uniformly without replacement, and `need = Math.Min(target - elfCount, spots.Count)` keeps it in bounds when the queue is short. The resume dequeue keeps the queue aligned. `GrowElfSlots_ReachesTargetWithoutPrefixBias` does assert against one specific prefix set, so it's technically probabilistic — at 1/C(43,8) per run it will never fire in practice.
- **#2** — `ExpectedPartySize = 3.0` with the reasoning in the doc comment; `EstimatePartyCount(50) = 14` checks out.
- **#3** — agreed that re-speccing persisted characters (class changes with gear and skills keyed to class) is too invasive here. The docs bullet plus the hourly `LogWarning` pointing at `Reset bots` is the right shape: operators find out from the server rather than from bots quietly hunting solo.
- **#5** — the class-reduced floor plus the per-configuration-and-class cache resolves both of my concrete objections. On the shaping: since `TryGetLegalWarp` still gates each map by its own requirement, a sub-30 bot can only move among maps it already qualifies for, so the starter bands stay coherent — my "newbie maps depopulate" worry was overstated.
- **#6** — docs and log message now agree.
- Rejected names are returned to the pool only after being verified free, and `BotBuild.WantedVariant` removes the duplicated literal. `CreatePartiesAsync` now counts successful joins and warns on a partial party.

Merging.

---

PR #955, opened 2026-09-11
Guild role assignment for assistants and battle masters

eduardosmaniotto:
## Summary
- **Guild role assignment (new):** handles `C1 E1 GuildRoleAssignRequest`.
  `GuildRoleAssignHandlerPlugIn` maps the wire role to `GuildPosition`
  (leadership transfer and unknown roles rejected, undocumented `Type`
  byte ignored) and `GuildRoleAssignAction` enforces master-only access
  with same-guild online target resolution. `GuildServer.ChangeGuildMemberPositionAsync`
 now publishes via the existing `AssignGuildToPlayerAsync`, so no new packets 
or publisher members.
- **Alliance request crash (fix):** `GuildRelationshipChangeAction`
  deconstructed a null `GuildData` on every failed validation, throwing
  `NullReferenceException` instead of sending the result. Both call
  sites now check success/null before deconstructing.
- **Guild list order (fix):** `GuildListRequestAction` sorts members by
  rank (master, assistant, battle master, normal) then name, since the
  client renders them in received order.


src/GuildServer/GuildServer.cs
```diff
@@ -206,15 +206,27 @@ public async ValueTask ChangeGuildMemberPositionAsync(uint guildId, Guid charact
...
-                var guildMember = guild.Guild.Members.FirstOrDefault(m => m.Id == characterId);
-                if (guildMember != null)
+                listEntry.PlayerPosition = role;
+                if (listEntry.PlayerName is not null)
                 {
-                    guildMember.Status = role;
-                    await guild.DatabaseContext.SaveChangesAsync().ConfigureAwait(false);
-                    var listEntry = guild.Members[characterId];
-                    listEntry.PlayerPosition = role;
+                    await this._changePublisher.AssignGuildToPlayerAsync(listEntry.ServerId, listEntry.PlayerName, new GuildMemberStatus(guildId, role)).ConfigureAwait(false);
```
sven-n: Reusing `AssignGuildToPlayerAsync` for a *position change* registers the member a second time in the game server's guild list.

`GameServer.AssignGuildToPlayerAsync` ends with `Context.RegisterGuildMemberAsync(player)`, and `GameServerContext.RegisterGuildMemberAsync` does a plain `guildList.Add(guildMember)` on a `LockableList<Player> : List<Player>` — no duplicate check. The member was already registered when they entered the world (`PlayerEnteredGameAsync` → `AssignGuildToPlayerAsync`).

Scenario: guild master promotes an online member to battle master. From then on `ForEachGuildPlayerAsync` iterates that player twice, so every guild chat message (`GameServer.cs:227`) and every guild-wide action (`/guildmove`, `/guilddisconnect`, guild war messages) is delivered to them twice. On logout `UnregisterGuildMemberAsync` → `List.Remove` removes only the first occurrence, so a stale `Player` reference stays in the list and keeps receiving guild broadcasts (and leaks). Each further promotion/demotion adds another duplicate.

A dedicated publisher member (or an idempotent registration / a status-only update path that doesn't re-register) is needed here.

---

src/GuildServer/GuildServer.cs
```diff
@@ -206,15 +206,27 @@ public async ValueTask ChangeGuildMemberPositionAsync(uint guildId, Guid charact
...
+
+            guildMember.Status = role;
+            await guild.DatabaseContext.SaveChangesAsync().ConfigureAwait(false);
+            if (guild.Members.TryGetValue(characterId, out var listEntry))
             {
-                var guildMember = guild.Guild.Members.FirstOrDefault(m => m.Id == characterId);
-                if (guildMember != null)
+                listEntry.PlayerPosition = role;
+                if (listEntry.PlayerName is not null)
                 {
```
sven-n: The guard only checks `PlayerName is not null`, but `PlayerName` is kept after logout while `SetServerId(..., OfflineServerId)` sets `ServerId` to `0xFF` (`GuildMemberLeftGameAsync`). So for an offline member whose name was cached the publish still happens with `serverId == 0xFF`: `GuildChangeToGameServerPublisher` silently drops it, and the Dapr `GuildChangePublisher` invokes `gameServer256` and logs an error on every call. Reachable when the target logs out between the action's online check and the guild server processing the change (and for any other future caller of this public API).

```suggestion
                if (listEntry.PlayerName is not null && listEntry.ServerId != OfflineServerId)
```

---

sven-n (commented):
Code review of the guild role assignment changes: 2 findings, both in `GuildServer.ChangeGuildMemberPositionAsync`. The alliance null-deconstruction fix and the guild list ordering look correct.

---

src/GuildServer/GuildServer.cs
```diff
@@ -206,17 +206,34 @@ public async ValueTask ChangeGuildMemberPositionAsync(uint guildId, Guid charact
...
+            }
+
+            guildMember.Status = role;
+            await guild.DatabaseContext.SaveChangesAsync().ConfigureAwait(false);
+                    if (guild.Members.TryGetValue(characterId, out var listEntry))
+                    {
+                        listEntry.PlayerPosition = role;
+
+                        // Offline members keep their cached name while their server id is
+                        // OfflineServerId; publishing to them is pointless (dropped or, over
```
sven-n: Reusing `AssignGuildToPlayerAsync` for a position change registers the player in the guild list a second time.

`GameServer.AssignGuildToPlayerAsync` unconditionally calls `Context.RegisterGuildMemberAsync`, and that does `guildList.Add(player)` on a plain `List<Player>` with no duplicate check (`GameServerContext.cs:215-226`). Until now that path only ran once per session (guild creation, join, player entered game). With this change, every role assignment re-adds the already registered target:

- guild chat (`ForEachGuildPlayerAsync`, `GameServer.cs:227`) is delivered twice (three times after two promotions, ...),
- `/guildmove`, `/guilddisconnect` and guild-war notifications act on the player multiple times,
- on logout `PlayerLeftWorldAsync` -> `UnregisterGuildMemberAsync` uses `List.Remove`, which removes only one occurrence, so a stale reference to the disconnected `Player` stays in `_playersByGuild` (leak + view calls on a dead connection).

Either make `RegisterGuildMemberAsync` idempotent, or publish the role change over a path that only updates `GuildStatus` without re-registering.

---

src/GuildServer/GuildServer.cs
```diff
@@ -206,17 +206,34 @@ public async ValueTask ChangeGuildMemberPositionAsync(uint guildId, Guid charact
...
+                return;
+            }
+
+            guildMember.Status = role;
+            await guild.DatabaseContext.SaveChangesAsync().ConfigureAwait(false);
+                    if (guild.Members.TryGetValue(characterId, out var listEntry))
+                    {
+                        listEntry.PlayerPosition = role;
+
+                        // Offline members keep their cached name while their server id is
```
sven-n: The publish isn't guarded against `OfflineServerId`. An offline member keeps its `PlayerName` while `SetServerId` sets `ServerId = 0xFF`, so a position change for an offline member (e.g. via the Dapr `GuildServerController`) is published to server id 255: the in-process publisher silently drops it, but `Dapr/GuildServer.Host/GuildChangePublisher.cs:63` invokes `gameServer256` and logs an error for every such change.

```suggestion
                if (listEntry.PlayerName is not null && listEntry.ServerId != OfflineServerId)
```

---

src/GameServer/MessageHandler/Guild/GuildRoleAssignHandlerPlugIn.cs
```diff
@@ -0,0 +1,50 @@
...
+    public bool IsEncryptionExpected => false;
+
+    /// <inheritdoc/>
+    public byte Key => GuildRoleAssignRequest.Code;
+
+    /// <inheritdoc/>
+    public async ValueTask HandlePacketAsync(Player player, Memory<byte> packet)
+    {
+        GuildRoleAssignRequest request = packet;
+        var position = request.Role.ConvertToPosition();
```
sven-n, line 45: Missing length validation: `request.Role` reads offset 4 and `request.PlayerName` does `ExtractString(5, 10, ...)`. A crafted short packet (e.g. `C1 05 E1 01`) makes the slice throw `ArgumentOutOfRangeException` (or `IndexOutOfRangeException` for an even shorter one) out of the handler. Other handlers guard this, e.g. `CharacterMoveBaseHandlerPlugIn`.

```suggestion
        if (packet.Length < GuildRoleAssignRequest.Length)
        {
            return;
        }

        GuildRoleAssignRequest request = packet;
```

---

src/GuildServer/GuildServer.cs
```diff
@@ -206,17 +206,34 @@ public async ValueTask ChangeGuildMemberPositionAsync(uint guildId, Guid charact
...
+
+            var guildMember = guild.Guild.Members.FirstOrDefault(m => m.Id == characterId);
+            if (guildMember is null)
+            {
+                this._logger.LogWarning("Guild {GuildId} member {CharacterId} not found, so its position can't be changed.", guildId, characterId);
+                return;
+            }
+
+            guildMember.Status = role;
+            await guild.DatabaseContext.SaveChangesAsync().ConfigureAwait(false);
```
sven-n: Minor: the `bool` result of `SaveChangesAsync` is discarded. Now that the change is also broadcast to the game server, a failed save means the client and the guild list show the new rank for the rest of the session while the database still holds the old one. Consider only updating the cached entry and publishing when the save succeeded.

---

sven-n (commented):
Code review: 4 findings, the first one is the important one (duplicate guild-member registration when the role change is published).

---

src/GameLogic/PlayerActions/Guild/GuildRoleAssignAction.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        var guildServer = (player.GameContext as IGameServerContext)?.GuildServer;
+        if (guildServer is null)
+        {
+            player.Logger.LogWarning("No guild server available");
+            return;
+        }
+
+        var target = player.GameContext.GetPlayerByCharacterName(targetName);
```
sven-n: The target is resolved only through the local game server's player list, so the feature silently does nothing for members who are online on another game server.

Scenario: the guild master plays on game server 1, the member on game server 2 (the guild list shows him with `ServerId` 2). The master assigns "Battle Master": `GetPlayerByCharacterName` returns `null` on server 1's context, the action logs "target is not online on this server" and returns. Since there is deliberately no response packet, the master gets no feedback either.

The guild server already holds the character id of every member (`GuildContainer.Members`), so `ChangeGuildMemberPositionAsync` could be reached for cross-server (and offline) members too — e.g. by resolving the name through the guild list instead of the local player list, and using the local player only for the same-guild/self checks.

---

src/GuildServer/GuildServer.cs
```diff
@@ -206,15 +206,37 @@ public async ValueTask ChangeGuildMemberPositionAsync(uint guildId, Guid charact
...
+            var guildMember = guild.Guild.Members.FirstOrDefault(m => m.Id == characterId);
+            if (guildMember is null)
             {
-                var guildMember = guild.Guild.Members.FirstOrDefault(m => m.Id == characterId);
-                if (guildMember != null)
+                this._logger.LogWarning("Guild {GuildId} member {CharacterId} not found, so its position can't be changed.", guildId, characterId);
+                return;
+            }
+
+            guildMember.Status = role;
```
sven-n: The new role is written into the tracked entity *before* the save is attempted, so the "only publish when the save succeeded" guarantee doesn't hold for the persisted state.

Scenario: the database is briefly unavailable. `SaveChangesAsync` throws (the EF implementation returns `true` or throws) or returns `false`. Either way the method returns without publishing, but `guild.Guild.Members[characterId].Status` already carries the new role in the container's shared `DatabaseContext`. The next successful save on that same container (e.g. `CreateGuildMemberAsync` or `KickMemberAsync` for the same guild) silently commits the role change that was just reported as failed — while the `GuildListEntry` and every online client keep showing the old position until that member relogs.

Writing `guildMember.Status` only after a successful save (or restoring the previous value in the failure/exception path) would make the persisted and published state consistent.

---

src/GameLogic/PlayerActions/Guild/GuildRoleAssignAction.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        // The fixed-size name field may be space-padded by the client.
+        var targetName = nickname.Trim();
+        if (string.IsNullOrEmpty(targetName))
+        {
+            player.Logger.LogWarning("Rejected guild role assignment of player {PlayerName}: empty target name.", player.Name);
+            return;
+        }
+
+        if (newPosition is not (GuildPosition.NormalMember or GuildPosition.BattleMaster or GuildPosition.AssistantMaster))
```
sven-n: Only the requested position is validated, not how many members may hold it. The original game allows one assistant master per guild and limits battle masters by guild level.

Scenario: the guild master sends C1 E1 with role 64 for ten different members; all ten become `AssistantMaster`. Besides the client's guild UI showing several assistants, each of them then passes `CastleSiegeNpcAuthorization` (which accepts `GuildMaster or AssistantMaster`), so castle siege NPC control is handed to an arbitrary number of members.

Worth at least a TODO if the limits are intentionally left out of this PR.

---

sven-n (commented):
Code review of the guild role assignment feature. The alliance null-deconstruction fix, the idempotent `RegisterGuildMemberAsync` and the guild list ordering look correct, and the packet handling (length check, role mapping, space-padded names) matches the packet definition. Three findings below: one functional gap for cross-server members, one persist/publish consistency issue in the guild server, and a missing role-count limit.

---

sven-n (changes requested):

PR #957, opened 2026-09-11
Fix monsters attacking through walls with ranged attacks

eduardosmaniotto:
### Problem
Monsters with ranged attacks (e.g. Hell Spider, Cursed Wizard)
could target and hit players standing behind walls. Target search
and the attack decision used pure Chebyshev distance with no terrain
check, and damage application only blocks on safezone.

### Fix
- `GameMapTerrain.HasLineOfSight(from, to)`: integer Bresenham over
  the raw `.att` attribute flags; only tiles with the `Blocked` (wall)
  bit stop projectiles — holes (`NoGround`) and water can be shot
  across, as in the original game (Dungeon pits, Chaos Castle ground).
  Endpoints are excluded so occupants never block themselves, and a
  diagonal step squeezing between two corner-touching wall tiles is
  treated as blocked. O(range) reads, zero allocations.
- `GameMapTerrain.AttributeMap` preserves the raw `TerrainAttributeType`
  bitmask per tile (the old code collapsed it to walkable/unwalkable and
  lost the wall-vs-hole distinction), plus `BlocksSight(x, y)` and
  `ApplyTerrainAttribute(x, y, attribute, set)` for runtime changes.
- `LocateableExtensions.HasLineOfSightTo`: same-map guard + delegation,
  mirroring the existing `IsInRange`/`IsAtSafezone` extension pattern.
  Call sites keep cheap `IsInRange()` as pre-filter, LOS second.
- AI gates (decision point, so monsters repath instead of camping):
  `BasicMonsterIntelligence` (search prefers nearest visible with
  nearest fallback, validity + attack gate require sight — also covers
  guards and summoned monsters' base behavior),
  `SummonedMonsterIntelligence` (same visible-preferred pattern),
  `RandomAttackInRangeTrapIntelligence`,
  `AttackAreaTargetInDirectionTrapIntelligence`,
  `AttackAreaWhenPressedTrapIntelligence`.
- Runtime terrain writers routed through the terrain so sight stays
  truthful: `MiniGameContext` (ground collapse), Kanturu barrier, and
  castle siege gates (close/open now saves and restores the wall bit
  instead of a stale walkability snapshot). Non-safezone changes never
  alter safezone status, exactly as before.
- Player skills intentionally untouched (client already blocks; server
  stays lenient).

### Known limitations
- Unknown high bits in newer `.att` files (e.g. `32`) default to
  non-sight-blocking — the permissive direction, closest to pre-fix
  behavior. If a map ever shows shooting through something solid, that
  tile's flag is the first lead.

### Tests
- `dotnet build src/GameLogic` — 0 errors, no new warnings.
- Full `MUnique.OpenMU.Tests`: 892/892 passed (incl. 27 siege,
  133 mini-game, 18 terrain).
- `GameMapTerrainTests` now covers: same-tile, clear lines,
  wall blocks symmetrically, endpoints never block, closed diagonal
  corner blocked, single-wall corner visible, diagonally-adjacent
  tiles visible, hole/water don't block (while asserting they're still
  unwalkable), `Blocked|NoGround` blocks, diagonal-between-holes
  passes, and runtime wall/hole/safezone transitions incl. safezone
  preservation.
- One `CastleSiegeNpcTests` setup updated to express its pre-blocked
  tile through the terrain API (natural wall in gate area) instead of
  a raw `WalkMap` poke unreachable in production.

#### Before patch
[Screencast_20260911_193219.webm](https://github.com/user-attachments/assets/62f42002-6f95-478e-939d-01400553dc91)

#### After patch

https://github.com/user-attachments/assets/8ba12bb5-fe3a-4883-92e6-6986ce61ace7


sven-n:
I think we should also prevent selecting targets which are not in line of sight. But there are some things to consider - in some terrains are not only walls, but also deep holes, e.g. in dungeon or chaos castle. Monsters with range attacks can attack over these holes.

sven-n (changes requested):
Check my comment

eduardosmaniotto:
Now it checks for holes/water. Tested in Chaos Castle:
https://github.com/user-attachments/assets/e24df5d6-d27a-4ce8-ad74-1304ce1118ed

src/GameLogic/MiniGames/MiniGameContext.cs
```diff
@@ -968,15 +968,13 @@ private void UpdateServerTerrain(ICollection<MiniGameTerrainChange> changes)
...
+            // The counters are ints on purpose: an end coordinate of 255 would make
+            // a byte counter wrap around and loop forever.
+            for (int x = change.StartX; x <= change.EndX; x++)
             {
-                for (var y = change.StartY; y <= change.EndY; y++)
+                for (int y = change.StartY; y <= change.EndY; y++)
                 {
-                    map[x, y] = targetValue;
-                    this.Map.Terrain.UpdateAiGridValue(x, y);
+                    this.Map.Terrain.ApplyTerrainAttribute((byte)x, (byte)y, change.TerrainAttribute, change.SetTerrainAttribute);
```
sven-n, line 977: This changes the semantics of a terrain change in a way that breaks Blood Castle.

Before, a non-safezone change was authoritative about walkability, independent of which attribute was configured:

```csharp
var targetValue = isSafezone ? change.SetTerrainAttribute : !change.SetTerrainAttribute;
map[x, y] = targetValue;   // WalkMap
```

So `SetTerrainAttribute = false` always made the area walkable. Now walkability is recomputed from the raw `.att` bits, so the configured `TerrainAttribute` has to match the bit that is actually set in the terrain file — otherwise the area stays blocked.

The Blood Castle bridge toggle is configured as `NoGround` (`BloodCastleInitializer.cs:280-281`, area X 13-15 / Y 70-75), but in the terrain data those tiles are not `NoGround` on most levels:

| map | terrain file | raw value at X 13-15 / Y 70-75 | after `Remove NoGround` |
|---|---|---|---|
| 11 (BC1) | `Terrain11.att` | `4` (Blocked) | `4` → still unwalkable |
| 12 (BC2) | `Terrain12.att` | `8` (NoGround) | `0` → walkable ✔ |
| 13-17 (BC3-7) | falls back to `Terrain11.att` | `4` | `4` → still unwalkable |
| 52 (BC8) | `Terrain52.att` | `12` (Blocked\|NoGround) | `4` → still unwalkable (and still sight-blocking) |

Result: on Blood Castle 1 and 3-8 the bridge never opens after the required kills, and the event can't be finished. (The other three BC areas — entrance, behind-gate, altar — are `4` and configured as `Blocked`, so those are fine, and Kanturu's barrier really is all `8`, so that one is correct.)

Options, roughly in order of preference:

1. Keep `ApplyTerrainAttribute` for the attribute bookkeeping, but when `SetTerrainAttribute == false`, clear *all* movement-blocking bits (`Blocked | NoGround | Water`) rather than only the configured one — that reproduces the old "this area is now open" semantics exactly while keeping the sight bit truthful.
2. Or fix the configuration (`bridgeToggleArea.TerrainAttribute = Blocked`) — but that only works if every terrain file agrees, and `Terrain52.att` has `12` there, so it would still need option 1's behaviour or a second change entry.

Worth a regression test that asserts a `Remove` change makes the area walkable regardless of the raw value underneath.

---

src/GameLogic/GameMapTerrain.cs
```diff
@@ -159,6 +176,125 @@ public Point GetRandomCoordinate(Point point, byte maximumRadius)
...
+            }
+
+            if (e2 <= dx)
+            {
+                err += dx;
+                y += sy;
+            }
+
+            // Reached the target tile: endpoints never block sight.
+            if (x == x1 && y == y1)
```
sven-n: The early return here is placed *before* the closed-corner check, which makes the result asymmetric when the corner sits next to the target instead of next to the viewer.

With walls at `(11,12)` and `(12,11)`:

* `HasLineOfSight((10,10), (12,12))` → `true` (last step is the diagonal into the target, so the corner check is skipped)
* `HasLineOfSight((12,12), (10,10))` → `false` (same corner, now checked on the first step)

So a player tucked into a corner pocket can still be shot, depending on which side the monster is standing. `HasLineOfSightWallBetweenIsFalse` asserts symmetry for the straight-wall case, but `HasLineOfSightDiagonalCornerBetweenTwoWallsIsFalse` only checks one direction.

Moving the corner check above the endpoint return fixes it:

```csharp
if (x != previousX && y != previousY
    && this.BlocksSight(previousX, y) && this.BlocksSight(x, previousY))
{
    return false;
}

if (x == x1 && y == y1)
{
    return true;
}
```

`HasLineOfSightDiagonallyAdjacentTilesAreTrue` would then need a look: with walls at `(10,11)` and `(11,10)`, `(10,10) → (11,11)` becomes blocked under that ordering. Which behaviour you want for the melee-adjacent case is a judgement call — if diagonally adjacent should always be visible, an explicit `Chebyshev distance <= 1 → true` guard at the top expresses that better than relying on the return order. Either way, it's worth asserting both directions in the corner tests.

---

src/GameLogic/NPC/RandomAttackInRangeTrapIntelligence.cs
```diff
@@ -44,9 +44,10 @@ protected override async ValueTask TickAsync()
             return;
         }
 
-        // Target in Attack Range?
+        // Target in Attack Range with a clear line of sight?
         ushort dist = (ushort)this._currentTarget.GetDistanceTo(this.Trap);
-        if (this.Trap.Definition.AttackRange + 1 >= dist)
+        if (this.Trap.Definition.AttackRange + 1 >= dist
+            && this.Trap.HasLineOfSightTo(this._currentTarget))
```
sven-n, line 54: The sight check gates the attack, but nothing invalidates `_currentTarget` when it loses sight. `TickAsync` only re-searches when the target becomes inactive, enters a safezone, or leaves the observers — so a target that simply steps behind a wall stays latched, and the trap stops attacking *anyone*, including players standing right on top of it.

Unlike a monster, a trap can't walk around the wall, so it never recovers on its own. That's exploitable: one player standing behind a wall in the trap's observer range disables it for everyone else.

Adding the sight check to the invalidation condition makes the trap fall back to `SearchNextTargetAsync` (which already prefers a visible target):

```csharp
if (!this._currentTarget.IsActive()
    || this._currentTarget.IsAtSafezone()
    || !this.Trap.HasLineOfSightTo(this._currentTarget)
    || !await this.IsTargetInObserversAsync().ConfigureAwait(false))
```

`BasicMonsterIntelligence` already does this correctly via `IsCurrentTargetValid`.

---

src/GameLogic/NPC/SummonedMonsterIntelligence.cs
```diff
@@ -48,13 +48,33 @@ public override void RegisterHit(IAttacker attacker)
...
+        NonPlayerCharacter? nearest = null;
+        double nearestDistance = double.MaxValue;
+        NonPlayerCharacter? nearestVisible = null;
+        double nearestVisibleDistance = double.MaxValue;
+        foreach (var candidate in currentMap?.GetAttackablesInRange(this.Owner.Position, 8)
+                     .Where(o => o is Monster { SummonedBy: null, IsAlive: true })
+                     .OfType<NonPlayerCharacter>()
+                 ?? [])
+        {
+            var distance = candidate.GetDistanceTo(this.Owner);
```
sven-n: Minor, but the two criteria use different reference points: the distance is measured from `this.Owner` (the player) while the sight check runs from `this.Npc` (the summon). The distance-from-owner part is pre-existing and intentional (the search radius is around the owner), so it's not wrong, just easy to misread — a short comment would help.

More generally, this is now the third copy of the same "track nearest and nearest-visible, prefer visible" loop (`BasicMonsterIntelligence.SearchNextTarget`, `RandomAttackInRangeTrapIntelligence.SearchNextTargetAsync`, here). A small helper, e.g.

```csharp
internal static T? GetNearestPreferVisible<T>(IEnumerable<T> candidates, ILocateable viewer, Func<T, double> distance)
    where T : class, ILocateable
```

would remove the triplication and keep the tie-breaking consistent if it ever changes.

---

sven-n (commented):
Thanks, this is a well-structured change — keeping the raw `.att` bitmask instead of collapsing it to walkable/unwalkable is the right foundation, the Bresenham implementation is allocation-free, and routing all runtime terrain writers through `ApplyTerrainAttribute` is a real improvement over the scattered `WalkMap` pokes (the `byte` loop counter fix at coordinate 255 is a nice catch along the way).

Four things from my side, one of them blocking:

**1. Blood Castle regression (blocking)** — recomputing walkability from the raw bits means a mini-game terrain change now only opens an area if the configured `TerrainAttribute` matches the bit actually present in the terrain file. The Blood Castle bridge is configured as `NoGround`, but the tiles are `Blocked` (`4`) on maps 11 and 13-17, and `Blocked|NoGround` (`12`) on map 52. Only Blood Castle 2 has plain `NoGround` there. So on 7 of 8 levels the bridge would never open. Details and suggested fixes in the inline comment on `MiniGameContext.cs`. Kanturu's barrier really is all `8`, so that part of the change is correct.

**2. `HasLineOfSight` is not symmetric for the diagonal corner case** — the endpoint short-circuit runs before the closed-corner check, so a corner adjacent to the *target* doesn't block while the same corner adjacent to the *viewer* does. Inline on `GameMapTerrain.cs`.

**3. Traps latch onto targets that lose sight** — `RandomAttackInRangeTrapIntelligence` gates the attack on sight but never invalidates `_currentTarget`, and a trap can't reposition. Inline on that file.

**4. Duplication / reference point in the target search** — inline on `SummonedMonsterIntelligence.cs`, non-blocking.

### One question on the sight mask

`Blocked|NoGround` (`12`) currently blocks sight, and the test `HasLineOfSightBlockedHoleCombinationBlocks` pins that down. That value is not rare in the shipped terrain data — it's the dominant out-of-bounds marker on the arena maps (`Terrain41-47`, `Terrain51`: ~58-63k tiles each), but it also appears inside playable maps: `Terrain2.att` (Dungeon) has 21592 tiles of pure `8` and 4660 of `12`, and `Terrain52.att` has it on the Blood Castle bridge. Since your original concern was exactly the dungeon/Chaos Castle pits, it'd be worth confirming in-game that the pits you meant are plain `8` and not `12` — otherwise the fix would still block those shots on some maps.

On the "unknown high bits" limitation in the description: the data supports the permissive choice. `Terrain19`/`Terrain54` use bit `32` (659 + 2707 + 32 tiles), and `Terrain11`/`25`/`31`/`32`/`35`-`37` carry bits `64`/`128` on scattered tiles. These look like the client's `TW_ACTION`/`TW_HEIGHT`/`TW_CAMERA_UP` flags rather than walls, so not blocking on them is right — and where they're combined with bit `4` the mask still blocks, which is what you want. Might be worth naming those bits in the comment on `SightBlockingAttributes`.

### Smaller notes

- `AttributeMap` is exposed as a public mutable `byte[,]`, so a direct write silently desyncs `WalkMap`/`AIgrid`/`SafezoneMap`. Same shape as the existing `WalkMap`, so not a new sin — but now that all in-tree writers go through `ApplyTerrainAttribute`, consider whether `UpdateAiGridValue` still needs to be public.
- `IsWalkableValue` could be written as `(value & ~(byte)TerrainAttributeType.Safezone) == 0`, which reads closer to the new bitmask model than `value == 0 || value == 1`.
- The two branches of `ApplyTerrainAttribute` duplicate the set/clear expression; hoisting it above the safezone check would shorten the method.
- `IsCurrentTargetValid` now drops the target the moment it leaves sight. That's fine for repathing, but it also means a player attacking a monster from behind a wall (player skills stay lenient, as you note) causes the monster to re-search every tick. Worth a quick look in-game that it doesn't produce visible target-flapping.

---

src/GameLogic/GameMapTerrain.cs
```diff
@@ -159,13 +189,151 @@ public Point GetRandomCoordinate(Point point, byte maximumRadius)
...
+        {
+            this.AttributeMap[x, y] |= (byte)attribute;
+        }
+        else
+        {
+            // Removing an attribute re-opens the area: the configured attribute
+            // doesn't necessarily match the bits in the terrain file (e.g. the Blood
+            // Castle bridge is toggled as NoGround while most levels carry Blocked),
+            // so every movement-blocking bit is cleared instead of only the configured one.
+            this.AttributeMap[x, y] = (byte)(this.AttributeMap[x, y] & ~MovementBlockingAttributes);
```
sven-n: **Correction to my previous review: the Blood Castle regression I reported does not exist. My analysis was wrong.**

I matched map number to `Terrain{N}.att`, but `TerrainUpdateHelper.GetTerrainFileName` starts its search at `gameMapDefinition.Number + 1` — map numbers are 0-based while the terrain files are 1-based (Lorencia, map `0`, loads `Terrain1.att`). So Blood Castle 1 (map `11`) loads **`Terrain12.att`**, not `Terrain11.att`, and Blood Castle 8 (map `52`) loads **`Terrain53.att`**. Blood Castle 2-7 have no file of their own and fall back to `Terrain12.att`.

Re-checked against the files that are actually loaded — both are identical in these areas:

| area | configured | raw value in `Terrain12.att` / `Terrain53.att` |
|---|---|---|
| bridge 13-15 / 70-75 | `NoGround` | `8` NoGround (+ one `0`) ✔ |
| gate 13-15 / 76-79 | `Blocked` | `4` ✔ |
| entrance 13-15 / 15-23 | `Blocked` | `4` ✔ |
| behind gate 11-25 / 78-89 | `Blocked` | `4` ✔ |
| altar 8-11 / 78-83 | `Blocked` | `4` ✔ |

Every configured attribute matches the bits actually in the terrain. Your original per-bit removal was correct, and the bridge would have opened fine. Sorry for the wrong lead — and for the wasted work.

That said, this version isn't wrong either: pre-PR `master` set `WalkMap[x, y] = true` unconditionally on removal, so clearing every movement-blocking bit reproduces the old behaviour exactly, while the per-bit version was slightly stricter than `master`. So both are defensible and I don't think anything here blocks the PR.

If you want my preference now that the motivation is gone: I'd go back to clearing only the configured bit. It matches what the method signature promises, and the broad clear has one side effect worth knowing about — `CastleSiegeNpcController.ReleaseGateTerrain` calls `ApplyTerrainAttribute(x, y, Blocked, state.WasBlocked)`, so for a tile that was unwalkable *without* the `Blocked` bit (`WasBlocked == false`), opening the gate would now clear its `NoGround`/`Water` and leave it walkable, where `master` restored the exact prior state. I checked the shipped data and it doesn't bite today — the six gate rectangles on Valley of Loren (map `30` → `Terrain31.att`) contain only `0`, `4` and `20` (`Blocked|Water`), and every unwalkable tile there carries the `Blocked` bit — so this is about the API shape, not a live bug.

Either way, the new `ApplyTerrainAttributeRemoveOpensAreaRegardlessOfRawValue` test is worth keeping; it pins the semantics down whichever you choose.

---

src/GameLogic/GameMapTerrain.cs
```diff
@@ -159,13 +189,151 @@ public Point GetRandomCoordinate(Point point, byte maximumRadius)
...
+        }
+
+        // Melee range is always visible: adjacent tiles share at most a corner,
+        // which must never read as a wall.
+        if (Math.Abs(x1 - x0) <= 1 && Math.Abs(y1 - y0) <= 1)
+        {
+            return true;
+        }
+
+        int dx = Math.Abs(x1 - x0);
```
sven-n, line 228: Moving the corner check above the endpoint return fixed the corner-pocket case, but sight is still not symmetric in general — the remaining cause is Bresenham's tie-breaking rather than the corner rule. When the line passes exactly through a lattice corner, the traversal picks a different cell depending on which end it starts from, so the two directions walk different tiles.

Minimal repro — a single wall, nothing diagonal about the geometry:

```
wall at (10, 11)
HasLineOfSight((10,10), (11,12)) → true   // steps through (11,11), clear
HasLineOfSight((11,12), (10,10)) → false  // steps through (10,11), the wall
```

I brute-forced it over random 12×12 wall layouts (~3.1M ordered pairs): **3.9%** of pairs disagree between the two directions.

Normalizing the endpoints to a canonical order before the walk makes it exactly symmetric — same brute force, 0 disagreements out of 3.1M:

```csharp
// Normalize the direction so the traversal doesn't depend on which side is
// looking: Bresenham picks a different cell at exact lattice corners
// depending on the starting endpoint.
if (x1 < x0 || (x1 == x0 && y1 < y0))
{
    (x0, x1) = (x1, x0);
    (y0, y1) = (y1, y0);
}
```

This is safe with respect to the endpoint rule: the start tile is never tested (the loop steps before checking) and the end tile returns early, so both endpoints stay excluded whichever way round they are.

On impact — every call site today passes the NPC as the viewer (`monster.HasLineOfSightTo(target)`), so no single check is evaluated both ways in production and nothing is visibly broken. What it does mean is that two monsters mirrored about a player can get different answers for the same geometry, and that the tests assert a property the code doesn't hold: `HasLineOfSightWallBetweenIsFalse` says *"Attacker and target swapped: sight is symmetric"*, and the corner test's new doc comment says *"Both directions are asserted because sight is symmetric"*. Both pass, but only because their geometries avoid the tie. Your call whether that's worth the four lines — I'd take it, since a future player-side check would otherwise disagree with the monster-side one.

---

sven-n (commented):
Re-reviewed at `5a9019a`. All four points from the last round are addressed, and CI is green.

**First, a correction I owe you: the Blood Castle regression I reported in my last review was wrong.** I looked up `Terrain{mapNumber}.att`, but `TerrainUpdateHelper` resolves map `N` to `Terrain{N+1}.att` — map numbers are 0-based, terrain files 1-based. Blood Castle 1 loads `Terrain12.att` (not `Terrain11.att`) and BC8 loads `Terrain53.att`; in both, the bridge tiles really are `NoGround`, exactly as configured. Your original per-bit removal would have worked. Details and what I'd suggest doing about the current version are in the inline comment on `ApplyTerrainAttribute` — short version: what you have now is also fine (it matches pre-PR `master` semantics), so nothing blocks the PR either way. Sorry for sending you down that path.

**One thing still open:** `HasLineOfSight` remains asymmetric, now because of Bresenham's tie-breaking rather than the corner rule. Minimal repro is a single wall at `(10,11)` with `(10,10) ↔ (11,12)` — true one way, false the other. Brute force over random layouts puts it at ~3.9% of pairs; a four-line endpoint normalization takes it to exactly 0. No production call site evaluates a pair both ways today, so nothing is visibly broken, but the tests do assert symmetry in their comments. Inline on the Bresenham setup.

**What's fixed well:**

- **Diagonal corner pocket** — checking the corner before the endpoint return is the right ordering, and the explicit melee-adjacency short-circuit at the top is a cleaner way to express "adjacent is always visible" than relying on return order. Good that the corner test now asserts both directions.
- **Trap latching** — adding the sight check to the invalidation condition is exactly right, and the comment explaining *why* a trap needs this when a monster doesn't (it can't walk around) is the kind of thing that stops someone "simplifying" it back out later.
- **`NpcTargetSelection.GetNearestPreferVisible`** — nice extraction. Passing `getDistance` as a delegate rather than assuming distance-to-viewer is what lets the summon keep its owner-centered radius while checking sight from itself, and the comment there now makes that split explicit instead of looking like a bug.
- **Small notes** — `UpdateAiGridValue` going private, the bitmask form of `IsWalkableValue`, and documenting the high `.att` bits all landed.

Nothing else from me. With the symmetry point resolved (or consciously deferred), this looks good to go from my side.

---

src/GameLogic/MiniGames/MiniGameContext.cs
sven-n: Retracted — this finding was wrong. I resolved map `N` to `Terrain{N}.att`, but `TerrainUpdateHelper` resolves it to `Terrain{N+1}.att` (map numbers are 0-based, terrain files 1-based). Blood Castle 1 loads `Terrain12.att` and BC8 loads `Terrain53.att`; in both, the bridge tiles are `NoGround` exactly as configured, so the per-bit removal always worked. Details in the thread on `ApplyTerrainAttribute`. Resolving this one so it doesn't stand as the last word.

---

sven-n (commented):
Re-reviewed at `d5fbefc`. Both open points are resolved, CI is green, and I have nothing further. I've resolved all six threads.

**Endpoint normalization** — verified. I re-ran the brute force against the exact code as it now stands (including the melee short-circuit and the corner-before-endpoint ordering): **0 asymmetric pairs out of 7.6M** ordered pairs over random wall layouts, against 3.9% before. All the existing sight cases still behave as their tests describe — clear lines, wall-between in both directions, endpoint exemptions, closed corner in both directions, single-wall corner visible, diagonal-adjacent visible. Placing the swap after the same-tile and melee checks is right: those are direction-independent anyway, so they don't need it, and the swap can't affect them. `HasLineOfSightTieBreakingIsSymmetric` pins the exact case down.

**Revert of the remove-path** — agreed, and I re-verified the whole caller surface on this branch rather than trusting the earlier analysis. There are exactly five `ApplyTerrainAttribute` call sites, and only three can pass `false`:

- **`MiniGameContext`** — the only configured non-safezone removals in the codebase are Blood Castle's five areas, and every one matches the bits in the file actually loaded (`Terrain12.att` / `Terrain53.att`): bridge `NoGround` on `8`, the other four `Blocked` on `4`. Chaos Castle only *sets* `NoGround` and toggles `Safezone`, so it takes the other branches.
- **`KanturuContext`** — removes `NoGround` from the barrier column, which is uniformly `8` in `Terrain39.att` (map `38`).
- **`CastleSiegeNpcController.ReleaseGateTerrain`** — this is the one the revert actually improves. With the per-bit clear the close/open cycle round-trips exactly: a gate tile at `20` (`Blocked|Water`) has `WasBlocked == true` and comes back as `20`, and a hypothetical `16` (water, no wall) closes to `20` and reopens to `16` instead of becoming walkable. Matches `master`'s "restore prior state" intent.

`ApplyTerrainAttributeRemoveKeepsOtherBits` is a good addition — it's the test that would have caught the round-trip problem, and it documents the contract the castle gates depend on.

The rest stands as reviewed: the corner-pocket ordering, the trap invalidation, and `NpcTargetSelection.GetNearestPreferVisible` all look right, and the comments explaining *why* each one is shaped that way should survive future refactors.

Good to merge from my side.

---

PR #964, opened 2026-09-15
Add scripted test actors, their control endpoint and a login probe

### LINKED ISSUE #963: Scripted test actors: connection-less players driven by commands, for testing game mechanics
**Is your feature request related to a problem? Please describe.**

Game mechanics which involve more than one player are hard to test without
humans at keyboards: PvP needs two players, a party needs several, and a
victim's health is only known to the victim itself. The unit tests in `tests/`
cover logic in isolation, and #905 is about the persistence layer, but there
is nothing that can *act* in a running world and report what happened to it.
The population bots (`Bots`) animate characters, but they follow their own AI
and do not report anything back.

**Describe the solution you'd like**

A `ScriptedPlayer` in `GameLogic` (next to `Bots/` and `Offline/`): a
connection-less `Player` which logs in like `OfflinePlayer` but is *not* one
(so bots defend against it, mini games do not skip it, the admin panel does
not list it as offline), executes commands - `walk`, `attack`, `skill`, `say`,
`pickup`, `warp`, `halt` - and records everything the client would have seen
as a stream of JSON events with attribution (hit *by whom*, recorded from
`IAttackableGotHitPlugIn`). Every command answers with a result or an error
code; nothing is a silent no-op.

A small control endpoint in the `Startup` host - newline-delimited JSON over
TCP, hosted as a `BackgroundService` - which is off by default and only
started when `OPENMU_ACTOR_PORT` is set. It also exposes `bots on/off/status`
through the same plug-in configuration the admin panel edits, so an actor has
opponents that fight back.

Beside it, a nested `Network/LoginProbe` console tool which logs an account in
over the real wire (connect server → game server, using
`MUnique.OpenMU.Network`) and holds the session, for the cases which need a
genuine connection (e.g. "an account which is connected cannot be animated").

I have this implemented and in daily use (~4.7k lines incl. 43 unit tests and a
docs-website page) and would like to send it as a PR. Before I do, two design
questions where I'd rather follow your preference:

1. **Endpoint configuration.** Today: environment variable, unauthenticated,
   meant for loopback. Would you prefer a plug-in configuration editable in
   the admin panel (with the port and an on/off switch), and/or a shared
   secret on the endpoint?
2. **Distributed deployment.** The endpoint is registered in
   `Startup/Program.cs` only; `Dapr/GameServer.Host` does not open it. Is
   "single-process host only" acceptable, or should it be wired there too?

**Describe alternatives you've considered**

- An ASP.NET controller on the admin panel port: rejected because the panel's
  controllers `RequireAuthorization()` and following an event stream would
  need SSE/WebSockets, which is awkward from a shell script.
- A network bot speaking the client protocol: it would test the wire too, but
  it cannot see the victim's health, and it is a much bigger thing.
- Extending `Bots` with a command mode: the bots are deliberately offline
  players and the rest of the server treats them as such.

**Additional context**

Related: #905 (integration tests), #734 (castle siege testing strategy - the
actors would give those tests players to drive).

alandarev:
Closes #963.

## What

Connection-less players which execute commands and report what the game shows
them, for testing game mechanics without a client. Docs page:
`docs-website/docs/server-features/test-actors.md`.

- **`src/GameLogic/TestActors/`** - `ScriptedPlayer` (a `Player`, not an
  `OfflinePlayer`: bots defend against it, mini games count it, the admin panel
  shows it as a player), the command loop (runs on the player's own tick,
  inside the persistence lock; long commands are interruptible), the event log
  (strictly increasing sequence numbers, UTC timestamps), a recording view
  plug-in container, a hit recorder on `IAttackableGotHitPlugIn` for
  attacker/victim attribution, the registry, the protocol handler and JSON.
- **`src/Startup/TestActors/`** - `ActorControlService`, a `BackgroundService`
  speaking newline-delimited JSON over TCP. Registered in `Program.cs` only
  when `OPENMU_ACTOR_PORT` names a port; otherwise nothing is listening. It
  binds `127.0.0.1` unless `OPENMU_ACTOR_ADDRESS` names another address. It
  can also switch the population bots on and off through their persisted
  plug-in configuration, so an actor has opponents.
- **`src/Network/LoginProbe/`** - nested console tool (like `Network/Analyzer`
  and `ChatServer/ExDbConnector`) which logs an account in over the real wire
  and holds the session; used to test that a connected account cannot be
  animated by an actor. The password comes from `LOGINPROBE_PASSWORD`
  (default: the account name).
- **`tests/MUnique.OpenMU.Tests/TestActors/`** - 53 unit tests.

Commands: `spawn`, `stop`, `list`, `state`, `nearby`, `walk`, `attack`,
`skill`, `say`, `pickup`, `warp`, `halt`, `events [--since N] [--follow]`,
`bots on|off|status`. Every command answers with a result or an error code
(`out_of_range`, `safezone`, `not_in_view`, `unknown_skill`, `no_path`,
`interrupted`, ...).

## Try it

```bash
OPENMU_ACTOR_PORT=55990 dotnet run --project src/Startup
printf '%s\n' '{"id":"1","cmd":"spawn","actor":"test1"}' '{"id":"2","cmd":"state","actor":"test1"}' | nc 127.0.0.1 55990
```

## Design notes / open points

- The endpoint has **no authentication**; off by default, loopback by
  default. Per the review: environment variable rather than a plug-in
  configuration, no shared secret while it is loopback-only.
- `ActorHitRecorderPlugIn` is deliberately not a discoverable `[PlugIn]`; the
  endpoint registers it at the `IAttackableGotHitPlugIn` point on start, so it
  neither runs nor shows up on servers which never enable the actors.
- Registered in the all-in-one `Startup` host only; the Dapr
  `GameServer.Host` does not open it (stated in the docs page).
- `skill` casts on a target only; area/position casting
  (`AreaSkillAttackAction`) is left out on purpose.
- Touches outside the new folders: `Startup/Program.cs` (registration),
  `Network/MUnique.OpenMU.Network.csproj` (excludes the nested tool, like
  `Packets\**`), `MUnique.OpenMU.sln`, `docs-website/sidebars.js`.

## Testing

A recording of the feature being exercised live on a running server, driven
by a coding agent on my behalf: https://www.youtube.com/watch?v=lQuFN4KtkuI

- `dotnet test tests/MUnique.OpenMU.Tests --filter FullyQualifiedName~TestActors`: 53 passed.
- `MUnique.OpenMU.GameLogic` and `Network.LoginProbe` build with zero StyleCop
  warnings in the added files.
- `Startup/Dockerfile` image built from this branch; started with and without
  `OPENMU_ACTOR_PORT` (listener present / absent), with the port alone
  (bound to `127.0.0.1` inside the container) and with
  `OPENMU_ACTOR_ADDRESS=0.0.0.0`, and a spawn → walk → attack → events run
  against a population bot on a fresh PostgreSQL database.
- Review round 1 (166ca3ede): see the comment below for what changed and how
  it was verified. Codacy: the nine tool artefacts listed in the comments
  plus one more S1128 of the same kind (`using System.Net;` in
  `ActorEndpointOptions.cs`, where `IPEndPoint` and `IPAddress` are used).
- Used daily for a few weeks on a private server built on OpenMU.

Written with AI assistance and tested by hand, per CONTRIBUTING.


alandarev:
shortly, without AI slop:

this is a great tool for (semi-)automated testing of features, where one can instruct an AI agent via local port to control the bot and issue it commands for testing.
the list if supported operations may get extended as requested/required

alandarev:
response to Codacy's 9 issues (the rest fixed):

 What cannot be fixed (9)

 Codacy runs SonarC# without a compilation, so semantic rules misfire:

 - 3 × S1128 on System.Buffers / System.Net — ReadOnlySequence and IPAddress are used; removing them breaks the build.
 - 3 × S1172 unused parameters on the test fakes' IActorFactory / ILoginServer members — interface implementations cannot drop parameters.
 - 1 × S3453 on the test Fixture — it is created by its own static factory.
 - 1 × MD025 — every page under docs-website/docs has a front-matter title and an H1; changing only ours breaks the convention.
 - 1 × ESLint quotes on sidebars.js — the whole Docusaurus file is single-quoted.

 Verification (actual outcomes)

 - dotnet test … --filter TestActors → 43 passed, 0 failed in both repos.
 - cd src && docker build -f Startup/Dockerfile → exit 0; no StyleCop warning in any TestActors/LoginProbe file (the two Startup/Program.cs warnings are pre-existing
   upstream lines, confirmed by git blame).
 - docs-website: npm run build → SUCCESS.
 - Image run with OPENMU_ACTOR_PORT=55991: spawn / state / stop all answered {"ok":true,…}; run without the variable → 0 actor log lines (endpoint still off by
   default).
 - Codacy after the push: 9 findings, exactly the 9 listed above.

src/Startup/TestActors/ActorControlService.cs
```diff
@@ -0,0 +1,201 @@
...
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        // `docker stop` sends SIGTERM, but this host's main loop only stops the services from an
+        // AppDomain.ProcessExit handler, which the runtime cuts short - so the hosted services'
+        // StopAsync usually never runs in a container, and an actor's unsaved progress (anything
+        // since the last periodic save) would be lost on every `dev down`. Logging the actors out
+        // straight from the signal handler makes the shutdown promise hold regardless.
+        this._sigTermRegistration = PosixSignalRegistration.Create(PosixSignal.SIGTERM, _ => this.StopActorsOnSignal());
+
+        var listener = new TcpListener(IPAddress.Any, this._port);
```
sven-n: This is the one thing I'd like changed before anything else: the listener binds `IPAddress.Any`, while the class remark above says "the local stack publishes that port on the host's loopback address only" and the docs page says "Bind it to a loopback address, never expose it". Inside Docker that holds because of the port publishing, but the host is also run directly (`dotnet run --project src/Startup`, which is exactly what the PR description suggests) — there the variable alone opens an unauthenticated endpoint on every interface which can animate any account, walk it, chat as it, and switch the population bots on and off.

Please bind `IPAddress.Loopback` by default and only widen it when the operator explicitly asks, e.g. by letting the variable take `host:port` and defaulting the host to loopback. Then the code matches the documentation instead of relying on the deployment to contain it, and the "off by default" promise also covers "not reachable from outside by accident".

---

src/GameLogic/TestActors/ActorHitRecorderPlugIn.cs
```diff
@@ -0,0 +1,56 @@
...
+/// Records every hit an actor deals or receives, with the attribution the client views do not carry.
+/// </summary>
+/// <remarks>
+/// <see cref="MUnique.OpenMU.GameLogic.Views.World.IShowHitPlugIn"/> only tells the victim how much
+/// damage it took - not who dealt it - so this plugin point is the only place where attacker and
+/// victim are both known. It is therefore the single source of <c>hit</c> events: neither the
+/// recording view container nor a command result appends one, so a hit is recorded exactly once per
+/// involved actor (once on each side when two actors fight each other).
+/// </remarks>
+[PlugIn]
```
sven-n: A `[PlugIn]` which does not implement `IDisabledByDefault` is created as active by the data initialization (`DataInitializationBase`: `plugInConfiguration.IsActive = !plugInType.IsAssignableTo(typeof(IDisabledByDefault))`). So this test-only plugin runs on every single hit of every server, including production installations which never set `OPENMU_ACTOR_PORT`, and it shows up in the admin panel's plugin list for everyone.

The two type checks per hit are cheap, so this is about intent rather than throughput: please add `IDisabledByDefault` and mention in the docs page that the plugin has to be activated together with the endpoint, or drop the plugin attribute and register the plugin point only when the endpoint is enabled.

---

src/GameLogic/TestActors/ActorRegistry.cs
```diff
@@ -0,0 +1,219 @@
...
+        finally
+        {
+            this._lock.Release();
+        }
+    }
+
+    /// <inheritdoc />
+    public ScriptedPlayer? Find(string loginName)
+    {
+        this._lock.Wait();
```
sven-n: `List()` and `Find()` block synchronously on `_lock`, and that same semaphore is held across awaits in `SpawnAsync` — a database load of the account, the login server call and the whole world entry. Every `ping`, `list` and every actor command goes through `Find`, so while one spawn is running, each of those blocks a thread pool thread for its whole duration instead of yielding. With a couple of scripts against one endpoint that is easy to notice, and it is the kind of sync-over-async the rest of the code base avoids.

Since both callers (`ActorProtocolHandler`) are already async, I'd make them `ListAsync`/`FindAsync` and `await _lock.WaitAsync()`.

---

src/GameLogic/TestActors/ScriptedIntelligence.cs
```diff
@@ -0,0 +1,691 @@
...
+            walked += chunk.Count;
+        }
+
+        return ActorCommandResult.Success(
+            new ActorEventField("steps", path.Count),
+            new ActorEventField("walked", walked),
+            new ActorEventField("x", this._player.Position.X),
+            new ActorEventField("y", this._player.Position.Y));
+    }
+
```
sven-n, line 346: `walk` can report `ok` although the actor never moved, which is exactly the "silent no-op" the rest of the design rules out.

`PlayerMovement.WalkToAsync` refuses a walk whose first step is blocked (it resynchronizes the client and returns), and truncates the path at the first blocked later step. In the refusal case `IsWalking` is false immediately, the `while` loop falls through, `walked += chunk.Count` counts the chunk as done, and the command answers success with `steps == walked == path.Count`. In the truncation case the next chunk is built from the actual position but continues with the original path nodes, so its steps are no longer adjacent tiles.

The path finder uses `Terrain.AIgrid` while the movement check uses the walk map, so the two do disagree in practice (safe zones, temporarily blocked tiles). Suggestion: after each chunk compare `this._player.Position` with the chunk's last node, and on a mismatch stop and answer `no_path` (or a partial result) with the position actually reached; the final success can then also assert `Position == target`.

---

src/GameLogic/TestActors/BotsController.cs
```diff
@@ -0,0 +1,183 @@
...
+        configuration.Enabled = enabled;
+        if (enabled)
+        {
+            configuration.NumberOfAccounts = Math.Max(0, count ?? DefaultBotCount);
+
+            // One character per account, so the requested number is what stands in the world, and no
+            // bot logs out again while a scenario runs.
+            configuration.MaxCharactersPerAccount = 1;
+            configuration.PresenceRotation = false;
+        }
```
sven-n: `bots on` overwrites `NumberOfAccounts`, `MaxCharactersPerAccount` and `PresenceRotation` in the persisted plugin configuration and never restores them, so a single test command permanently replaces whatever the operator had configured in the admin panel, surviving the restart. On a dedicated test server that is harmless, but it is a surprising amount of state for a tool which is otherwise careful not to disturb the server.

Either restrict the write to `Enabled` (+ `NumberOfAccounts` when a count is given) and leave the character/rotation settings alone, or keep the overwrite but return the previous values in the result and document it on the docs page so a scenario can put them back.

---

src/GameLogic/TestActors/ActorProtocolHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+        ActorCommand? actorCommand = command switch
+        {
+            "walk" => new WalkCommand((byte)(GetNumber(request, "x") ?? 0), (byte)(GetNumber(request, "y") ?? 0)),
+            "attack" => new AttackCommand(
+                GetTarget(request) ?? string.Empty,
+                (int)(GetNumber(request, "times") ?? 1),
+                (int)(GetNumber(request, "interval") ?? 0)),
+            "skill" => new SkillCommand((ushort)(GetNumber(request, "skill") ?? 0), GetTarget(request) ?? string.Empty),
+            "say" => new SayCommand(GetString(request, "text") ?? string.Empty),
+            "pickup" => new PickupCommand((ushort)(GetNumber(request, "id") ?? 0)),
```
sven-n: Nit, same spirit as "nothing is a silent no-op": these casts are unchecked, so `{"cmd":"walk","x":300,"y":132}` walks to x=44 and a negative or oversized `skill`/`id`/`slot` wraps just as quietly. A range check answering `bad_request` would be a few lines in `GetNumber` (e.g. an overload taking the allowed range) and would keep a typo in a scenario from producing a plausible-looking wrong result.

---

sven-n (commented):
Thanks for writing this up as an issue first and for the very thorough PR description — that made the review a lot easier. I read through the whole diff; overall the design is sound and it clearly comes from actually using the thing: deriving from `Player` instead of `OfflinePlayer` with the reasoning spelled out, running every command through the player's persistence lock, the interruptible long commands, the double-checked account occupancy in `ActorRegistry.SpawnAsync` (login server *and* a scan of every context of the process, because bots and offline sessions bypass the login server), and taking hit attribution from `IAttackableGotHitPlugIn` because `IShowHitPlugIn` does not carry the attacker. The remarks sections explain the *why* in the places where it is not obvious, which is what I want in this code base.

I left six inline comments. The first one is the only one I consider blocking; the rest are things I'd like to see addressed or at least answered.

1. **The endpoint binds `IPAddress.Any`** while both the class remark and the docs page promise loopback. Please bind loopback by default.
2. `ActorHitRecorderPlugIn` is active by default on every installation, including ones which never enable the feature.
3. `ActorRegistry.List()`/`Find()` block synchronously on a semaphore which is held across the whole spawn.
4. `walk` can answer `ok` when the engine refused or truncated the walk.
5. `bots on` permanently rewrites persisted bot settings which it does not own.
6. Unchecked numeric casts in the protocol handler.

On the two design questions from #963:

1. **Endpoint configuration.** I'd keep the environment variable — a plugin configuration would be a nicer fit for the admin panel, but it would also mean the endpoint can be switched on in a UI on a live server, which is the opposite of what I want for an unauthenticated port. Environment variable plus a loopback default is the safer combination. A shared secret is not needed as long as it is loopback-only; if you want one anyway, a simple `{"token":"…"}` field checked before dispatch would do, and I would not build anything more than that.
2. **Distributed deployment.** "Single-process host only" is fine, and the docs page states it. The Dapr host would need the registry to reach across processes, which is a different feature; let's not carry that complexity for a test tool.

Two smaller things for the PR body rather than the diff:

- **`PosixSignalRegistration` for SIGTERM in `ActorControlService`.** The comment is right that this host's shutdown goes through `AppDomain.ProcessExit` and that `StopAsync` often does not get to run in a container — but that is a shortcoming of the host, not of the actors, and a test feature installing a process-wide signal handler to work around it is the wrong layer. It also blocks the signal thread for up to eight seconds. Could you split that out? If the host's shutdown is fixed properly (so hosted services are stopped on SIGTERM), every player benefits, and the handler here can go away. A separate PR or issue for that is welcome.
- **`--password` on the command line** of the login probe is visible in `ps` and lands in shell history; an environment variable or stdin would be the usual way. Minor, given it targets local test accounts, but easy to do.

Placement in `src/GameLogic/TestActors/` is acceptable to me — it needs too much internal access to live elsewhere, and `Bots/` and `Offline/` already set the precedent. Test count and the docs page are appreciated.

One process note, since the PR says the code was written with AI assistance: I read it as a whole and the reasoning holds up, but please make sure every non-obvious remark still matches the code after the next round — the `IPAddress.Any` case above is exactly the kind of drift between a confident comment and the implementation that I want to avoid in this repository.

---

alandarev:
Thanks for the careful read - all eight points are addressed in
166ca3ede, as one commit on top so the delta is easy to review.

1. **Loopback by default.** `OPENMU_ACTOR_PORT` binds `127.0.0.1` now. I went
   with a second variable instead of `host:port` in the same one, because a
   variable named PORT should not carry an address: `OPENMU_ACTOR_ADDRESS`
   (optional) names the bind address, e.g. `0.0.0.0` inside a container whose
   port publishing does the containment. An invalid port or address is logged
   and ignored, so the endpoint stays off. The class remark and the docs page
   now say what the code does. Verified on the image: the port alone shows
   `0100007F:<port>` in `/proc/net/tcp` and the published port resets; with
   `OPENMU_ACTOR_ADDRESS=0.0.0.0` it answers.
2. **Hit recorder.** Took your second option: `ActorHitRecorderPlugIn` is no
   longer a `[PlugIn]` - not discovered, no configuration row, not in the
   admin panel - and `ActorControlService.StartAsync` registers it at the
   `IAttackableGotHitPlugIn` point through the host's `PlugInManager`. The
   remark on the class explains why it is not discoverable.
3. **Registry lookups.** `List`/`Find` are `ListAsync`/`FindAsync` and await
   the semaphore; the remark on `ActorRegistry` now states why the lookups are
   asynchronous.
4. **Walk.** After each chunk the actor's position is compared with the
   chunk's last node; a refused or truncated walk answers `no_path` with
   `walked`, `x`, `y` of the position actually reached, and the final success
   asserts `Position == target`. Two tests cover the refused (first step
   blocked on the walk map only) and the truncated case.
5. **Bots.** `bots on|off` writes `Enabled`, plus `NumberOfAccounts` when
   `--count` is given, and nothing else; `MaxCharactersPerAccount` and
   `PresenceRotation` stay as configured and `bots status` reports all three
   (`presence_rotation` is new in the result). Documented on the page.
6. **Numbers.** Every numeric field is range-checked (`byte`, `ushort`, `int`,
   `times >= 1`); a value outside answers `bad_request` naming the field and
   the range. Eight test cases, one per field.

On the two PR-body notes:

- **SIGTERM handler** - removed. The actors are now stopped from `StopAsync`
  only, like every other hosted service. I'll open a separate issue for the
  host's shutdown (stopping the hosted services on SIGTERM instead of relying
  on `ProcessExit`), since that is where the fix belongs.
- **`--password`** - gone; the probe reads `LOGINPROBE_PASSWORD` and falls
  back to the account name. Readme and usage text updated.

And on the process note: I re-read every remark against the code after this
round, including the ones I did not touch.

Testing: 53 unit tests pass (`--filter FullyQualifiedName~TestActors`);
`GameLogic`, `Network.LoginProbe` and the `Startup` image build with no
StyleCop warning in the changed files; docs website builds; the image was run
with the port alone, with `OPENMU_ACTOR_ADDRESS=0.0.0.0`, without the variables
and with an invalid address, and a spawn → walk → attack → events run against a Noria goblin
shows `hit` events on both directions with the runtime-registered recorder.

Codacy will report one more S1128 (`using System.Net;` in
`ActorEndpointOptions.cs`); `IPEndPoint` and `IPAddress` are used there, it is
the same no-compilation artefact as the three already listed.

sven-n (commented):
Reviewed 166ca3ede against the previous round. All six inline points and both body points are addressed, and I verified the two which could have gone wrong in the fix rather than in the original:

- **Bind address.** `ActorEndpointOptions.TryParse` defaults to `IPAddress.Loopback`, an unparsable address returns `null` so the endpoint stays off rather than falling back to something wider, and `Program.cs` warns when the port is set but the pair is invalid — good, silent "it is set but nothing is listening" would have been the annoying outcome. The class remark and the docs page now describe the real behaviour.
- **Hit recorder registration.** Dropping `[PlugIn]` and calling `RegisterPlugIn<IAttackableGotHitPlugIn, ActorHitRecorderPlugIn>()` from `StartAsync` does reach the right instance: the `PlugInManager` singleton from line 315 of `Program.cs` is the one `GameServerContainer` hands to every `GameServer`, `IAttackableGotHitPlugIn` carries `[PlugInPoint]`, and the `[Guid]` the class kept is what `RegisterPlugInAtPlugInPoint` requires. Because the point already has discovered implementations (`SelfDefensePlugIn` and friends), the late registration only appends to the existing proxy — same thing the admin panel does when a plugin is activated at runtime — so the ordering behind the game servers is fine.

The rest reads as asked: `ListAsync`/`FindAsync` awaiting the semaphore, the per-chunk position check with the "the position is the only reliable report" reasoning, `bots` writing only what was requested and reporting all three settings, the range-checked numeric fields answering `bad_request`, no more signal handler, and the probe's password out of the process list. The added tests cover the refused and truncated walk, the out-of-range fields and the untouched bot settings, which is exactly where I would have wanted them.

One nit, nothing blocking: `InterruptWalkAsync` still reports `walked` as whole chunks only, so a `halt` in the middle of a chunk under-reports by up to 15 steps while `x`/`y` are exact. Now that `StepsWalked` exists, passing the in-flight chunk to it would make the interrupted result as precise as the failed one.

CI is green; the Codacy findings are the tool artefacts you listed.

---

PR #966, opened 2026-09-16
Item Durability Logic Refactor

ze-dom:
**To-do**

- [ ] Comment changes with source references
- [x] UpdatePlugin(s)

**Developments**

- Added durability thresholds, past which some of the item power-ups decrease by a certain factor
- Added [DurabilityDecayPlugIn.cs](https://github.com/MUnique/OpenMU/compare/master...ze-dom:item_durability_refactor?expand=1#diff-d873b55103e78d326f7de80cd756d4d0748bd05ea441db71ca6b38c6f3090186), for jewelry and wings durability decrease
- Reworked durability reduction calculations to reflect original values while keeping it configurable 

**Bugfixes**
- Possible durability reduction target items after dealing and receiving damage
- Master tree logic and values for skills _DurabilityReduction1_, _DurabilityReduction2_, _DurabilityReduction3_, _PetDurabilityStr_, _DurabilityReduction1FistMaster_, _DurabilityReduction2FistMaster_, _DurabilityReduction3FistMaster_
- Repairing items now raises the event `EquippedItemsChanged`. Previously repaired items w/ durability = 0 were not reinstating their power-ups.


src/GameLogic/Player.cs
```diff
@@ -731,16 +733,19 @@ public async ValueTask KillInstantlyAsync()
         }
 
         await this.HitAsync(hitInfo, attacker, skill?.Skill, isFinalStreakHit).ConfigureAwait(false);
-        await this.DecreaseItemDurabilityAfterHitAsync(hitInfo, skill).ConfigureAwait(false);
+
+        bool isPvpDamage = attacker is Player or IPlayerSurrogate;
+        float damage = isPvpDamage ? hitInfo.TotalDamage : attacker.Attributes[Stats.MinimumPhysBaseDmg];
+        await this.DecreaseItemDurabilityAfterHitAsync(damage, isPvpDamage, skill).ConfigureAwait(false);
```
ze-dom, line 757: zTeamS6.3: [ObjAttack.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjAttack.cpp#L2335), [user.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L9518)
emu: [ObjAttack.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjAttack.cpp#L2901), [user.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L8559)

src/GameLogic/NPC/AttackableNpcBase.cs
```diff
@@ -129,7 +129,7 @@ public int Health
 
             if (attacker is Player player)
             {
-                await player.AfterHitTargetAsync().ConfigureAwait(false);
+                await player.AfterHitTargetAsync(this.Attributes[Stats.DefensePvm], skill?.Skill?.DamageType).ConfigureAwait(false);
```
ze-dom, line 132: zTeamS6.3: [ObjAttack.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjAttack.cpp#L484), [user.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L9318)
emu: [ObjAttack.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjAttack.cpp#L522), [user.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L8363)

src/GameLogic/NPC/AttackableNpcBase.cs
```diff
@@ -139,7 +139,7 @@ public int Health
 
             if (attacker as IPlayerSurrogate is { } playerSurrogate)
             {
-                await playerSurrogate.Owner.AfterHitTargetAsync().ConfigureAwait(false);
+                await playerSurrogate.Owner.DecreaseRavenDurabilityAfterHitAsync(hitInfo.TotalDamage).ConfigureAwait(false);
```
ze-dom, line 142: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/DarkSpirit.cpp#L908), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/DarkSpirit.cpp#L911)

Unfortunately Dark Raven durability decrease is originally calculated in the middle of damage calcs, so we diverge here a bit.

src/GameLogic/PlayerActions/Items/SimpleItemCraftingHandler.cs
```diff
@@ -268,13 +269,14 @@ private async ValueTask<List<Item>> CreateResultItemsAsync(Player player, IList<
...
                 : (byte)Rand.NextInt(craftingResultItem.RandomMinimumLevel, craftingResultItem.RandomMaximumLevel + 1);
-            resultItem.Durability = craftingResultItem.Durability ?? resultItem.GetMaximumDurabilityOfOnePiece();
 
             this.AddRandomLuckOption(resultItem, player, successRate);
             this.AddRandomItemOption(resultItem, player, successRate);
             this.AddRandomSkill(resultItem, successRate);
             this.AddRandomExcellentOptions(resultItem, player);
 
+            resultItem.Durability = craftingResultItem.Durability ?? resultItem.GetMaximumDurabilityOfOnePiece();
+
```
ze-dom, line 279: Durability depends on item being excellent

src/GameLogic/PlayerActions/Skills/PlasmaStormSkillPlugIn.cs
ze-dom, line 1: Moved to [Player.cs](https://github.com/MUnique/OpenMU/pull/966/changes#diff-7095817e5ea85690f8e681594b50539993eaf0a73fa1d6aab6d7f8dd421eb752)

src/GameLogic/PlugIns/PeriodicTasks/DurabilityDecayPlugIn.cs
ze-dom, line 1: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8240), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7883)

src/GameLogic/PlugIns/PeriodicTasks/DurabilityDecayPlugIn.cs
```diff
@@ -0,0 +1,80 @@
...
+                        {
+                            // Ring of warrior, etc.
+                            continue;
+                        }
+
+                        if (attributes[Stats.IsInSafezone] < 1 || isTransformationRing || item.IsWing())
+                        {
+                            double decrementWeight = new ItemIdentifier(item.Definition!.Number, item.Definition.Group) switch
+                            {
+                                var _ when isTransformationRing => 11.28, // 11.28 / 564 = 0.02
```
ze-dom: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8784), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L8140)

src/GameLogic/PlugIns/PeriodicTasks/DurabilityDecayPlugIn.cs
```diff
@@ -0,0 +1,80 @@
...
+                                _ => 1,
+                            };
+
+                            var durationIncrease = attributes[Stats.JewelryAndWingsDurationIncrease];
+                            if (durationIncrease == 0 || isTransformationRing)
+                            {
+                                durationIncrease = 1;
+                            }
+
+                            var decrement = decrementWeight / (player.GameContext.Configuration.HitsPerOneItemDurability * durationIncrease);
```
ze-dom, line 65: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5752), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3735)
564 hard limit check

src/GameLogic/Attributes/Stats.cs
```diff
@@ -1332,14 +1332,31 @@ public class Stats
     public static AttributeDefinition ShieldRateIncrease { get; } = new(new Guid("C2FFBA46-4DC3-4861-84F5-313DED038997"), "SD Rate Increase", "Increases the shield’s damage absorb rate in PVP according to set value.");
 
     /// <summary>
-    /// Gets the item duration increase attribute definition.
+    /// Gets the weapon and armor duration increase MST attribute definition.
     /// </summary>
-    public static AttributeDefinition ItemDurationIncrease { get; } = new(new Guid("03EBE702-90FF-473C-8CBB-E83669FE3C68"), "Item Duration Increase", string.Empty);
+    /// <remarks>Includes weapons, shields and armor items. Excludes jewelry, wings and pets.</remarks>
+    public static AttributeDefinition WeaponAndArmorDurationIncrease { get; } = new(new Guid("03EBE702-90FF-473C-8CBB-E83669FE3C68"), "Weapon And Armor Duration Increase (MST)", string.Empty);
```
ze-dom, line 1338: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L326), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L314)

Image

The tooltip is inaccurate: it's actually 67% increase: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5852), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3831).

src/GameLogic/Attributes/Stats.cs
```diff
@@ -1332,14 +1332,31 @@ public class Stats
...
 
     /// <summary>
-    /// Gets the pet duration increase attribute definition.
+    /// Gets the jewelry and wings duration increase MST attribute definition.
     /// </summary>
-    public static AttributeDefinition PetDurationIncrease { get; } = new(new Guid("B4455150-D3A9-4A5F-914B-F41F9387FE9A"), "Pet Duration Increase", string.Empty);
+    public static AttributeDefinition JewelryAndWingsDurationIncrease { get; } = new(new Guid("E539FDD2-72C5-484A-B0F4-40565ABAEDD9"), "Jewelry And Wings Duration Increase (MST)", string.Empty);
```
ze-dom, line 1343: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L331), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L319)

Image

The tooltip is inaccurate: it's actually 67% increase: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5750), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3733).

src/GameLogic/Attributes/Stats.cs
```diff
@@ -1332,14 +1332,31 @@ public class Stats
...
+
+    /// <summary>
+    /// Gets the pet duration increase MST attribute definition.
+    /// </summary>
+    /// <remarks>Applies to imp, guardian angel, uniria, dinorant and fenrir. Cash shop items excluded.</remarks>
+    public static AttributeDefinition PetDurationIncrease { get; } = new(new Guid("E9CCA156-E0DE-4113-A249-B7CD4C0F610D"), "Pet Duration Increase (MST)", string.Empty);
```
ze-dom, line 1349: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L335), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L323)

Image

The tooltip is inaccurate: it's actually x66.78 increase: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L7982) (missing), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7613).

src/GameLogic/Attributes/Stats.cs
```diff
@@ -1332,14 +1332,31 @@ public class Stats
...
+
+    /// <summary>
+    /// Gets the trainable pet (dark horse, dark raven) duration increase MST attribute definition.
+    /// </summary>
+    public static AttributeDefinition TrainablePetDurationIncrease { get; } = new(new Guid("B4455150-D3A9-4A5F-914B-F41F9387FE9A"), "Trainable Pet Duration Increase (MST)", string.Empty);
```
ze-dom, line 1354: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/MasterLevelSkillTreeSystem.cpp#L417), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/MasterLevelSkillTreeSystem.cpp#L405)
Image
The tooltip is innacurate: it's not actually %, but absolute in this case: 
zTeamS6.3: [Horse](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8069), [Raven](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/DarkSpirit.cpp#L912)
emu: [Horse](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7709), [Raven](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/DarkSpirit.cpp#L915).

src/GameLogic/PlayerActions/Skills/PlasmaStormSkillPlugIn.cs
ze-dom, line 1: Move this to [Player.cs](https://github.com/MUnique/OpenMU/pull/966/changes#diff-7095817e5ea85690f8e681594b50539993eaf0a73fa1d6aab6d7f8dd421eb752), where all durability reduction is addressed.

src/GameLogic/PlugIns/PeriodicTasks/DurabilityDecayPlugIn.cs
```diff
@@ -0,0 +1,77 @@
...
+                            continue;
+                        }
+
+                        var isTransformationRing = item.IsTransformationRing();
+                        if (attributes[Stats.IsInSafezone] < 1 || isTransformationRing || item.IsWing())
+                        {
+                            double decrementWeight = identifier switch
+                            {
+                                var _ when isTransformationRing => 11.28, // 11.28 / 564 = 0.02
+                                var itm when itm == ItemConstants.WizardsRing => 7,
```
ze-dom, line 45: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8351), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7960)

src/GameLogic/PlugIns/PeriodicTasks/DurabilityDecayPlugIn.cs
```diff
@@ -0,0 +1,77 @@
...
+                                var itm when itm == ItemConstants.MoonstonePendant => 6.3,
```
ze-dom, line 46: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8366), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7986)
"Pendant" that goes on ring slot 😅

src/GameLogic/ItemExtensions.cs
```diff
@@ -15,10 +15,13 @@ namespace MUnique.OpenMU.GameLogic;
 /// </summary>
 public static class ItemExtensions
 {
+    private const byte StaffItemGroup = 5;
     private const byte ShieldItemGroup = 6;
 
     private static readonly byte[] AdditionalDurabilityPerLevel = { 0, 1, 2, 3, 4, 6, 8, 10, 12, 14, 17, 21, 26, 32, 39, 47 };
 
+    private static readonly float[] DurabilityFactors = { 0.2f, 0.3f, 0.5f };
+
```
ze-dom, line 24: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L1522), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L330)

src/GameLogic/ItemExtensions.cs
```diff
@@ -249,46 +264,91 @@ public static bool IsBook(this Item item, [NotNullWhen(true)] out float? bookRis
...
         }
 
-        var random = Rand.NextInt(3, 6);
-        var result = left ?? right ?? pendant;
+        if (damageType is DamageType.Wizardry or DamageType.Curse)
+        {
+            if (!(left?.IsWizardryWeapon(out _) ?? true) || (left?.CanHaveSkill() ?? false))
+            {
+                left = null;
+            }
```
ze-dom, line 323: We need the second condition to clear MG "magic swords"

src/GameLogic/ItemExtensions.cs
```diff
@@ -444,6 +504,53 @@ public static int CalculateDropLevel(this Item item)
...
+    /// Calculates the durability thresholds for the item.
+    /// </summary>
+    /// <param name="item">The item.</param>
+    private static void CalculateDurabilityThresholds(this Item item)
+    {
+        item.DurabilityThresholds ??= new double[3];
+        var maxDurability = item.GetMaximumDurabilityOfOnePiece();
+        for (int i = 0; i < DurabilityFactors.Length; i++)
+        {
+            item.DurabilityThresholds[i] = maxDurability * DurabilityFactors[i];
```
ze-dom, line 550: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L1522), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L330)

src/GameLogic/ItemExtensions.cs
```diff
@@ -444,6 +504,53 @@ public static int CalculateDropLevel(this Item item)
...
+        if (item.DurabilityThresholds is null)
+        {
+            item.CalculateDurabilityThresholds();
+        }
+
+        for (int i = 0; i < item.DurabilityThresholds!.Length; i++)
+        {
+            if (item.Durability < item.DurabilityThresholds[i])
+            {
+                return 1 - DurabilityFactors[^(i + 1)];
```
ze-dom, line 533: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L1565), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L372)

src/GameLogic/ItemPowerUpFactory.cs
```diff
@@ -17,6 +17,18 @@ public class ItemPowerUpFactory : IItemPowerUpFactory
...
+    [
+        Stats.DefenseBase,
+        Stats.DefenseShield,
+        Stats.DefenseRatePvm,
+        Stats.MinimumPhysBaseDmgByWeapon,
+        Stats.MaximumPhysBaseDmgByWeapon,
+        Stats.StaffRise,
+        Stats.ScepterRise,
+        Stats.BookRise,
+    ];
```
ze-dom, line 30: zTeamS6.3: [zzzitem.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L2767), [ObjAttack.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjAttack.cpp#L3369), [DarkSpirit.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/DarkSpirit.cpp#L1403)
emu: [zzzitem.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L1741), [ObjAttack.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjAttack.cpp#L4105), [DarkSpirit.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/DarkSpirit.cpp#L1511)

src/GameLogic/ItemPowerUpFactory.cs
```diff
@@ -296,7 +309,9 @@ private IEnumerable<PowerUpWrapper> GetPowerUpsOfItemOptions(Item item, Attribut
                 }
             }
 
-            foreach (var wrapper in PowerUpWrapper.CreateByPowerUpDefinition(powerUp, attributeHolder, aggregateType))
+            var durabilityFactor = (option.OptionType == ItemOptionTypes.Option && !item.IsJewelry()) || option.OptionType == ItemOptionTypes.AncientBonus
+                ? item.GetCurrentDurabilityFactor() : 1.0f;
```
ze-dom, line 314: zTeamS6.3: [zzzitem.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L4865) (regular item options  - except rings and wings hp rec), [zzzitem.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L4979) (ancient bonus opt)
emu: [zzzitem.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L2891), [zzzitem.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3006)

src/GameLogic/Player.cs
```diff
@@ -749,16 +751,19 @@ public async ValueTask KillInstantlyAsync()
...
         if (attacker as IPlayerSurrogate is { } playerSurrogate)
         {
-            await playerSurrogate.Owner.AfterHitTargetAsync().ConfigureAwait(false);
+            await playerSurrogate.Owner.DecreaseRavenDurabilityAfterHitAsync(hitInfo.TotalDamage).ConfigureAwait(false);
         }
 
         if (attacker is Player attackerPlayer)
         {
-            await attackerPlayer.AfterHitTargetAsync().ConfigureAwait(false);
+            await attackerPlayer.AfterHitTargetAsync(this.Attributes[Stats.DefenseFinal], skill?.Skill?.DamageType).ConfigureAwait(false);
```
ze-dom, line 766: zTeamS6.3: [ObjAttack.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjAttack.cpp#L484), [user.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L9318)
emu: [ObjAttack.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjAttack.cpp#L522), [user.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L8363)

src/GameLogic/Player.cs
```diff
@@ -1179,6 +1186,54 @@ public async ValueTask ResetPetBehaviorAsync()
...
+        var decrement = damage * 0.02 / (hitsPerOneItemDurability + attributes[Stats.TrainablePetDurationIncrease]);
+        await this.DecreaseItemDurabilityAsync(item, decrement).ConfigureAwait(false);
+
+        if (item.Durability == 0.0)
+        {
+            var minimumExp = item.Definition!.GetExperienceOfPetLevel(item.Level, item.Definition!.MaximumItemLevel);
+            item.PetExperience = (int)Math.Max((int)(item.PetExperience * 0.9), minimumExp);
+            await this.ResetPetBehaviorAsync().ConfigureAwait(false);
+        }
+    }
```
ze-dom, line 1216: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/DarkSpirit.cpp#L908), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/DarkSpirit.cpp#L911)

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
-        var randomDefensiveItem = this.Inventory?.EquippedItems.Where(ItemExtensions.IsDefensiveItem).SelectRandom();
-        if (randomDefensiveItem is { })
+        if (!isPvpDamage)
+        {
+            var randomDefensiveItem = this.Inventory?.EquippedItems.Where(ItemExtensions.IsDefensiveItem).SelectRandom();
+            if (randomDefensiveItem is { Durability: > 0.0 })
+            {
+                await this.DecreaseDefenseItemDurabilityAsync(randomDefensiveItem, damage).ConfigureAwait(false);
+            }
+        }
```
ze-dom, line 2020: Only mobs have `m_AttackDamageMin` > 0
[zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L9529), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L8570)

I think this also makes sense, because otherwise all PvP would wear armor down.

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
-                {
-                    await this.DestroyInventoryItemAsync(pet).ConfigureAwait(false);
-                }
-            }
+            await this.DecreasePetDurabilityAsync(pet, damage).ConfigureAwait(false);
         }
     }
 
-    private async ValueTask DecreaseDefenseItemDurabilityAsync(Item targetItem, HitInfo hitInfo)
+    private async ValueTask DecreaseDefenseItemDurabilityAsync(Item targetItem, float damage)
```
ze-dom, line 2049: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L6039), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3960)

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
-                {
-                    var minimumExp = pet.Definition!.GetExperienceOfPetLevel(pet.Level, pet.Definition!.MaximumItemLevel);
-                    pet.PetExperience = (int)Math.Max((int)(pet.PetExperience * 0.9), minimumExp);
-                }
-                else
-                {
-                    await this.DestroyInventoryItemAsync(pet).ConfigureAwait(false);
-                }
-            }
+            await this.DecreasePetDurabilityAsync(pet, damage).ConfigureAwait(false);
```
ze-dom, line 2045: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ObjAttack.cpp#L698), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/ObjAttack.cpp#L769)

Unfortunately pet durability decrease is originally calculated in the middle of damage calcs, so we diverge here a bit.

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+            return;
         }
+
+        var damageToDefenseRatio = damage / (itemDefense * itemDefenseFactor);
+        var decrement = damageToDefenseRatio / (this.GameContext.Configuration.DamagePerOneItemDurability * itemDurationIncrease);
+        await this.DecreaseItemDurabilityAsync(targetItem, decrement).ConfigureAwait(false);
     }
 
-    private async ValueTask DecreaseWeaponDurabilityAfterHitAsync()
+    private async ValueTask DecreasePetDurabilityAsync(Item pet, float damage)
```
ze-dom, line 2104: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L7955), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7588)

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
         {
-            await this.InvokeViewPlugInAsync<IItemDurabilityChangedPlugIn>(p => p.ItemDurabilityChangedAsync(targetItem, false)).ConfigureAwait(false);
+            var item when item == ItemConstants.Angel
+                       || item == ItemConstants.Rudolf => 3,
+            var item when item == ItemConstants.Imp
+                       || item == ItemConstants.DarkHorse
+                       || item == ItemConstants.Fenrir => 2,
+            var item when item == ItemConstants.Dinorant => 10,
+            var item when item == ItemConstants.Demon => 1.5,
+            _ => 1,
```
ze-dom, line 2136: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L7982), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7611)

A little bit of divergence between the two sources regarding dinorant and the cash shop pets. Went with emu because it has the master skill increase in place, which is missing in zTeamS6.3.

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+        };
 
-            if (targetItem is { Durability: 0.0 } pet && pet.IsTrainablePet())
+        double decrement;
+        double damagePerOnePetDurability = this.GameContext.Configuration.DamagePerOnePetDurability;
+        if (identifier == ItemConstants.DarkHorse)
+        {
+            decrement = 1 + (damage * damageFactor / damagePerOnePetDurability);
+            decrement /= 1500 + itemDurationIncrease;
+        }
```
ze-dom, line 2145: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8053), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7695)

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+        else if (identifier == ItemConstants.Fenrir)
+        {
+            decrement = 1 + (damage * damageFactor / (damagePerOnePetDurability * itemDurationIncrease));
+            decrement /= attributes[Stats.ArcheryMaxDmg] > 0 ? 160 : 200;
+        }
```
ze-dom, line 2150: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8095), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7739)

Elf-or-else divisor: [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/Template/zGameServer/Data/CommonServer.cfg#L409)

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+                        {
+                            await this.InvokeViewPlugInAsync<IUpdateInventoryListPlugIn>(p => p.UpdateInventoryListAsync()).ConfigureAwait(false);
+                            await this.ShowLocalizedBlueMessageAsync(nameof(PlayerMessage.EquipmentHasChangedMessage)).ConfigureAwait(false);
+                            return;
+                        }
+                    }
+
+                    await this._movement.StopWalkingAsync().ConfigureAwait(false);
+                    await this.WarpToSafezoneAsync().ConfigureAwait(false);
+                }
```
ze-dom, line 2188: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L8198), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7837)

Note: Icarus safezone map is actually Devias, not Lost Tower => todo

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+            }
+        }
+    }
+
+    private async ValueTask DecreaseWeaponDurabilityAfterHitAsync(float targetDefense, DamageType damageType)
```
ze-dom, line 2193: `NormalWeaponDurabilityDown()`: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5817), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3799)
`BowWeaponDurabilityDown()`: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5879), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3859)
`StaffWeaponDurabilityDown()`: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5935), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3909)

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+            return;
+        }
+
+        var defenseToDamageRatio = targetDefense * defenseFactor / weaponDamageFactor;
+        if (weapon.ItemOptions.FirstOrDefault(o => o.ItemOption?.PowerUpDefinition?.TargetAttribute == Stats.WeaponDurationIncrease) is { } opt
+            && opt.ItemOption?.LevelDependentOptions.FirstOrDefault(o => o.Level == opt.Level)?.PowerUpDefinition is { } pu
+            && pu.Boost?.ConstantValue is { } socketInc)
+        {
+            defenseToDamageRatio -= defenseToDamageRatio * socketInc.Value;
+        }
```
ze-dom, line 2263: zTeamS6.3: [zzzitem.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5902), [ItemSocketOptionSystem.cpp](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/ItemSocketOptionSystem.cpp#L1683)
emu: [zzzitem.cpp,](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3878), [SocketOptionSystem.cpp](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/SocketItemOption.cpp#L1719)

src/Persistence/Initialization/GameConfigurationInitializerBase.cs
```diff
@@ -58,9 +58,9 @@ public override void Initialize()
         this.GameConfiguration.ShouldDropMoney = true;
         this.GameConfiguration.ItemDropDuration = TimeSpan.FromSeconds(60);
         this.GameConfiguration.MaximumItemOptionLevelDrop = 3;
-        this.GameConfiguration.DamagePerOneItemDurability = 2000;
-        this.GameConfiguration.DamagePerOnePetDurability = 100000;
-        this.GameConfiguration.HitsPerOneItemDurability = 10000;
+        this.GameConfiguration.DamagePerOneItemDurability = 69;
+        this.GameConfiguration.DamagePerOnePetDurability = 100;
+        this.GameConfiguration.HitsPerOneItemDurability = 564;
```
ze-dom, line 63: zTeamS6.3: [NormalWeaponDurabilityDown()](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5850) (weapons), [DurabilityDown()](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L5748) (jewelry and wings)
emu: [NormalWeaponDurabilityDown()](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3829), [DurabilityDown()](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L3731)

src/Persistence/Initialization/GameConfigurationInitializerBase.cs
```diff
@@ -58,9 +58,9 @@ public override void Initialize()
         this.GameConfiguration.ShouldDropMoney = true;
         this.GameConfiguration.ItemDropDuration = TimeSpan.FromSeconds(60);
         this.GameConfiguration.MaximumItemOptionLevelDrop = 3;
-        this.GameConfiguration.DamagePerOneItemDurability = 2000;
-        this.GameConfiguration.DamagePerOnePetDurability = 100000;
-        this.GameConfiguration.HitsPerOneItemDurability = 10000;
+        this.GameConfiguration.DamagePerOneItemDurability = 69;
+        this.GameConfiguration.DamagePerOnePetDurability = 100;
```
ze-dom, line 62: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/GameServer/user.cpp#L7989), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/Source/user.cpp#L7619)

src/Persistence/Initialization/GameConfigurationInitializerBase.cs
```diff
@@ -58,9 +58,9 @@ public override void Initialize()
         this.GameConfiguration.ShouldDropMoney = true;
         this.GameConfiguration.ItemDropDuration = TimeSpan.FromSeconds(60);
         this.GameConfiguration.MaximumItemOptionLevelDrop = 3;
-        this.GameConfiguration.DamagePerOneItemDurability = 2000;
-        this.GameConfiguration.DamagePerOnePetDurability = 100000;
-        this.GameConfiguration.HitsPerOneItemDurability = 10000;
+        this.GameConfiguration.DamagePerOneItemDurability = 69;
```
ze-dom, line 61: [zTeamS6.3](https://github.com/MuSeason6FixTeam/zTeamS6.3/blob/565c579e61f150f6019dee60d22443d30ddfaf42/zGameServer/common/zzzitem.cpp#L6114), [emu](https://github.com/kyleruss/emu-server/blob/202856a74c905c203b9b2795fd161f564ca8b257/GameServer/common/zzzitem.cpp#L4025)

src/Persistence/Initialization/Updates/ItemDurabilityRefactorPlugInSeason6.cs
```diff
@@ -0,0 +1,118 @@
...
+        // Update ice socket option
+        var iceSocketOptionsId = new Guid("00000083-0034-0000-0000-000000000000");
+        if (gameConfiguration.ItemOptions.FirstOrDefault(io => io.GetId() == iceSocketOptionsId) is { } iceSocketOptions
+            && iceSocketOptions.PossibleOptions.FirstOrDefault(po => po.Number == 4) is { } durationIncrease)
+        {
+            foreach (var levelOption in durationIncrease.LevelDependentOptions)
+            {
+                levelOption.PowerUpDefinition?.TargetAttribute = weaponDurationIncrease;
+                levelOption.PowerUpDefinition?.Boost?.ConstantValue.Value -= 1.0f;
+                levelOption.PowerUpDefinition?.Boost?.ConstantValue.AggregateType = AggregateType.AddRaw;
```
sven-n: This doesn't compile: a null-conditional access can't be the target of an assignment (CS0131), which applies to all three lines (including the compound `-=`). This is most likely why the `MUnique.OpenMU` build check is red.

Something like this works:

```csharp
foreach (var levelOption in durationIncrease.LevelDependentOptions)
{
    if (levelOption.PowerUpDefinition is not { } powerUpDefinition)
    {
        continue;
    }

    powerUpDefinition.TargetAttribute = weaponDurationIncrease;
    if (powerUpDefinition.Boost?.ConstantValue is { } constantValue)
    {
        constantValue.Value -= 1.0f;
        constantValue.AggregateType = AggregateType.AddRaw;
    }
}
```

---

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+        if (targetItem.Durability == 0 || this.Attributes is not { } attributes)
+        {
+            return;
+        }
+
+        float itemDurationIncrease = attributes[Stats.WeaponAndArmorDurationIncrease];
         if (itemDurationIncrease == 0)
         {
             itemDurationIncrease = 1;
         }
```
sven-n: The duration-increase stat is used as a divisor, but the master skills now feed it values **below 1**, so investing skill points makes items decay *faster*.

`DurabilityReduction1` (and `DurabilityReduction1FistMaster`) is `AddRaw` of `Formula1204 / 100`, which is ~`0.12` at skill level 1 and ~`0.90` at level 20. The `== 0 → 1` guard only covers "no skill at all", so:

* skill level 1 → `decrement = ratio / (69 * 0.12)` ⇒ **~8x faster** durability loss than without the skill
* skill level 20 → `ratio / (69 * 0.90)` ⇒ still ~11% faster than without the skill

I think it wants to be a bonus on top of 1:

```csharp
float itemDurationIncrease = 1 + attributes[Stats.WeaponAndArmorDurationIncrease];
```

and then the `if (itemDurationIncrease == 0)` guard can go away.

The same pattern is in `DecreaseWeaponDurabilityAfterHitAsync` (line ~2201) and in `DurabilityDecayPlugIn` (`JewelryAndWingsDurationIncrease`), both fed by `Formula1204 / 100` skills as well.

---

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+            _ => 1,
+        };
 
-            if (targetItem is { Durability: 0.0 } pet && pet.IsTrainablePet())
+        double decrement;
+        double damagePerOnePetDurability = this.GameContext.Configuration.DamagePerOnePetDurability;
+        if (identifier == ItemConstants.DarkHorse)
+        {
+            decrement = 1 + (damage * damageFactor / damagePerOnePetDurability);
+            decrement /= 1500 + itemDurationIncrease;
```
sven-n: The four durability master skills end up with three different units, which makes their effects wildly inconsistent:

* `DurabilityReduction1/2` (+ fist master variants) → `Formula1204 / 100` (0.12 … 0.90), used as a **multiplicative divisor** — see the separate comment, this one currently backfires.
* `DurabilityReduction3` / `DurabilityReduction3FistMaster` → raw `Formula1204` (12 … 90) for `PetDurationIncrease`, also used as a multiplicative divisor (`damagePerOnePetDurability * itemDurationIncrease`) ⇒ **12x slower** pet decay from a single skill point, up to 90x.
* `PetDurabilityStr` → raw `Formula1204` for `TrainablePetDurationIncrease`, but **added** to a constant here (`1500 + …`) and in `DecreaseRavenDurabilityAfterHitAsync` (`800 + …`) ⇒ a ~1–6% effect.

Was the missing `/ 100` on `DurabilityReduction3`/`3FistMaster` intentional? It would be good to settle on one unit (percentage or factor) for the whole family and apply it the same way in all consumers.

---

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+            itemDefense = attributes[Stats.DefenseShield];
+            itemDefenseFactor = 5;
+        }
+        else
         {
-            damageDivisor *= (double)itemDurationIncrease;
+            var itemPowerUps = attributes.ItemPowerUps.FirstOrDefault(ipu => ipu.Key == targetItem).Value;
+            itemDefense = attributes.GetComposableAttribute(Stats.DefenseBase)?.Elements
+                .Where(e => e.AggregateType == AggregateType.AddRaw && itemPowerUps.Contains(e))
+                .Sum(e => e.Value) ?? 0;
```
sven-n: `FirstOrDefault` on a `KeyValuePair` returns `default` when the item isn't in the dictionary, so `itemPowerUps` can be `null` and `itemPowerUps.Contains(e)` then throws a `NullReferenceException` — the `?? 0` only guards `GetComposableAttribute` returning null.

`TryGetValue` avoids that and is O(1) instead of scanning the whole dictionary on every hit:

```csharp
if (!attributes.ItemPowerUps.TryGetValue(targetItem, out var itemPowerUps))
{
    return;
}
```

Same at line ~2228 in `DecreaseWeaponDurabilityAfterHitAsync`.

---

src/GameLogic/ItemExtensions.cs
```diff
@@ -249,46 +264,91 @@ public static bool IsBook(this Item item, [NotNullWhen(true)] out float? bookRis
...
+            {
+                right = null;
+            }
+        }
+        else
+        {
+            if (left?.Definition?.Group == StaffItemGroup || right?.Definition?.Group == StaffItemGroup)
+            {
+                right = null;
+            }
```
sven-n: This nulls the *right* hand when the *left* one holds the staff — is that intended, or should the staff itself (`left`) be excluded from taking durability loss on a non-magic hit? As written, a left-hand staff both survives the filter and removes the right-hand weapon from the selection, so the staff is always the one that loses durability.

---

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
...
+        }
+
+        if (skill?.Skill?.DamageType == DamageType.Fenrir)
         {
-            await this.DecreaseDefenseItemDurabilityAsync(randomDefensiveItem, hitInfo).ConfigureAwait(false);
+            var randomItem = this.Inventory?.EquippedItems.Where(i => i.ItemSlot <= InventoryConstants.BootsSlot).SelectRandom();
+            if (randomItem is { Durability: > 0.0 })
+            {
+                await this.DecreaseItemDurabilityAsync(randomItem, randomItem.Durability * 0.5).ConfigureAwait(false);
+            }
```
sven-n: The deleted `PlasmaStormSkillPlugIn` halved a random item's durability with a **25% chance** (`Rand.NextRandomBool(25)`); here it happens on *every* hit with `DamageType.Fenrir`. Since Plasma Storm is an area skill with automatic hits, that's a big change for the defender. Intentional, or should the chance be kept?

---

src/GameLogic/Player.cs
```diff
@@ -1953,85 +2008,262 @@ private async void OnAmmunitionAmountChanged(object? sender, EventArgs args)
         }
     }
 
-    private async ValueTask DecreaseItemDurabilityAfterHitAsync(HitInfo hitInfo, SkillEntry? skill)
+    private async ValueTask DecreaseItemDurabilityAfterHitAsync(float damage, bool isPvpDamage, SkillEntry? skill)
     {
-        var randomDefensiveItem = this.Inventory?.EquippedItems.Where(ItemExtensions.IsDefensiveItem).SelectRandom();
-        if (randomDefensiveItem is { })
+        if (!isPvpDamage)
```
sven-n: With this guard, defensive items take no durability loss at all when the attacker is a player (or a player surrogate) — only monsters wear armor down. Is that the intended original behaviour? It isn't mentioned in the PR description, so it would be nice to have a source reference/comment here.

---

src/DataModel/ItemExtensions.cs
```diff
@@ -173,6 +173,11 @@ public static bool DecreaseDurability(this Item item, double decrement)
     {
         var previous = item.Durability;
         item.Durability = Math.Max(previous - decrement, 0.0);
+        if (item.Durability < 1.0)
+        {
+            item.Durability = 0.0;
+        }
+
```
sven-n: This collapses the whole last durability point to 0, so an item breaks as soon as its durability drops below 1 instead of when it actually reaches 0 — roughly one durability point of use lost per item. If that's intended (to match what the client displays), a short comment would help; otherwise the clamping could be dropped since the callers already compare `(byte)Math.Floor(...)`.

---

src/DataModel/Entities/Item.cs
```diff
@@ -31,6 +32,15 @@ public partial class Item
...
     public double Durability { get; set; }
 
+    /// <summary>
+    /// Gets or sets the durability thresholds, past which the item's performance degrades.
+    /// </summary>
+    /// <remarks>
+    /// These are directly related to the client's UI item repair icons, whose colors change (yellow, light red, red) as durability decreases.
+    /// </remarks>
+    [NotMapped]
+    public double[]? DurabilityThresholds { get; set; }
```
sven-n: Two things about this cache:

1. The repo's convention for non-persisted, run-time-only members is `[Transient]` (`MUnique.OpenMU.Annotations`), which both `EfCoreModelGenerator` and the generated `MapsterConfigurator` key off. `[NotMapped]` keeps it out of the EF schema, but the mapping layer will still copy the array reference between the entity and the data model. `[NotMapped]` is otherwise not used anywhere in `DataModel`.

2. The cache is invalidated by hand in five places (`UpgradeItemLevelJewelConsumeHandlerPlugIn`, `BaseItemCraftingHandler` x2, `SimpleItemCraftingHandler`, `ItemViewModel` x2), and other places that change `item.Level` don't reset it — e.g. `BaseEventTicketCrafting.cs:107` and the bot gear equippers. It'd be more robust to store the level (and ancient/excellent state) the thresholds were computed for and recompute on mismatch, so it can't go stale by omission.

---

src/GameLogic/PlayerActions/Items/ItemRepairAction.cs
```diff
@@ -42,7 +42,13 @@ public async ValueTask RepairItemAsync(Player player, byte slot)
 
         if (IsMoneySufficient(player, item))
         {
+            var previousFactor = item.GetCurrentDurabilityFactor();
             item.Durability = item.GetMaximumDurabilityOfOnePiece();
+            if (previousFactor != item.GetCurrentDurabilityFactor())
+            {
+                await ((InventoryStorage?)player.Inventory)!.RaiseEquippedItemsChangedAsync(item, true).ConfigureAwait(false);
```
sven-n: `RepairItemAsync(player, slot)` can be called for any inventory slot, not just equipped ones. For a non-equipped item this now raises `EquippedItemsChanged`, which runs `UpdateItemsOnChangeAsync` → `OnAppearanceChanged` and broadcasts an appearance change to all observers for what is just a backpack repair. Worth guarding on the item actually being equipped.

Also, the `((InventoryStorage?)player.Inventory)!` cast (here, below, and in `Player.DecreaseItemDurabilityAsync`) would be nicer as a member on `IInventoryStorage` — the interface is what everything else consumes.

---

src/GameLogic/ItemPowerUpFactory.cs
sven-n: Nit: this can be `static readonly`, and since it's hit with `Contains` for every base power-up of every item, a `HashSet<AttributeDefinition>` (or `FrozenSet`) would avoid the linear scan.

---

src/Persistence/Initialization/Updates/UpdateVersion.cs
```diff
@@ -579,4 +579,19 @@ public enum UpdateVersion
...
     /// </summary>
     AddKanturuMapContent = 114,
+
+    /// <summary>
+    /// The version of the <see cref="ItemDurabilityRefactorPlugIn075"/>.
+    /// </summary>
+    ItemDurabilityRefactor075 = 115,
+
+    /// <summary>
+    /// The version of the <see cref="ItemDurabilityRefactorPlugIn095D"/>.
```
sven-n: Nit: the class is `ItemDurabilityRefactorPlugIn095d` (lower-case `d`), so this `cref` doesn't resolve (CS1574).

---

src/GameLogic/PlugIns/PeriodicTasks/DurabilityDecayPlugIn.cs
```diff
@@ -0,0 +1,77 @@
...
+                                _ => 0.1,
+                            };
+
+                            var durationIncrease = attributes[Stats.JewelryAndWingsDurationIncrease];
+                            if (durationIncrease == 0 || isTransformationRing)
+                            {
+                                durationIncrease = 1;
+                            }
+
+                            var decrement = decrementWeight / (player.GameContext.Configuration.HitsPerOneItemDurability * durationIncrease);
```
sven-n: Two notes here:

* `durationIncrease` has the same divisor problem as the weapon/armor path: `JewelryAndWingsDurationIncrease` comes from `DurabilityReduction2` as `AddRaw` of `Formula1204 / 100` (0.12 … 0.90), so one skill point makes jewelry and wings decay up to ~8x *faster*. Looks like it should be `1 + attributes[...]`.
* The condition means wings decay even inside a safezone, while jewelry doesn't. Is that deliberate (i.e. wings wear out while idling in town too)? A short comment with the source reference would help, like the ones above.

---

sven-n (commented):
Nice work — the direction looks good to me: the time-decay plug-in, the durability → power-up factor, and the repair event fix all make sense, and the update plug-ins cover the three data versions. A quick review with a few things that stood out.

**Blocking**

* `ItemDurabilityRefactorPlugInSeason6.cs` doesn't compile (assignment to a null-conditional access). That's very likely the failing `MUnique.OpenMU` check.
* The `*DurationIncrease` stats are used as divisors while the master skills feed them values below 1, so putting points into the durability skills currently makes items decay *faster* (~8x at skill level 1). Details inline.

**Worth a second look**

* Unit inconsistency across the four durability master skills (percent vs. factor, multiplied vs. added).
* Possible `NullReferenceException` from `ItemPowerUps.FirstOrDefault(...).Value` in both durability paths.
* `GetRandomWeapon`: the left-hand-staff case nulls the right hand.
* Two behaviour changes that aren't in the description: Plasma Storm's 25% chance is now 100%, and defensive items no longer lose durability in PvP.
* `DurabilityThresholds` cache: `[NotMapped]` vs. the repo's `[Transient]` convention, and manual invalidation that a few `item.Level` writers miss.

**Nits**

* `ItemDurabilityRefactorPlugIn075.ApplyAsync` only calls `base` — the override can go; `ItemDurabilityRefactorPlugInBase.ApplyAsync` is `async` without an `await` (CS1998), plus a stray blank line before `});`.
* Stale resources left behind: `PlasmaStormSkillPlugIn_Name` / `_Description` (plug-in deleted) and `ItemBasePowerUpDefinition_BaseValueElement_*` (property removed).
* `PlugInResources.Designer.cs` lost its UTF-8 BOM, and the new `.resx` entry is indented by four spaces instead of two.
* `GetRandomWeapon` mixes `!(x ?? true)` and `!x ?? true` in adjacent branches — both work, but they read as if one is a typo.
* `ItemPowerUpFactory:342` now uses `IsArmorItem()` where it was `IsDefenseItemSlot && !IsJewelry`, which drops wings from the excellent/ancient defense bonus. Harmless in practice, but it is a semantic change.

---

PR #967, opened 2026-09-18
Refactor Minigames and add new GM commands for testing

eduardosmaniotto:
Game masters get proper tooling to test and control mini game events (Blood Castle, Chaos Castle, Devil Square, Kanturu), plus a set of fixes that make manual starts/announcements reliable and two behavior-preserving refactors of the mini game hosting code.

## New GM commands

| Command | Effect |
|---|---|
| `/skipwait` | Skips the currently running timed wait (entering phase, countdown, event standby) for everyone in the event. Announces the skip to all participants. Only affects the wait running right now — kill-based waits report "nothing to skip" (use `/killall`) instead of arming a skip that would leak into a later phase. |
| `/killall` | Kills all alive monsters on the GM's map via the regular damage pipeline, so event kill counters, drops and XP work as usual. Guards/traps/players are spared. |
| `/startkanturu` | Force-starts Kanturu, matching the existing `/startcc`, `/startbc`, `/startds`. All four start commands now report back: "will start soon", or stop a live run first ("the running event was stopped…") and start fresh. |

## Fixes (found while testing the commands)

- **Second `/startX` silently died**: every start sets `NextRunUtc = start + TaskDuration`, so ticks bailed out for up to 15 min and the forced flag was never evaluated. A pending forced start now bypasses that check (limited to `NotStarted`, so `PreStartMessageDelay` still holds).
- **Finished runs blocked restarts**: `IsPreviousEventStillRunning` was pure wall-clock. It now also requires a live instance, so an early-ended event doesn't block a new start.
- **Restarts reused dead instances**: starting now disposes `Ended` leftovers (and full live runs on forced restart, warping players out) with a bounded retry, guaranteeing fresh instances.
- **Stale countdowns after skip/restart**: entrance announcements were a fire-and-forget `Task.Delay` loop unaware of the lifecycle. They are now derived from live game state on the 1s periodic tick (`EnterEndsAtUtc` + per-run progress) — nothing sleeps in the background, so there is nothing to skip, cancel, or supersede.
- **Ghost lobbies**: read-only entrance queries (`GetMiniGameAsync`) created real game instances as a side effect. The enter gate, the opening-state query and the Kanturu gateway now use non-creating lookups.

## Refactoring

- `MiniGameContext` (1151 → ~700 lines) split into `MiniGamePlayerRegistry`, `MiniGameRewardService`, `MiniGameChangeEventProcessor`, `MiniGameSpawnWaveRunner`.
- Mini game hosting moved from `GameContext` into `IMiniGameManager`/`MiniGameManager` (`IGameContext.MiniGames`): `GetOrCreateAsync`, `TryGetRunningMiniGame` (nullable return, never creates), `GetRunningMiniGames`, `RemoveAsync`. `GameContext` only forwards map created/removed events.

https://github.com/user-attachments/assets/9ca03b7c-c4f9-464d-8625-bdc0680f8b74


sven-n:
## Review

Nice piece of work — the GM tooling is genuinely useful, and replacing the fire-and-forget `Task.Delay` announcement loop with state derived from the live games on the periodic tick is the right call. The extraction of `MiniGamePlayerRegistry` / `MiniGameRewardService` / `MiniGameChangeEventProcessor` / `MiniGameSpawnWaveRunner` reads well and makes `MiniGameContext` much easier to follow.

I reviewed the diff statically (no .NET SDK in my environment, so I didn't build or run the tests locally — Azure CI is green, Codacy is `action_required`). Findings below, ordered by severity.

---

### 1. The entering / state-transition race: `_enterLock` no longer covers `State`

This is the one I'd want addressed before merge.

Previously `_enterLock` guarded the player set **and** every `State` transition atomically: `TryEnterAsync` checked `State != Open` and added the player under the *same* writer lock that `CloseEntranceAsync`, `StartAsync`, `StopAsync` and `DisposeAsyncCore` took to mutate `State`.

After the split, the registry owns its own lock (`MiniGamePlayerRegistry:41`) and only *reads* `State` through the `Func<MiniGameState>`, while the transitions happen completely outside it (`MiniGameContext.cs:254, 627, 633, 648`). So:

* `TryEnterAsync` observes `State == Open`, then `await areEquippedItemsAllowedAsync(player)` yields — that's an async call made while holding the registry lock, but `CloseEntranceAsync`/`StartAsync` don't contend for that lock any more;
* `CloseEntranceAsync` sets `Closed`, `StartAsync` sets `Playing` and takes `GetSnapshotAsync()`;
* `TryEnterAsync` resumes and adds the player.

Result: a player registered in a game that is already `Playing`, who never went through `OnGameStartAsync`. The same shape against `DisposeAsyncCore` is worse — `MovePlayersToSafezoneAsync` calls `ClearAsync()`, then the late `TryEnterAsync` re-adds the player and `MiniGameContext.TryEnterAsync` sets `player.CurrentMiniGame = this` on a disposed context, with nothing left to warp them out.

Simplest fix that keeps the extraction: let the registry own the state, e.g. `ValueTask SetStateAsync(MiniGameState)` on `MiniGamePlayerRegistry` taking the writer lock, and have `MiniGameContext.State` transitions go through it — restoring the original atomicity without moving the collection back.

### 2. `MiniGameManager` reads `_miniGames` outside its lock

`Maps` (`MiniGameManager.cs:43`), `TryGetRunningMiniGame` (`:103`) and `GetRunningMiniGames` (`:114`) enumerate / probe a plain `Dictionary<,>` that `GetOrCreateAsync` and `RemoveAsync` mutate under `_lock`.

This is a regression, not inherited: in `GameContext` the enumeration lived in `GetMapsAsync`, which held `_mapInitializerLock` — the *same* lock `GetMiniGameAsync` used. Now `GameContext.GetMapsAsync` holds `_mapInitializerLock` and calls `MiniGames.Maps`, which is guarded by a different lock it doesn't hold, so `Maps` can throw `InvalidOperationException: Collection was modified` (or read a torn bucket) when a game is created concurrently. And `GetRunningMiniGames` is now called on *every* 1s tick per plugin per context (`AnnounceEntranceAsync`), so the window is hit constantly.

A `ConcurrentDictionary` for `_miniGames` (keeping `_lock` only to serialize the create-check-add) would fix all three at once.

### 3. `GetOrCreateAsync` can hand out a mid-dispose instance

The fast path checks both flags (`MiniGameManager.cs:50`), but inside the lock only `IsDisposed` is checked (`:59`) — a context that is `IsDisposing` but not yet `IsDisposed` is returned as if it were live.

That is exactly what the new retry loop in `MiniGameStartBasePlugIn.OnStartedAsync` papers over ("A stale instance slipped through (e.g. caught mid-dispose)"). I'd rather fix the cause:

```csharp
if (this._miniGames.TryGetValue(miniGameKey, out miniGameContext))
{
    if (miniGameContext is { IsDisposed: false, IsDisposing: false })
    {
        return miniGameContext;
    }

    this._miniGames.Remove(miniGameKey);
}
```

and then drop the bounded retry, which is hard to reason about (it disposes whatever `GetOrCreateAsync` just returned — including, in principle, a fresh instance that simply isn't `Open` yet).

### 4. `RemoveAsync` removes by key without an identity check

`this._miniGames.Remove(miniGameContext.Key)` (`MiniGameManager.cs:127`) removes whatever currently sits under that key. With the forced-restart flow this PR adds, dispose and re-create for the same key now happen back to back: if `GetOrCreateAsync` already evicted the disposed entry and added a **new** context, the old context's `DisposeAsyncCore` → `RemoveAsync` unregisters the new one. The new game keeps running but is invisible to `Maps`, `TryGetRunningMiniGame` and `GetRunningMiniGames`, and `GameMapRemoved` is raised with the old map. `MiniGameCounter` drifts too.

```csharp
if (this._miniGames.TryGetValue(miniGameContext.Key, out var current) && ReferenceEquals(current, miniGameContext))
{
    this._miniGames.Remove(miniGameContext.Key);
}
```

### 5. `TryGetRunningMiniGame(definition, null)` throws for non-`Shared` definitions

`MiniGameMapKey.Create` dereferences `requester.Name` / `requester.Party` for `OnePerPlayer` and `OnePerParty`, so `MiniGameMapKey.Create(miniGameDefinition, requester!)` (`MiniGameManager.cs:103`) is an NRE for those policies — despite the XML doc saying the requester "may be `null` for shared games".

Before this PR the equivalent `GetMiniGameAsync(definition, null!)` calls were gated behind `state.State == PeriodicTaskState.Started`; now `GetDurationUntilNextStartAsync` and `GetMiniGameContextAsync` call it unconditionally, and both are reachable from `MiniGameOpeningStateRequestAction:27/30` and `KanturuGatewayPlugIn:59/63` with any suitable definition — `EnterMiniGameAction` is the only caller that guards on `MapCreationPolicy == Shared`. The shipped initializers all use `Shared`, so this is latent rather than live, but the policy is admin-configurable. Cheapest fix: in `TryGetRunningMiniGame`, return `null` when `requester is null && definition.MapCreationPolicy != Shared`.

### 6. Chaos Castle reward drops silently change drop generator

`MiniGameContext.GiveItemRewardAsync` used `this._gameContext.DropGenerator`; the extracted service is wired with `() => this.DropGenerator` (`MiniGameContext.cs:63`) and uses it at `MiniGameRewardService.cs:190`. For Chaos Castle that property is replaced by `ChaosCastleDropGenerator` (`ChaosCastleContext.cs:518`), so reward item generation now goes through a different generator than before.

Might well be the more correct behaviour — but it isn't behaviour-preserving, so either pass `() => this._gameContext.DropGenerator` or call it out explicitly in the PR description.

---

### Medium

* **`EntranceClosedMessage` lost for short entrances.** `OnStartedAsync` only creates the `EntranceAnnouncement` when `initialMinutes >= 1`, and `AnnounceEntranceAsync` only sends the closing message when an entry with `OpenAnnounced` exists. With `EnterDuration < 1 min` the old loop still sent `EntranceClosedMessage`; now nothing is sent. Creating the entry unconditionally (with `OpenAnnounced` reflecting whether the open message went out) would keep it.
* **`EnterEndsAtUtc` doesn't match the real entrance window.** It's `creation + Definition.EnterDuration` (`MiniGameContext.cs:46`), but `RunGameAsync` waits `EnterDuration.AtLeast(30s)`. For short entrance configs the announced remaining time undershoots the actual close. Deriving it from the same `AtLeast(countdownMessageDuration)` value would keep the two in sync.
* **`AnnounceEntranceAsync` doesn't filter to `Shared`.** `GetRunningMiniGames(this.Key)` returns per-player/per-party instances too, so a `OnePerPlayer` lobby created by an entering player would drive the *global* entrance countdown. `OnStartedAsync` only ever creates `Shared` games — worth matching that filter here.
* **`ChangeEventContext` is a public API break.** It went from `protected MiniGameContext.ChangeEventContext` to a public top-level type in `MiniGameChangeEventProcessor.cs:212`. `NextEvent` still compiles, but any external subclass referring to `MiniGameContext.ChangeEventContext` breaks — worth noting in the description (the PR says "Public/protected API unchanged"). Also, two types in one file trips SA1402; consider its own file.
* **`ExecuteTaskAsync` override runs before the base's guards.** The `Started → NotStarted` + `OnFinishedAsync` transition (`MiniGameStartBasePlugIn.cs:80`) happens outside `logger.BeginScope(gameContext)`, before the null-configuration check, and before the `NextRunUtc` gate — so it fires ~1s after the event starts and the `Started` state is effectively never observable. Harmless today (`OnFinishedAsync` is a no-op and nothing else reads `PeriodicTaskState.Started`), but it's a trap for a future override. A `protected virtual bool FinishImmediatelyAfterStart => true` hook inside the base's switch would express the same intent without reordering the guards.
* **`Announcements` entries can outlive their game context.** The static dictionary is only cleaned in the close-announcement path; a context torn down mid-run (or a run whose entry was never created, see above) leaves a strong reference to the `IGameContext`. It matches the existing `States` pattern, so not a new class of problem, but that one is bounded by the number of contexts while this one isn't cleaned on shutdown either.

### Nits

* The four `Start*EventChatCommandPlugIn` classes are now ~20 identical lines apiece differing only by `MiniGameType`. A small `StartMiniGameEventChatCommandPlugInBase` with an abstract `MiniGameType Key` would collapse all four (and make the next event a one-liner).
* `GetDurationUntilNextStartAsync` is `async` with no remaining `await` (CS1998 — suppressed to suggestion here, but still).
* `CloseEntranceAsync` no longer does anything async; it can be a plain `private void CloseEntrance()`.
* `SkippableDelay.WaitAsync`: on the skip path the cancelled `delayTask` is never awaited, so it stays an unobserved cancelled task. Harmless in practice, but a discard-with-await in a `catch` would keep it tidy. Also, the "at most one active wait" invariant is documented but silently violated if two waits overlap — the second overwrites `_activeWait` and the first becomes unskippable; a `LogWarning` there would help debugging.
* `KillAllMonstersChatCommandPlugIn` logs per-monster failures at `LogDebug`; `LogWarning` seems more appropriate for an unexpected exception in a GM command.
* `IPeriodicMiniGameStartPlugIn.cs`, `StartBloodCastleEventChatCommandPlugIn.cs` and `StartDevilSquareEventChatCommandPlugIn.cs` still end without a trailing newline (`StartChaosCastleEventChatCommandPlugIn.cs` got one) — worth making consistent while you're in there.
* Test coverage is `SkippableDelay` only. The genuinely tricky new logic — the forced-start bypass of `NextRunUtc`, `IsPreviousEventStillRunning` + `IsEventActive`, and the tick-derived announcement state machine — has none. Those are all pure-ish and testable with a fake `IGameContext`/`IMiniGameManager`; a couple of tests there would be worth more than the delay tests.
* Codacy is reporting `action_required` on this PR — worth a look before merge.

---

eduardosmaniotto:
Fixed as suggested:
- The enter/state race (1): registry owns the state now, all transitions go through SetStateAsync under the writer lock.
- Manager dictionary races (2), mid-dispose handoff (3), remove-by-key without identity check (4), and the null-requester NRE (5) — all fixed, counter stays balanced in the evict/replace paths.
- Drop generator (6): kept () => this.DropGenerator — verified ChaosCastleDropGenerator delegates item generation back to the context generator, so behavior is unchanged. Left a comment saying so.
- Short entrances now get their closing message, EnterEndsAtUtc uses the same AtLeast window as the run loop, tick only looks at Shared games, ChangeEventContext got its own file.
- Nits: start commands share a base class, GetDurationUntilNextStartAsync is sync again, KillAll logs warnings, newlines fixed, SkippableDelay observes the cancelled delay and warns on overlapping waits.
One thing I did differently:
The FinishImmediatelyAfterStart hook for (11) is in, but on its own it reintroduces slow restarts — the Started branch sits behind the NextRunUtc gate, so a forced restart would park its flag for up to TaskDuration again (that's the "event never starts" bug from testing). I kept the hook but extended the forced bypass to cover Started too, so a forced tick finishes the stale run through the existing branch and starts ~2 ticks later. Automatic flows are untouched. Added a test for exactly that case since the old bypass test only covers NotStarted.

One more thing that landed on this branch since your review: fixed event area reopening for mixed wall/hole terrain (GameMapTerrain). Removing a non-safezone attribute now clears every movement-blocking bit instead of just the configured one. Needed it for the Kanturu barrier — its rect on the loaded Terrain40.att mixes walls and holes, and per-bit removal left the 527-tile wall band shut with no south-north passage, so Elphis stayed unreachable after Nightmare died. This just restores the previous semantics (removal always opened the area) for all shipped event configs, with tests updated accordingly.

sven-n:
## Re-review (head `66ad9f46`)

Thanks — I went through `a37a6260` and `7119115d` against the previous head. All six blocking items and the medium/nit list are genuinely fixed, not just papered over: the registry owning the state is the right shape, `GetOrCreateAsync` evicting dead entries lets the retry loop disappear entirely, and the counter now balances across the evict/replace paths. The new `MiniGameStartPlugInTests` cover the cases that actually needed covering, including the stuck-`Started` restart.

On (6): your justification checks out — `ChaosCastleDropGenerator.GenerateItemDrop(DropItemGroup)` delegates to `_gameContext.DropGenerator` (`ChaosCastleDropGenerator.cs:86`), and that's the overload the reward service uses. Behaviour is preserved; the comment you left is accurate.

Four things this round, again worst first. The terrain commit is the one I'd hold the merge on.

---

### 1. The terrain change breaks the Castle Siege gate round-trip

`7119115d` removes the per-bit clear (`GameMapTerrain.cs:335`) and with it the comment that warned about exactly this caller, plus the test that pinned it (`ApplyTerrainAttributeRemoveKeepsOtherBits`).

`CastleSiegeNpcController` stores the pre-close state as **only the `Blocked` bit** — `this._gateTerrain[point] = (map.Terrain.BlocksSight(x, y), 1)` (`:236`) — and restores it with `ApplyTerrainAttribute(x, y, Blocked, state.WasBlocked)` (`:277`). When `WasBlocked` is `false`, that call now clears `Blocked | NoGround | Water` instead of just `Blocked`. So any gate tile that was originally a hole or water — unwalkable, but not a wall, so `WasBlocked == false` — is permanently converted to walkable ground the first time a gate opens. The old per-bit clear restored it correctly.

I can't check the shipped `Land1.att` gate rects from here, so I don't know whether that's live today or latent. But the contract the caller depends on is gone, and the commit message's "reproduces the previous semantics for all shipped event configs" only reasons about *event* configs — the castle gate isn't one.

Two smaller things on the same change:

* `MovementBlockingAttributes` is `Blocked | NoGround | Water` (4|8|16), but `IsWalkableValue` treats *any* non-safezone bit as blocking — including `Character` (2) and any high bits present in raw `.att` data. So "removal always opens the area" isn't actually guaranteed by this mask either; it just happens to cover the cases you tested.
* Clearing `Water` as well widens the change beyond the walls+holes problem you were solving.

Suggested scope-down: make "open the area" an explicit intent rather than a new global meaning for every removal — e.g. a flag on `MiniGameTerrainChange` (or an extra parameter on `ApplyTerrainAttribute`) that the Kanturu barrier opts into, leaving the castle gate on the per-bit restore it was written against. That also lets `ApplyTerrainAttributeRemoveKeepsOtherBits` come back instead of being inverted.

Separately: this is a distinct bug in a different subsystem from the PR's subject. It'd be easier to review — and to revert if it turns out to regress sieges — as its own PR.

### 2. `RemoveAsync` no longer raises `GameMapRemoved` when the key was taken over

The identity check (`MiniGameManager.cs:139`) correctly gates the dictionary removal and the counter, but it also gates `GameMapRemoved` (`:143`). In the exact scenario the check exists for — a forced restart already replaced the dying instance — the old instance's map is now disposed without ever announcing its removal.

`GameMap.Id` is a `Guid` per instance, so there's no risk of the event hitting the replacement's entry: `ObservableGameServerAdapter.OnGameMapRemoved` looks up by that Guid. Skipping it just leaves a stale map in the admin map view plus a live `PropertyChanged` subscription on a disposed map, for the lifetime of the server.

The event is about the map instance, not the key, so it should fire unconditionally:

```csharp
if (this._miniGames.TryGetValue(miniGameContext.Key, out var current)
    && ReferenceEquals(current, miniGameContext)
    && this._miniGames.TryRemove(miniGameContext.Key, out _))
{
    MiniGameCounter.Add(-1);
}

this.GameMapRemoved?.Invoke(this, miniGameContext.Map);
```

### 3. `State` is now a blocking lock acquisition on hot paths

`MiniGameContext.State` was a plain auto-property; it's now `this._players.State` (`MiniGameContext.cs:117`), which takes a **synchronous** `ReaderLock()` (`MiniGamePlayerRegistry.cs:40`). Every read pays for it, and some of those reads are hot:

* `AttackableNpcBase.ShouldRespawn` → `IsEventRunning` (`:101`), per monster respawn check;
* `BotPvpRules` per PvP evaluation;
* `AnnounceEntranceAsync` → `GetRunningMiniGames(...).Where(g => g.State == Open)`, every second per plugin per context.

That matters because the same lock is held across I/O elsewhere: `TryEnterAsync` holds the **writer** lock across `areEquippedItemsAllowedAsync`, which sends packets (`ShowLocalizedBlueMessageAsync`), and `ForEachAsync` holds the reader lock across arbitrary per-player work. While a writer is in there, every `State` read blocks a thread-pool thread rather than yielding.

Correctness doesn't need the lock on the read side — what has to be atomic is *check-and-add* inside `TryEnterAsync`, and that stays atomic as long as writes go through `SetStateAsync`. So keep `SetStateAsync` as-is and make the read lock-free:

```csharp
private volatile MiniGameState _state = MiniGameState.Open;   // enum with int base, volatile is allowed

public MiniGameState State => this._state;
```

(Writes stay inside the writer lock; `TryEnterAsync` keeps reading the field under it.)

### 4. Spurious "entrance closed" when no shared definition exists

`OnStartedAsync` now creates the announcement entry unconditionally (`MiniGameStartBasePlugIn.cs:190`) — which is the right fix for short entrances. But the loop above it (`:175`) may not create a single game: if the type has no `Shared` definition configured (removed in the admin panel, or switched to `OnePerPlayer`), `enterDuration` stays `TimeSpan.Zero` and no game exists. On the next tick `maxRemaining` is `null`, the entry is found, and `EntranceClosedMessage` goes out globally for a run that never opened. The old loop returned early on `enterDuration <= 0` and sent nothing.

One-line guard: only write the entry when at least one game was actually created in the loop.

---

### Smaller notes

* **`FinishImmediatelyAfterStart` is never overridden to `false`.** The seam is in and the default preserves existing behaviour for `HappyHourPlugIn` / `BaseInvasionPlugIn` / `WanderingMerchantsPlugIn`, which is right — but nothing opts out, so the `break` branch is dead today. Its doc comment ("finished right away on the next tick instead of lingering ... until the task duration elapses") also describes what happens only on the forced path; on the automatic path the `Started` branch still sits behind `NextRunUtc`, so the state does linger. Worth rewording so the next reader isn't misled. Your call on keeping an unused hook — extending the forced bypass to `Started` instead of pre-empting the base was the better half of that change either way.
* **`MiniGamePlayerRegistryTests` passes `null!` as the player.** `EnterWhenFullIsRejectedAsync` only reports `Full` because both adds are the same `null` reference into a `HashSet`; with two real players it would still pass, but with the *same* player twice it wouldn't — the set dedupes and the count never reaches the maximum. The test would also not survive `CountAlive()` being called. Two distinct stub players would make it test what it claims.
* **`MapsterConfigurator.Generated.cs` is unrelated churn.** The regenerated file picks up transient-member handling and a `GuildMember` `ConstructUsing` that have nothing to do with mini games — presumably the committed copy on master is stale and your local run refreshed it. Either way it's noise in this diff; worth splitting out or reverting.
* Codacy is still `action_required` on this head. Azure CI is green (build `4786`). I reviewed statically again — no .NET SDK here, so I haven't run the new tests locally.

---

eduardosmaniotto:
## 1. Terrain scope-down (the merge-blocker)

You're right, and it was worse than described: I read the shipped `.att` bytes instead of trusting the initializer comments. The d5fbefc74 revert comment ("bridge is NoGround on NoGround tiles") is factually wrong — BC1's bridge rect (13–15, 70–75 in `Terrain11.att`) is **all `Blocked`**, toggled via `NoGround` removal, so per-bit removal never opens it. Same story on BC8 (all `Blocked|NoGround`, breaks both bridge and gate toggles). Kanturu barrier confirmed mixed (402 holes + 527 walls in `Terrain40.att`).

So neither global default works, and I didn't take the entity-flag route — no shipped change-event needs per-bit-where-mixed except through paths I could cover without an EF migration:
- default removal is per-bit again (gate round-trip correct),
- `MiniGameChangeEventProcessor` passes `openArea` on removals (every shipped event config authors removals as "open this rect"),
- `KanturuContext` barrier passes it explicitly,
- the opt-in clears all non-safezone bits, so the Character/high-bit and Water sub-points hold on those paths (no shipped rect contains water anyway — checked).

The old `...RegardlessOfRawValue` test pinned the buggy global behavior, so I rewrote it to the scoped semantics plus kept mixed-tile tests for both paths. 62/62 green.

## 2. `RemoveAsync` event gating

Fixed as suggested — counter stays inside the identity check, `GameMapRemoved` fires unconditionally.

## 3. Lock-free `State`

Done: `volatile` field, lock-free read, writes still through `SetStateAsync`. Left `CountAlive()` locked — only `State` was contended on the paths you listed.

## 4. Spurious close with no shared definitions

Guarded: the announcement entry is only seeded when ≥1 game was created in the loop.

## Smaller notes

- Hook doc reworded to say what actually happens (forced path finishes promptly, automatic path still lingers behind the gate). Kept the hook.
- Registry tests now use real `Player` stubs (`InMemoryPersistenceContextProvider`, TradeTest pattern) — `Full` genuinely means two occupants.
- Mapster churn isn't in the tree — nothing to revert.
- Codacy loop nit fixed by bounding the retry explicitly; the empty `catch (OperationCanceledException)` stays (cancellation *is* the game-ended signal, same as the original).
- `ChangeEventContext` is `internal` now, with `NextEventRequiredKills`/`NextEventActualKills` replacing the exposed object — so no public break left to note.

## One thing I want to flag back

The `FinishImmediatelyAfterStart` hook as merged can't fire promptly: the `Started` branch sits behind the `NextRunUtc` gate, so a forced restart parks its flag for up to `TaskDuration` (the "event never starts" bug from testing, back again). The existing bypass test only covers `NotStarted`, so nothing catches it. I extended the bypass to `NotStarted or Started` — one line, automatic flows and other task types untouched — plus a test asserting the stuck-`Started` → `NotStarted` → `Prepared` sequence. Worth a look at `PeriodicTaskBasePlugIn.cs:59` since it interacts with everything above.

sven-n:
## Re-review (head `aed27f96`) — good to merge

All four items from the last round are fixed, and the terrain one is fixed the right way rather than the convenient way.

**Terrain scoping.** `openArea` as an explicit opt-in is exactly right, and I went back through the history to check the premise rather than take it on trust. Before `59d76b11` ("Distinguish walls from holes in line-of-sight checks"), `MiniGameContext.UpdateServerTerrain` set `WalkMap[x, y] = !setAttribute` directly — mini game removals opened the rect unconditionally, whatever the raw bits were. That commit routed them through the then-new per-bit `ApplyTerrainAttribute`, which is what silently shut the mixed-tile rects. So `openArea: true` on `MiniGameChangeEventProcessor` removals genuinely restores the long-standing semantics and the Kanturu barrier is a regression from `59d76b11`, while the castle siege gate keeps the per-bit restore it was written against (`CastleSiegeNpcController:277` → default overload). Both callers now get what they need, and `ApplyTerrainAttributeRemoveKeepsOtherBits` is back to pin it. Clearing down to `SafezoneBit` also closes the gap I raised about the old `Blocked|NoGround|Water` mask not covering every bit `IsWalkableValue` treats as blocking.

**The other three.** `GameMapRemoved` fires unconditionally with the identity check still gating the dictionary removal and the counter; `_state` is a lock-free `volatile` read with writes still inside `SetStateAsync`; the announcement entry is only seeded when a game was actually created. `ChangeEventContext` going `internal` with `NextEventRequiredKills`/`NextEventActualKills` on the context is better than what I suggested — it keeps the public surface unchanged instead of adding a type to it.

Two leftovers, neither worth holding the merge for:

* `FinishImmediatelyAfterStart` is still never overridden to `false`, so the `break` branch is dead today — but the doc comment is accurate now, and it's a reasonable seam to leave in place.
* `MapsterConfigurator.Generated.cs` stays in the diff. I checked the generator: `EfCoreModelGenerator.cs:237/273/277` already emits the transient-member ignore, the value-type-collection `UseDestinationValue` and the `GuildMember` `ConstructUsing`, so master's committed copy was simply stale and this is what a build produces. Incidental, but correct.

The PR description still says the refactor leaves the public/protected API unchanged; `protected ChangeEventContext? NextEvent` was replaced by the two int accessors. Worth a one-line tweak whenever you next touch it.

Azure CI is green on this head (build `4789`), the branch merges cleanly (no conflicts, and none of master's 24 newer commits touch a file this PR changes). Codacy is still `action_required`, as it has been on every head of this PR — it completes in zero seconds, which reads like an authorization state for a fork PR rather than a finding. I reviewed statically throughout; no .NET SDK in my environment, so the test runs are CI's.

Merging on @sven-n's instruction.

---

sven-n:
Merged as `4ee8b04b`.

For the record, the head moved from `aed27f96` (the SHA my sign-off named) to `4894e162` after I wrote it, so I re-reviewed the delta before merging: it's the removal of the unused `FinishImmediatelyAfterStart` hook and nothing else. Good call removing it — the net change to `PeriodicTaskBasePlugIn` vs master is now just `virtual` plus the forced-start bypass, which is as small as that file can get. I also checked that `ForceStart()` on a `PeriodicTaskBasePlugIn` subclass is only ever reached from `StartMiniGameEventChatCommandPlugInBase`, so the `Started` bypass can't surprise Happy Hour, the invasions or the wandering merchants. Azure CI green on that head (build `4812`).

On your flag-back: agreed, and you were right to push on it. My `FinishImmediatelyAfterStart` suggestion was aimed at the wrong half of the problem — I wanted the transition out of the base's pre-empted position, but the seam alone reinstates the restart latency, and extending the bypass to `Started` is the part that actually fixes it. The test asserting `Started → NotStarted → Prepared` is the one that would have caught my version.

And thanks for reading the `.att` bytes rather than the initializer comments on the terrain question — BC1's rect being all `Blocked` under a `NoGround` removal is the detail that settles it, and it lines up with what the history shows: removals opened the rect unconditionally until `59d76b11` routed them through the per-bit path. Worth keeping in mind that master carried that bug for both Blood Castle and Kanturu in the meantime.

---

